PR #4, opened 2018-10-05
Develop/new_fluview_data

stevenabreu7:


src/acquisition/fluview/fluview_update.py
```diff
@@ -132,23 +179,201 @@ def get_ilinet_data(row):
...
+    'PERCENT B'
+  ]:
+    raise Exception('header row has changed for clinical lab data.')
+  if len(row) == 1 or row[0] == 'REGION TYPE':
+    # this is a header row
+    return None
+  if row[4] == 'X':
+    # data is not reported, ignore this row
+    return None
+  # ignore percentage calculations for now
```
undefx: Just curious why percent positive is ignored for now? Are you planning to add it later?

src/acquisition/fluview/fluview_update.py
```diff
@@ -132,23 +179,201 @@ def get_ilinet_data(row):
...
+    'total_b': optional_int(row[6])
+  }
 
-def load_zipped_csv(filename):
-  """Read rows from a zipped CSV, which is expected to be named ILINet.csv."""
+def get_public_data(row):
+  if row[0] == 'REGION TYPE' and row != [
+    'REGION TYPE', 
+    'REGION', 
+    'SEASON_DESCRIPTION',
```
undefx: I'm not familiar with this data, but it's strange to me that there's not an `epiweek` field. Is `epiweek` included in this description somehow?

src/acquisition/fluview/fluview_update.py
```diff
@@ -132,23 +179,201 @@ def get_ilinet_data(row):
...
+  by the sheetname parameter. Default is ILINet.csv, for the default flu data."""
   with zipfile.ZipFile(filename) as f:
-    with f.open('ILINet.csv') as ff:
+    with f.open(sheetname) as ff:
       return [row for row in csv.reader(io.StringIO(str(ff.read(), 'utf-8')))]
 
-
-def get_rows(cnx):
+def get_rows(cnx, table='fluview'):
   """Count and return the number of rows in the `fluview` table."""
```
undefx: Small request: please update doc to say something to the effect of "number of rows in the given table, `fluview` by default"

src/acquisition/fluview/fluview_update.py
```diff
@@ -132,23 +179,201 @@ def get_ilinet_data(row):
...
-  select.execute('SELECT count(1) num FROM fluview')
+  select.execute('SELECT count(1) num FROM %s' % table)
   for (num,) in select:
     pass
   select.close()
   return num
 
+def update_from_file_clinical(issue, date, filename, test_mode=False):
+  """
+  Read ILINet data from a zipped CSV and insert into (or update) the database.
```
undefx: Small correction: instead of ILINet, this data comes from WHO/NREVSS.

src/acquisition/fluview/fluview_update.py
```diff
@@ -132,23 +179,201 @@ def get_ilinet_data(row):
...
+    rows2 = rows1
+  else:
+    cnx.commit()
+    rows2 = get_rows(cnx)
+  print('rows after: %d (added %d)' % (rows2, rows2 - rows1))
+  cnx.close()
+
+def update_from_file_public(issue, date, filename, test_mode=False):
+  """
+  Read ILINet data from a zipped CSV and insert into (or update) the database.
```
undefx: (Same here: WHO/NREVSS data)

src/acquisition/fluview/fluview_update.py
```diff
@@ -235,6 +459,16 @@ def main():
     type=int,
     help='issue of the file (e.g. 201740); used iff --file is given'
   )
+  parser.add_argument(
```
undefx: Are these two flags used anywhere? If not, should they be? Or can they be removed?

src/acquisition/fluview/fluview_update.py
stevenabreu7: I'm working on this right now, I noticed that there are some inconsistencies in the data. For state-wise public health labs, data comes per season. For hhs, cen, nat, the data comes weekly as expected. I will commit my proposed way of dealing with that in a bit

src/acquisition/fluview/fluview_update.py
stevenabreu7: I added it in a commit just now. I was thinking about how to handle missing values - I just let be NULL now.

undefx: That's great! This will be helpful. In case you didn't know, there's a measure called "ILI+" that's supposed to be a more precise measure flu prevalence than ILI. It's defined as the product of ILI and percent-positive. Since we already have ILI, the addition of percent-positive will allow us to also compute ILI+.

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3873365/

src/acquisition/fluview/fluview_update.py
```diff
@@ -355,12 +399,13 @@ def update_from_file_public(issue, date, filename, test_mode=False):
   # insert each row
   insert = cnx.cursor()
   for row in entries:
+    lag = delta_epiweeks(row['epiweek'], issue)
```
undefx: Looks like `lag` won't be NULL/None, but the comment above says it'll be NULL for state data. Either way is fine with me (making `lag` `NOT NULL` in database, or setting it to `None` here).

stevenabreu7: I decided to set it to `NOT NULL` in the database! Thanks for all the comments, I really appreciate it!

PR #7, opened 2018-10-24
Enable ght parsing of countries other than US

NLTrieu:
Added optional two-letter country argument to ght_update, as required by the google health trends API.
Nothing should change for the US. Other countries are now saved in the server database with location "XX_" where XX denotes the ISO 1366 alpha-2 country code.

### DIFF
--- a/src/acquisition/ght/ght_update.py
+++ b/src/acquisition/ght/ght_update.py
@@ -75,6 +75,7 @@ from apiclient.discovery import build
 
 # first party
 from .google_health_trends import GHT
+from .google_health_trends import NO_LOCATION_STR
 import delphi.operations.secrets as secrets
 from delphi.utils.epidate import EpiDate
 import delphi.utils.epiweek as flu
@@ -248,7 +249,7 @@ LOCATIONS = [
 ]
 
 
-def update(locations, terms, first=None, last=None):
+def update(locations, terms, first=None, last=None, country='US'):
   # connect to the database
   u, p = secrets.db.epi
   cnx = mysql.connector.connect(user=u, password=p, database='epidata')
@@ -292,7 +293,7 @@ def update(locations, terms, first=None, last=None):
         while True:
           attempt += 1
           try:
-            result = ght.get_data(ew0, ew1, location, term2)
+            result = ght.get_data(ew0, ew1, location, term2, country)
             break
           except Exception as ex:
             if attempt >= 5:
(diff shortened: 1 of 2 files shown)


src/acquisition/ght/ght_update.py
```diff
@@ -305,7 +306,12 @@ def get_num_rows():
         ew = result['start_week']
         num_missing = 0
         for v in values:
-          sql_data = (term, location, ew, v, v)
+          sql_location = location
+          if country != 'US':
+            sql_location = country + "_"
```
undefx: Are you sure you want to add the underscore here? If there is no location (i.e. country code only), then if I'm reading it correctly it'll end up like "MX_" for example. Do you want to add the underscore inside the `if` below so that it's only used when there's a location?

src/acquisition/ght/ght_update.py
```diff
@@ -334,6 +340,7 @@ def main():
   parser.add_argument('term', action='store', type=str, default=None, help='term/query/topic (ex: all; /m/0cycc; "flu fever")')
   parser.add_argument('--first', '-f', default=None, type=int, help='first epiweek override')
   parser.add_argument('--last', '-l', default=None, type=int, help='last epiweek override')
+  parser.add_argument('--country', '-c', default='US', type=string, help='location country (ex: US; BR)')
```
undefx: This should be `str` instead of `string`

src/acquisition/ght/ght_update.py
```diff
@@ -348,15 +355,24 @@ def main():
...
   else:
     terms = [args.term]
 
+  # country argument
+  # Check that country follows ISO 1366 Alpha-2 code. 
+  # See https://www.iso.org/obp/ui/#search.
+  if len(args.country) != 2:
+    raise Exception('country name must be two letters (ISO 1366 Alpha-2)')
+  country = args.country.upper()
+
```
undefx: I recommend also checking that `location` "none" isn't allowed to be used when `country` is "US". (Otherwise I think it'll end up with a query for "US-none"?)

src/acquisition/ght/ght_update.py
NLTrieu: The underscore is meant to distinguish between countries and states, i.e. CA can mean both Canada and California. I added a comment to clarify. 
In the future, maybe we can store countries based on the ISO 1366 alpha-3 code (3 letters for the country), but that would require some look-up table to translate between alpha-2 and alpha-3 codes.

src/acquisition/ght/ght_update.py
NLTrieu: Done, thanks!

src/acquisition/ght/ght_update.py
NLTrieu: Updated to allow users to do both (`location`="none", `country`="US") and (`location`="US", `country`="US").
This only applies to the US country; the second case is enabled for backwards compatibility.

src/acquisition/ght/ght_update.py
undefx: Ok, makes sense.

undefx:
Looks good to me

PR #9, opened 2018-11-19
Update norostat xpath&regexes to handle new data-table.html format

brookslogan:
CDC updated the format of
https://www.cdc.gov/norovirus/reporting/norostat/data-table.html changing
capitalization and wording in the table explanation, breaking the current
scraper. Update the scraper to handle the new format (as well as the old).

Changes to page causing breakage:
- "Suspected and Confirmed Norovirus Outbreaks Reported by State Health
Departments" -> all lower case; handle by being case insensitive for SCNORHD
- "to the" -> "through the"; handle by matching either of these


undefx:
Looks good to me, feel free to merge.

PR #10, opened 2018-11-28
Add different language for wiki data

zhoujunpei:
There are two main modifications. The first modification is adding a value when extracting data from Wikimedia. The second modification is modifying the epidata server PHP file to make it compatible with the language parameter.

There are at least **three important things** to do before merging the code to the master branch!
1. Run the sql command to add a column of `language` with default value set to `en`. (You can refer to the command I wrote in `src/acquisition/wiki/create_wiki_meta_table.sql` and `src/acquisition/wiki/create_wiki_table.sql`, it will not influence the currently running procedure, because if a record is inserted into the table without specifying language, it will be set to `en`).
2. Review the code. I have run all codes locally with my local PHP server, and also make a little bit modifications for cross-platform running. But still not sure if it could work on the Linux server. By the way, I am not very familiar with PHP, so the PHP modification needs to be reviewed carefully.
3. Prepare a roll-back solution which could deal with the failure of this version. (Not sure if there is a testing environment on the server apart from the production environment?)

Thanks!


zhoujunpei:
Another thing is that the client code for R and coffee has not been updated for multi-lanuages wiki data.

PR #16, opened 2018-12-17
Fix norostat_update: use dateModified from meta tag (span is gone)

brookslogan:
Around December 2018, the NoroSTAT data-table page was updated and the old span
describing the modification date was removed. The norostat_update script expects
to be able to extract information from this tag and fails when it is not
present. Luckily, there is currently a meta tag (perhaps added to replace the
span) that serves the same purpose. Check for both types of elements to try to
extract the dateModified. (Prioritize the meta tag over the span.) (This change
should fix the error currently encountered by the norostat_update script.)


brookslogan:
Thanks for looking it over.  Merging in.

PR #17, opened 2019-05-27
Add granular sensor auth tokens

brookslogan:


src/server/api.php
```diff
@@ -34,14 +34,29 @@
...
+  'sensors'        => Secrets::$api['sensors'],
+  'sensor_subsets' => Secrets::$api['sensor_subsets'],
+  'quidel'         => Secrets::$api['quidel'],
+  'norostat'       => Secrets::$api['norostat'],
 );
+$GRANULAR_SENSOR_AUTH_TOKENS = array(
+  'twtr' => array($AUTH['sensor_subsets']['twtr_sensor']),
+  'gft' => array($AUTH['sensor_subsets']['gft_sensor']),
+  'ght' => array($AUTH['sensor_subsets']['ght_sensors']),
+  'ghtj' => array($AUTH['sensor_subsets']['ght_sensors']),
```
undefx: I don't see a corresponding entry in the operations repo for ghtj. Is that one used anywhere? Justin made it a long time ago, but I'm not sure if it has been used since.

src/server/api.php
```diff
@@ -34,14 +34,29 @@
...
+  'cdc'            => Secrets::$api['cdc'],
+  'sensors'        => Secrets::$api['sensors'],
+  'sensor_subsets' => Secrets::$api['sensor_subsets'],
+  'quidel'         => Secrets::$api['quidel'],
+  'norostat'       => Secrets::$api['norostat'],
 );
+$GRANULAR_SENSOR_AUTH_TOKENS = array(
+  'twtr' => array($AUTH['sensor_subsets']['twtr_sensor']),
+  'gft' => array($AUTH['sensor_subsets']['gft_sensor']),
+  'ght' => array($AUTH['sensor_subsets']['ght_sensors']),
```
undefx: Same question here about plural name for ght(j)_sensor**s**

src/server/api.php
```diff
@@ -34,14 +34,29 @@
...
+  'ghtj' => array($AUTH['sensor_subsets']['ght_sensors']),
+  'cdc' => array($AUTH['sensor_subsets']['cdc_sensor']),
+  'quidel' => array($AUTH['sensor_subsets']['quidel_sensor']),
+  'arch' => array($AUTH['sensor_subsets']['arch_sensor']),
+);
+$OPEN_SENSORS = array(
```
undefx: Repeating here for better context: ARCH could probably be public if you wanted. I'm happy either way.

src/server/api.php
```diff
@@ -1009,17 +1024,69 @@ function meta_delphi() {
...
+      if(!array_key_exists('auth', $_REQUEST)) {
+        $auth_tokens_presented = array();
       } else {
-        $data['message'] = 'unauthenticated';
+        $auth_tokens_presented = extract_values($_REQUEST['auth'], 'str');
+      }
+      $names = extract_values($_REQUEST['names'], 'str');
+      $n_names = count($names);
+      $n_auth_tokens_presented = count($auth_tokens_presented);
+      $max_valid_granular_tokens_per_name = max(array_map("count", $GRANULAR_SENSOR_AUTH_TOKENS));
```
undefx: Lack of PHP knowledge on my part here. The [docs](https://www.php.net/manual/en/function.array-map.php) say `array_map` takes a function and an array; but this is a string and an array. Does it know you want to use the `count` function?

src/server/api.php
```diff
@@ -1009,17 +1024,69 @@ function meta_delphi() {
...
+      // The number of valid granular tokens is related to the number of auth token checks that a single query could perform.  Use the max number of valid granular auth tokens per name in the check below as a way to prevent leakage of sensor names via this interface.  Treat all sensors as non-open for convenience of calculation.
+      if($n_names === 0) {
+        // Check whether no names were provided to prevent edge-case issues if surrounding behavior changes in the future:
+        $data['message'] = 'no sensor names provided';
+      } else if($n_auth_tokens_presented !== 1) {
```
undefx: In this case, no auth tokens given, it looks like open sensors won't be served. Am I reading that correctly? If so, is that the behavior you want? I thought you wanted to return open sensors even when an auth token is missing.

src/server/api.php
brookslogan: The 'ght_sensors' token gives access to both GHT-derived sensor sets: ght and ghtj.  I'm not clear if ghtj is being used in the nowcasts yet.

src/server/api.php
brookslogan: Great!  I wasn't sure if this version of ARCH was incorporating non-fluview data, which is why I didn't make it open initially.  Not sure if EPIC requires more thought due to the nature of the input, though.

src/server/api.php
brookslogan: This is uncertain area for me too.  From some examples, it looks like you pass "functionname" for built-ins and normally declared functions, and $functionvar if you have a function pointer.  I need to check on the name resolution rules for the string input.

src/server/api.php
brookslogan: Nice catch!  I was trying to prevent a potential (future) situation where we might return all sensors by default, but broke the open sensor functionality instead.  Some unit tests can guard against that scenario instead.

src/server/api.php
undefx: Ah ok, that makes sense.

src/server/api.php
undefx: IMO, EPIC is safe to share because it's an aggregate; no individual forecasts are shared. I'm happy to leave the decision to you and Roni.

src/server/api.php
undefx: Ok, good to know, thanks.

src/server/api.php
brookslogan: Fixed this and added some tests.  (Plus fixed naming: quidel->quid, and added missing wiki sensor key.)

PR #20, opened 2019-06-10
Update deploy.json to include python server code

brookslogan:


brookslogan:
Should fix breaking deploy.

PR #21, opened 2019-06-10
Add simulate_api_response.py

brookslogan:


brookslogan:
Also needed for fixing deployment.

PR #22, opened 2019-06-10
Fix simulate_api_response errors on deploy testing

brookslogan:


brookslogan:
This probably doesn't address the problem finding api.php during deploy; closing.

PR #24, opened 2019-06-21
Add wiki language parameter to rest of clients

brookslogan:
The language parameter in the wiki data source is already supported in the
python client. Add support for this parameter in the R, coffeescript, and js
clients.

Also, in R client, perform some checks to try to detect some alternative date
formats and generate errors or automatically convert to the expected format.


krivard:

PR #28, opened 2019-09-10
modify afhsb table structures

mengze-irene:
1. Merged two data dumps. Added an issue column to indicate which dump a record is from. 
2. Allow an optional query argument 'criterion', which specifies which aggregation criterion the user wants to view the data by.


krivard:

PR #55, opened 2020-04-19
Add covidcast and covidcast_meta endpoints to clients

pedritom-amzn:
add `covidcast` and `covidcast_meta` endpoints to clients, add integration tests for the python changes, and add documentation to the corresponding api docs. also some other minor fixes.

testing: the new integration tests pass.

### DIFF
--- a/docs/api/covidcast.md
+++ b/docs/api/covidcast.md
@@ -77,7 +77,7 @@ See [this documentation](README.md) for details on specifying epiweeks, dates, a
 
 # Example URLs
 
-### Delphi's COVID-19 Surveillance Streams from Facebook Survey ILI on 2020-04-06 - 2010-04-10 (county 06001)
+### Delphi's COVID-19 Surveillance Streams from Facebook Survey ILI on 2020-04-06 - 2020-04-10 (county 06001)
 https://delphi.midas.cs.cmu.edu/epidata/api.php?source=covidcast&data_source=fb_survey&signal=ili&time_type=day&geo_type=county&time_values=20200406-20200410&geo_value=06001
 ```json
 {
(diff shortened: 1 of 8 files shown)


pedritom-amzn:
> quick question: how do you run the new test? (very happy to see testing of course!)

i followed the instructions in https://github.com/cmu-delphi/delphi-epidata/blob/master/docs/epidata_development.md and ran
```
docker run --rm delphi_python python3 -m undefx.py3tester.py3tester --color \
  repos/delphi/delphi-epidata/integrations
```
btw, the other integration tests failed for me for some reason, but i haven't looked deeply into the reasons yet -- the failures didn't seem related to my change.

undefx:
got it, thanks.

the other integration tests will fail if database and web images aren't started up prior to testing. i assume this is the case since you didn't mention them, but correct me if i'm wrong. no big deal, i agree this PR shouldn't have any effect on those tests, which are all on the server side.

if/when i ever get some free time, i might make your new test hit the local docker web server instead of the prod api just to provide some isolation and hopefully make things more self-contained. no action needed from you, just putting this down for posterity and whatnot.

thanks again!

pedritom-amzn:
> the other integration tests will fail if database and web images aren't started up prior to testing. i assume this is the case since you didn't mention them, but correct me if i'm wrong.

yeah, it's strange because both of them are running, but yes, the error i'm getting is that it can't connect to the database. i'll dig into it some more.

> if/when i ever get some free time, i might make your new test hit the local docker web server instead of the prod api just to provide some isolation and hopefully make things more self-contained. no action needed from you, just putting this down for posterity and whatnot.

yeah, i was thinking about that too. since `BASE_URL` is hard-coded in the client files, i decided to leave as-is for now to get the tests out, at the risk of the tests failing if prod is down. i'll create an issue to track the isolation task.

undefx:
Noticed that there's no network flag in your `docker run` command. This could explain why the connection failed. Here's the full command:

```
docker run --rm --network delphi-net delphi_python \
python3 -m undefx.py3tester.py3tester --color \
  repos/delphi/delphi-epidata/integrations
```

Thanks, filing an issue sounds great so we don't lose track of this!

pedritom-amzn:
you're absolutely right! added the network parameter and all the tests succeeded. thanks!

opened [issue #56](https://github.com/cmu-delphi/delphi-epidata/issues/56) to track isolating the tests.

PR #69, opened 2020-04-20
formatting and update client integration test

### LINKED ISSUE #56: make client integration tests hit local docker web server
**problem:** the python client integration tests currently hit the prod api since the endpoint is hardcoded into the client.

**impact:** lack of isolation. the success of the tests depend on whether prod is up and working. also, the tests unnecessarily add test load and risk to prod.

**proposal:** make client integration tests hit the local docker web server instead of the prod api so as to provide some isolation and make things more self-contained.

dfarrow0:
fixes #56

### DIFF
--- a/docs/api/covidcast.md
+++ b/docs/api/covidcast.md
@@ -10,7 +10,34 @@ General topics not specific to any particular data source are discussed in the
 
 ## Delphi's COVID-19 Surveillance Streams Data
 
-... <!-- TODO -->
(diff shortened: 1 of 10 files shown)


undefx:
@pedritom-amzn FYI

let me know if you don't like the client integration test changes and we can iterate

pedritom-amzn:
looks good to me. thanks!

PR #70, opened 2020-04-20
covidcast acquisition

dfarrow0:
- the full implementation is split into multiple smaller commits for easier review
- should be trivial to automate once deployed

### DIFF
--- a/integrations/acquisition/covidcast/test_csv_uploading.py
+++ b/integrations/acquisition/covidcast/test_csv_uploading.py
@@ -1,0 +1,122 @@
+"""Integration tests for covidcast's CSV-to-database uploading."""
+
+# standard library
+import os
+import unittest
+from unittest.mock import MagicMock
+
+# third party
+import mysql.connector
+import requests
+
+# first party
+from delphi.epidata.client.delphi_epidata import Epidata
+import delphi.operations.secrets as secrets
+
+# py3tester coverage target (equivalent to `import *`)
+__test_target__ = 'delphi.epidata.acquisition.covidcast.csv_to_database'
+
+
+class CsvUploadingTests(unittest.TestCase):
+  """Tests covidcast CSV uploading."""
+
+  def setUp(self):
+    """Perform per-test setup."""
+
+    # connect to the `epidata` database and clear the `covidcast` table
+    cnx = mysql.connector.connect(
+        user='user',
+        password='pass',
+        host='delphi_database_epidata',
+        database='epidata')
+    cur = cnx.cursor()
+    cur.execute('truncate table covidcast')
+    cnx.commit()
+    cur.close()
+
+    # make connection and cursor available to test cases
+    self.cnx = cnx
+    self.cur = cnx.cursor()
+
+    # use the local instance of the epidata database
+    secrets.db.host = 'delphi_database_epidata'
+    secrets.db.epi = ('user', 'pass')
+
+    # use the local instance of the Epidata API
+    Epidata.BASE_URL = 'http://delphi_web_epidata/epidata/api.php'
+
+  def tearDown(self):
+    """Perform per-test teardown."""
+    self.cur.close()
+    self.cnx.close()
+
+  def test_uploading(self):
+    """Scan, parse, upload, archive, serve, and fetch a covidcast signal."""
+
+    # make some fake data files
+    data_dir = 'covid/data'
+    source_receiving_dir = data_dir + '/receiving/src-name'
+    os.makedirs(source_receiving_dir, exist_ok=True)
+
+    # valid
+    with open(source_receiving_dir + '/20200419_state_test.csv', 'w') as f:
+      f.write('geo_id,val,se,sample_size,direction\n')
+      f.write('ca,1,0.1,10,1\n')
+      f.write('tx,2,0.2,20,NA\n')
+      f.write('fl,3,0.3,30,-1\n')
+
+    # invalid
+    with open(source_receiving_dir + '/20200420_state_test.csv', 'w') as f:
+      f.write('this,header,is,wrong\n')
+
+    # invalid
+    with open(source_receiving_dir + '/hello.csv', 'w') as f:
+      f.write('file name is wrong\n')
+
+    # upload CSVs
+    args = MagicMock(data_dir=data_dir, test=False)
+    main(args)
+
+    # request CSV data from the API
+    response = Epidata.covidcast(
+        'src-name', 'test', 'day', 'state', 20200419, '*')
+
+    # verify data matches the CSV
+    self.assertEqual(response, {
+      'result': 1,
+      'epidata': [
+        {
+          'time_value': 20200419,
+          'geo_value': 'ca',
+          'value': 1,
+          'stderr': 0.1,
+          'sample_size': 10,
+          'direction': 1,
+        },
+        {
+          'time_value': 20200419,
+          'geo_value': 'fl',
+          'value': 3,
+          'stderr': 0.3,
+          'sample_size': 30,
+          'direction': -1,
+        },
+        {
+          'time_value': 20200419,
+          'geo_value': 'tx',
+          'value': 2,
+          'stderr': 0.2,
+          'sample_size': 20,
+          'direction': None,
+        },
+       ],
+      'message': 'success',
+    })
+
+    # verify that the CSVs were archived
+    path = data_dir + '/archive/successful/src-name/20200419_state_test.csv.gz'
+    self.assertIsNotNone(os.stat(path))
+    path = data_dir + '/archive/failed/src-name/20200420_state_test.csv'
+    self.assertIsNotNone(os.stat(path))
+    path = data_dir + '/archive/failed/unknown/hello.csv'
+    self.assertIsNotNone(os.stat(path))
--- a/integrations/acquisition/covidcast/test_csv_uploading.py
+++ b/integrations/acquisition/covidcast/test_csv_uploading.py
@@ -7,7 +7,6 @@ from unittest.mock import MagicMock
 
 # third party
 import mysql.connector
-import requests
 
 # first party
 from delphi.epidata.client.delphi_epidata import Epidata
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -1,0 +1,242 @@
+"""Collects and reads covidcast data from a set of local CSV files."""
+
+# standard library
+import glob
+import os
+import re
+
+# third party
+import pandas
+
+
+class CsvImporter:
+  """Finds and parses covidcast CSV files."""
+
+  # .../source/yyyymmdd_geo_signal.csv
+  PATTERN_DAILY = re.compile(r'^.*/([^/]*)/(\d{8})_(\w+?)_(\w+)\.csv$')
+
+  # .../source/weekly_yyyyww_geo_signal.csv
+  PATTERN_WEEKLY = re.compile(r'^.*/([^/]*)/weekly_(\d{6})_(\w+?)_(\w+)\.csv$')
+
+  # set of allowed resolutions (aka "geo_type")
+  GEOGRAPHIC_RESOLUTIONS = {'county', 'hrr', 'msa', 'dma', 'state'}
+
+  # set of required CSV columns
+  REQUIRED_COLUMNS = {'geo_id', 'val', 'se', 'direction', 'sample_size'}
+
+  # NOTE: this should be a Python 3.7+ `dataclass`, but the server is on 3.4
+  # See https://docs.python.org/3/library/dataclasses.html
+  class RowValues:
+    """A container for the values of a single covidcast row."""
+
+    def __init__(self, geo_value, value, stderr, sample_size, direction):
+      self.geo_value = geo_value
+      self.value = value
+      self.stderr = stderr
+      self.sample_size = sample_size
+      self.direction = direction
+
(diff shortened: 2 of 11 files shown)


dfarrow0:
@pedritom-amzn I'd would appreciate any comments!

I'm hoping to deploy this soon, so feel free to add comments even after merge.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -0,0 +1,242 @@
...
+      self.sample_size = sample_size
+      self.direction = direction
+
+  @staticmethod
+  def is_sane_day(value):
+    """Return whether `value` is a sane (maybe not valid) YYYYMMDD date."""
+
+    year, month, day = value // 10000, (value % 10000) // 100, value % 100
+
+    nearby_year = 2014 <= year <= 2023
```
pedritom-amzn, line 45: why such a low value for the max year?

in either case, perhaps put these year boundary values as constants near the top of the file where they are more visible and easier to maintain if needed? similar for `is_sane_week()`.

dfarrow0: a couple of the data sources have historical data that extends back 5+ years. it wasn't clear to me how much historical data we'd want to support. but on second thought, i'm inclined to agree that 2014 is too early with respect to the pandemic. i'll bump that up to, say, 2019.

good feedback, i will extract out those constants.

thanks for looking :)

integrations/acquisition/covidcast/test_csv_uploading.py
```diff
@@ -0,0 +1,122 @@
+"""Integration tests for covidcast's CSV-to-database uploading."""
+
+# standard library
+import os
+import unittest
+from unittest.mock import MagicMock
+
+# third party
+import mysql.connector
+import requests
```
dfarrow0, line 10: note to self: unused, can remove

src/acquisition/covidcast/csv_importer.py
```diff
@@ -0,0 +1,242 @@
...
+    value = float(row.val)
+
+    # optional nonnegative float
+    stderr = CsvImporter.maybe_apply(float, row.se)
+    if stderr is not None and stderr < 0:
+      return (None, 'se')
+
+    # optional not-too-small float
+    sample_size = CsvImporter.maybe_apply(float, row.sample_size)
+    if sample_size is not None and sample_size < 5:
```
pedritom-amzn, line 198: min sample size would be another i would recommend using a top-level constant for, and perhaps a comment as to why.

dfarrow0:
addressed all feedback

integrations/acquisition/covidcast/test_csv_uploading.py
dfarrow0: done

src/acquisition/covidcast/csv_importer.py
dfarrow0: done

src/acquisition/covidcast/csv_importer.py
dfarrow0: done

src/acquisition/covidcast/file_archiver.py
```diff
@@ -0,0 +1,57 @@
...
+
+    WARNING: This is a potentially destructive operation. See details below.
+
+    path_src: the directory which contains the file to be archived
+    path_dst: the directory into which the file should be moved
+    filename: the name of the file within `path_src`
+    compress: gzips the file if true, otherise moves the file unmodified
+
+    The destination directory will be created if necessary. If the destination
+    file already exists, it will be overwritten.
```
pedritom-amzn, line 32: do you want to perhaps at least log when this happens since it might indicate something unusual going on?

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -0,0 +1,127 @@
...
+          row_values.direction)
+
+    # archive the current file based on validation results
+    if all_rows_valid:
+      archive_as_successful(path_src, filename, source)
+    else:
+      archive_as_failed(path_src, filename, source)
+
+
+def main(
```
pedritom-amzn, line 106: is there a possibility of multiple runs happening simultaneously and, if so, is there a risk of race conditions? i haven't managed to come up with a scenario that would result in data loss, but i can at least imagine (unlikely) scenarios where files get moved to the wrong archive.

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -0,0 +1,127 @@
...
+  receiving_dir = os.path.join(data_dir, 'receiving')
+  archive_successful_dir = os.path.join(data_dir, 'archive', 'successful')
+  archive_failed_dir = os.path.join(data_dir, 'archive', 'failed')
+
+  # helper to archive a failed file without compression
+  def archive_as_failed(path_src, filename, source):
+    print('archiving as failed')
+    path_dst = os.path.join(archive_failed_dir, source)
+    compress = False
+    file_archiver_impl.archive_file(path_src, path_dst, filename, compress)
```
pedritom-amzn, line 55: just want to note that if there is a problem upstream that keeps sending a corrupted version of the same file, this will only keep the most recent, potentially resulting in some loss of debugging information. however, this may be by design since you're at least logging each time a file import fails, so you may not need the extra content info.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -0,0 +1,242 @@
...
+    valid, details is a tuple of (source, signal, time_type, geo_type,
+    time_value) (otherwise None).
+    """
+
+    for path in glob.glob(os.path.join(scan_dir, '*', '*')):
+      if not path.lower().endswith('.csv'):
+        # safe to ignore this file
+        continue
+
+      print('file:', path)
```
krivard, line 86: Double-checking that a print() is what is wanted here to collect error output (fine if it is, just making sure some part of the larger system will be saving this output to disk)

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -0,0 +1,127 @@
...
+
+  # helper to archive a successful file with compression
+  def archive_as_successful(path_src, filename, source):
+    print('archiving as successful')
+    path_dst = os.path.join(archive_successful_dir, source)
+    compress = True
+    file_archiver_impl.archive_file(path_src, path_dst, filename, compress)
```
pedritom-amzn, line 62: i do worry a little bit more about the potential silent overwrite in this scenario. for example, if you somehow get a truncated version of a file that was already ingested, having both versions archived could be useful for both determining the extent of the problem and for quickly back-filling.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -0,0 +1,242 @@
...
+        yield (path, None)
+        continue
+
+      # extract and validate time resolution
+      if daily_match:
+        time_type = 'day'
+        time_value = int(daily_match.group(2))
+        match = daily_match
+        if not CsvImporter.is_sane_day(time_value):
+          print(' invalid filename day')
```
krivard, line 102: Include the `time_value` parsed here, since if this goes wrong inappropriately that's the first thing we'll want to know

src/acquisition/covidcast/csv_to_database.py
dfarrow0: good point!

concurrent runs aren't anticipated and from the standpoint of this code are considered unsupported.

more generally, delphi's automation is single-process, single-thread, and so a race condition here _shouldn't_ be a risk.

concurrent runs of this program would be problematic, however it is assumed to be well-understood that this is the case.

src/acquisition/covidcast/csv_to_database.py
dfarrow0: fair point. my opinion is that the added complexity of unique naming isn't clearly (to me) justified by the benefit to debugging. (i.e. i took the path of least resistance, but i think it's ok in this case)

src/acquisition/covidcast/file_archiver.py
dfarrow0: done (and added new test)

src/acquisition/covidcast/csv_importer.py
```diff
@@ -0,0 +1,242 @@
...
+
+    else:
+      return (None, 'geo_type')
+
+    # required float
+    value = float(row.val)
+
+    # optional nonnegative float
+    stderr = CsvImporter.maybe_apply(float, row.se)
+    if stderr is not None and stderr < 0:
```
krivard, line 201: We're permitting se=0? I might be out of date, but I thought the spec excluded that..

src/acquisition/covidcast/csv_importer.py
dfarrow0: long-term i strongly advise upgrading to a "real" logging system. short-term, this is standard procedure. these messages will be stored to disk as part of automation's log file.

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -0,0 +1,218 @@
...
+      (glob_paths[0], ('fb_survey', 'cli', 'week', 'county', 202015)),
+      (glob_paths[1], ('ght', 'rawsearch', 'day', 'state', 20200408)),
+      (glob_paths[2], None),
+      (glob_paths[3], None),
+      (glob_paths[4], None),
+      (glob_paths[5], None),
+    ]
+    self.assertEqual(found, expected)
+
+  def test_is_header_valid_allows_extra_columns(self):
```
pedritom-amzn, line 72: maybe also a test that column order doesn't matter? but maybe it's not necessary if the risk of someone switching to a non-set-based implementation later is low.

src/acquisition/covidcast/csv_to_database.py
dfarrow0: yeah, completely agree. could i ask a favor of filing an issue for this? probably worth fixing in the long-run, but delphi needs something functional asap.

src/acquisition/covidcast/csv_importer.py
dfarrow0: done

src/acquisition/covidcast/data_dir_readme.md
```diff
@@ -0,0 +1,35 @@
+# covidcast data directory
+
+NOTE: Edit this file on GitHub!
```
krivard, line 3: Is this a lingering TODO or some weird thing where editing locally is ill-advised?

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -0,0 +1,218 @@
...
+      row = MagicMock(
+          geo_id=geo_id,
+          val=val,
+          se=se,
+          sample_size=sample_size,
+          direction=direction)
+      return geo_type, row
+
+    # cases to test each failure mode
+    failure_cases = [
```
pedritom-amzn, line 118: maybe also add cases for where non-optional fields are missing?

src/acquisition/covidcast/csv_importer.py
dfarrow0: i think the conclusion was that stderr can be zero when the sampling distribution allows. context was around the binomial distribution in particular which has standard error as a function of $p$. in my understanding, when $p \in {0, 1}$, stderr is zero.

tests/acquisition/covidcast/test_csv_importer.py
dfarrow0: done

src/acquisition/covidcast/data_dir_readme.md
dfarrow0: this file gets deployed to `/common/covidcast/README.md`. the note is there to prevent someone from editing that file directly on the server, only to have their changes clobbered the next time this repo gets deployed.

src/acquisition/covidcast/csv_importer.py
pedritom-amzn: thanks! i hadn't thought of the historical data scenario and was more concerned about the max date and this code being useful beyond 2023, but with the constants it should be easy/safe to extend in either direction if the need arises in the future.

src/acquisition/covidcast/data_dir_readme.md
dfarrow0: to clarify: editing locally is no problem so long as the change is committed to the repo

src/acquisition/covidcast/csv_to_database.py
pedritom-amzn: will do.

pedritom-amzn:
looks good to me. none of my feedback is blocking.

tests/acquisition/covidcast/test_csv_importer.py
dfarrow0: done

src/acquisition/covidcast/csv_importer.py
dfarrow0: oh i see, i misread your original comment. as you said, this should be easy to extend, but maybe it's better to just extend it now per the original comment. moved max year up to 2030.

dfarrow0:
@pedritom-amzn and @krivard 

thanks for taking the time to look closely and leave thoughtful feedback!

all comments addressed. unit and integration tests pass.

src/acquisition/covidcast/csv_to_database.py
pedritom-amzn: https://github.com/cmu-delphi/delphi-epidata/issues/73

PR #75, opened 2020-04-21
covidcast direction calculation

dfarrow0:
- just introducing a helper class, which is not used yet
- tests included, 100% coverage

### DIFF
--- a/src/acquisition/covidcast/direction.py
+++ b/src/acquisition/covidcast/direction.py
@@ -36,13 +36,13 @@ class Direction:
     if n < 0:
       raise ValueError('n must be non-negative')
 
+    if len(x) < Direction.MIN_SAMPLE_SIZE:
+      return None
+
     # check for coincident values in a way that is robust to extremely small
     # differences
     if numpy.isclose(min(numpy.diff(sorted(x))), 0):
       raise ValueError('x contains coincident values')
-
-    if len(x) < Direction.MIN_SAMPLE_SIZE:
-      return None
 
     fit = scipy.stats.linregress(x, y)
 
--- a/src/acquisition/covidcast/direction.py
+++ b/src/acquisition/covidcast/direction.py
@@ -1,0 +1,47 @@
+"""Classifies the direction of a signal based on its recent trajectory."""
+
+# third party
+import numpy
+import scipy.stats
+
+
+class Direction:
+  """Utility class for computing signal direction."""
+
+  # minimum number of points required to compute direction
+  MIN_SAMPLE_SIZE = 3
+
+  @staticmethod
+  def get_direction(x, y, n=1):
+    """Return the direction of the given time-series.
+
+    `x`: list of time offsets (e.g. 0, 1, 2, ...)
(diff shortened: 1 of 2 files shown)


dfarrow0:
requesting review also from @capnrefsmmat

src/acquisition/covidcast/direction.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    if len(x) != len(y):
+      raise ValueError('x and y must have same length')
+
+    if n < 0:
+      raise ValueError('n must be non-negative')
+
+    if len(x) < Direction.MIN_SAMPLE_SIZE:
+      return None
```
capnrefsmmat, line 40: I guess a paranoid thing you could do would be to ensure the number of unique values of `x` is greater than `Direction.MIN_SAMPLE_SIZE`; this would catch any slip-ups where the wrong days are being given or they're duplicated.

If there are not enough unique values to estimate the line and standard error, I'd expect SciPy to throw an exception too

src/acquisition/covidcast/direction.py
```diff
@@ -0,0 +1,47 @@
...
+
+    fit = scipy.stats.linregress(x, y)
+
+    if numpy.abs(fit.slope) <= n * fit.stderr:
+      return 0
+    else:
+      return numpy.sign(fit.slope)
```
capnrefsmmat, line 47: This looks exactly right to me

src/acquisition/covidcast/direction.py
```diff
@@ -0,0 +1,47 @@
+"""Classifies the direction of a signal based on its recent trajectory."""
+
+# third party
+import numpy
```
krivard, line 4: if numpy isn't used anywhere else, consider replacing use of `numpy.abs` with python base `abs` and `numpy.sign` with a conditional

krivard: Peace of mind about the implementation of `numpy.sign` might be worth it though.

dfarrow0: i agree it is preferable to use built-in `abs`. done.

however, what would be a benefit to reimplementing `sign`? if it's strictly a matter of shedding a dependency, note that scipy itself depends on numpy. this is confirmed by the timing report in the unit test. in particular, the incremental cost of importing numpy after scipy is negligible.

```
    4 import numpy                                                                     1x         163 ms     163 ms    
    5 import scipy.stats                                                               1x         270 ms     270 ms    
```

```
    4 import scipy.stats                                                               1x         437 ms     437 ms    
    5 import numpy                                                                     1x                              
```

src/acquisition/covidcast/direction.py
dfarrow0: i was feeling paranoid about this, thanks for confirming!

src/acquisition/covidcast/direction.py
dfarrow0: oh yes, great catch. done.

dfarrow0:
@capnrefsmmat i added check for coincident points

PR #77, opened 2020-04-22
covidcast automatic direction updater

dfarrow0:
- updates direction for all _daily_ covidcast rows
- new unit and integration tests
- all new and existing tests pass


dfarrow0:
@capnrefsmmat FYI

PR #78, opened 2020-04-22
faster direction for covidcast

dfarrow0:
avoid quadratic time at the database by scanning entire time-series in linear time in python

### DIFF
--- a/src/acquisition/covidcast/direction.py
+++ b/src/acquisition/covidcast/direction.py
@@ -1,7 +1,7 @@
 """Classifies the direction of a signal based on its recent trajectory."""
 
 # third party
-import numpy
+import numpy as np
 import scipy.stats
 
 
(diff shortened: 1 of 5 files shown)


dfarrow0:
@capnrefsmmat please take a look

src/acquisition/covidcast/direction.py
```diff
@@ -50,4 +50,39 @@ def get_direction(x, y, n=1):
       return 0
     else:
       # return an integer in {-1, +1}
-      return int(numpy.sign(fit.slope))
+      return int(np.sign(fit.slope))
+
+  @staticmethod
+  def scan_timeseries(offsets, row_days, values, timestamp1s, timestamp2s):
```
dfarrow0, line 56: @capnrefsmmat i assume you care most about this part

dfarrow0:
back from the future to report the results: 502573 directions computed in 5:05 minutes (down from 3-4 hours to do roughly the same).

as a quick sanity check, all 59434 directions called on days 20200417+ were the same as what was previously computed using the slow method.

i still owe unit tests.

dfarrow0:
@krivard thanks for taking a look and for the comment

capnrefsmmat:
Logic looks good to me. I also don't see a way to simplify the logic any, but it's already quite clear and simple. And 1,700 directions calculated per second isn't so bad.

PR #79, opened 2020-04-22
covidcast_meta optional caching

dfarrow0:
- use a short-lived cache to make `covidcast_meta` super fast
- integration tested
- all unit and integration tests pass
- i owe unit tests for this


dfarrow0:
@capnrefsmmat please take a look

PR #104, opened 2020-05-08
Remove garbage characters in get_norostat query

brookslogan:
Low priority for review as this is currently auth-gated from public use.

Can test with Epidata$norostat(auth="(auth token)","Minnesota, Ohio, Oregon, Tennessee, and Wisconsin",epiweeks=Epidata$range(100052,300052)).


krivard:

PR #115, opened 2020-05-27
Update to 1.2.1 combination signal, and fix map table

capnrefsmmat:
Fixes the API docs for the updated combination signal, and fixes the table showing the signals included in the map.

Will merge only after the 1.2.1 site update goes live.

### DIFF
--- a/docs/api/covidcast_signals.md
+++ b/docs/api/covidcast_signals.md
@@ -158,14 +158,20 @@ filtering, smoothing, or changes.
 This source provides signals which are statistical combinations of the other
 sources above. It is not a primary data source.
 
-* `nmf_day_doc_fbs_ght`: This signal uses a rank-1 approximation, from a
+* `nmf_day_doc_fbc_fbs_ght`: This signal uses a rank-1 approximation, from a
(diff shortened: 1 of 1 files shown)


docs/api/covidcast_signals.md
```diff
@@ -158,14 +158,20 @@ filtering, smoothing, or changes.
...
+  (`smoothed_hh_cmnty_cli`), and Search Trends (`smoothed_search`) indicators.
+  It does not include official reports (cases and deaths from the `jhu-csse`
+  source). Higher values of the combined signal correspond to higher values of
+  the other indicators, but the scale (units) of the combination is arbitrary.
+  Note that the Search Trends source is not available at the county level, so
+  county values of this signal do not use it.
+* `nmf_day_doc_fbs_ght`: This signal is calculated in the same way as
+  `nmf_day_doc_fbc_fbs_ght`, but does *not* include the Symptoms in Community
+  survey signal, which was not available at the time this signal was introduced.
+  This signal is deprecated and is no longer updated as of May 28, 2020.
```
krivard, line 174: Excellent solution here, I was wondering about that

PR #117, opened 2020-05-31
List data changes from cmu-delphi/covidcast-indicators#47 in a changelog

krivard:
* Creates a new page to hold the changelog
* Prototypes a changelog format that balances completeness with concision
* Adds references to changelog in each source section of `_signals.md`

We may want to pick a different anchor date, particularly for Quidel and Google Surveys.

### DIFF
--- a/docs/api/covidcast_changelog.md
+++ b/docs/api/covidcast_changelog.md
@@ -6,7 +6,6 @@ When Delphi makes substantial changes to the computation of a signal, we will ty
 ### `doctor-visits`
 ### `fb-survey`
 
-<a name="fb-survey-v1.3"></a> 
 #### 3 June 2020: v1.3
 
 Duplicate survey weights had corrupted historical figures for the following signals and dates. The correct data has been restored to the API.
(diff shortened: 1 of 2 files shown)


docs/api/covidcast_changelog.md
```diff
@@ -0,0 +1,28 @@
+# Delphi's COVID-19 Surveillance Streams Changelog
+
+When Delphi makes substantial changes to the computation of a signal, we will typically publish it under a new name rather than revising the existing signal. By contrast, the changes documented here are bug fixes and pipeline repairs.
+
+## Sources and Signals
+### `doctor-visits`
+### `fb-survey`
+
+<a name="fb-survey-v1.3"></a> 
+#### 3 June 2020: v1.3
```
capnrefsmmat, line 10: In theory, this heading already has an anchor with id `3-june-2020-v1.3` or something like that.

I think, though, that it would be easier for each source section in `covidcast_signals.md` to link to the corresponding section in `covidcast_changelog.md`, rather than the section for the most recent update. So you'd only have to link to `covidcast_changelog.md#fb-survey`, where the user would immediately see the latest changes. That would make updates easier.

krivard: Done.

PR #122, opened 2020-06-10
Add columns to covidcast for issue and lag.

krivard:
* Database schema change in ddl (Brian please double-check)
* Supporting infra in acquisition
* Extend all data insertion test cases to include toy issue and lag
* Test that we can insert a new issue for an extant (source,signal,date,geo)

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -5,10 +5,13 @@ import math
 import glob
 import os
 import re
+from datetime import date
 
 # third party
 import pandas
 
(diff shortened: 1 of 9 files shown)


src/acquisition/covidcast/csv_importer.py
```diff
@@ -78,6 +83,12 @@ def find_csv_files(scan_dir, glob=glob):
     time_value) (otherwise None).
```
undefx, line 83: suggest updating this to reflect the new `issue_value` and `lag_value` values returned

PR #126, opened 2020-06-12
Remove minimum sample size requirement.

krivard:
Indicators will do their own sample size validation from here on out.

This is necessary for the safegraph indicator, which uses a special definition of sample size (requested by forecasting) that is not based on individual users but on clusters.

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -229,8 +229,6 @@ class CsvImporter:
     except ValueError:
       # expected a number, but got a string
       return (None, 'sample_size')
-    if sample_size is not None and sample_size < CsvImporter.MIN_SAMPLE_SIZE:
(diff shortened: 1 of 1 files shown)


src/acquisition/covidcast/csv_importer.py
```diff
@@ -229,8 +229,6 @@ def extract_and_check_row(row, geo_type):
     except ValueError:
       # expected a number, but got a string
       return (None, 'sample_size')
-    if sample_size is not None and sample_size < CsvImporter.MIN_SAMPLE_SIZE:
-      return (None, 'sample_size')
```
capnrefsmmat, line 233: Probably can also remove the definition of `MIN_SAMPLE_SIZE` at the top of the file

capnrefsmmat:
Looks fine modulo eliminating dead constants

PR #128, opened 2020-06-13
Remove minimum sample size test; include unit tests in dev guide

krivard:
The deploy of #126 failed because of failing unit tests. Unit tests were not included in the developer guide, so I've added a section describing how to run them. The failing test has been fixed (we no longer want acquisition to fail on a sample size < 5), and all tests pass.

### DIFF
--- a/docs/epidata_development.md
+++ b/docs/epidata_development.md
@@ -106,7 +106,29 @@ docker run --rm -p 10080:80 \
   delphi_web_epidata
 ```
 
-## manual
+## unit tests
+
+Once the server containers are running, you can run unit tests.
+
+First, build the `delphi_python` image per the
+  [backend development guide](https://github.com/cmu-delphi/operations/blob/master/docs/backend_development.md#creating-an-image).
+  Your test sources will live in, and be executed from within, this image.
+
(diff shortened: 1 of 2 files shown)


docs/epidata_development.md
```diff
@@ -106,7 +106,29 @@ docker run --rm -p 10080:80 \
...
+Once the server containers are running, you can run unit tests.
+
+First, build the `delphi_python` image per the
+  [backend development guide](https://github.com/cmu-delphi/operations/blob/master/docs/backend_development.md#creating-an-image).
+  Your test sources will live in, and be executed from within, this image.
+
+Then run the test container:
+
+  ```bash
+  docker run --rm --network delphi-net delphi_python \
```
capnrefsmmat, line 120: Does the `delphi-net` network have to be created before this can be run? I don't see that in the backend development docs

krivard: It's not in the backend docs, but it is in the frontend ones, and is explicitly called out at the top of this guide: https://github.com/cmu-delphi/delphi-epidata/blob/master/docs/epidata_development.md#epidata-api-development-guide

PR #133, opened 2020-06-17
Feature/optimize direction computation

AlSaeed:
Performance Profiling of the existing implementation and optimized implementation. This is derived by average time taken over 5 runs on a table containing 1000 randomly chosen time-series (all stale with `direction==NULL` and `timestamp2==0`). On average these tables contained 90054.6 rows.

| Implementation | Retrieval Queries | Updating `direction` Queries | Updating `timestamp2` Queries | Direction Computation | Other  | Total Time |
|----------------|-------------------|------------------------------|-------------------------------|-----------------------|--------|------------|
| Existing       |            0.96 s |                      21.50 s |                       1.019 s |               18.91 s | 0.30 s | 42.69 s    |
| Optimized      |            3.08 s |                       1.83 s |                        1.14 s |               20.11 s | 7.09 s | 33.24 s    |

Remaining Points:
- Adding unit and integration tests for the optimized implementation.
- Reporting performance for the whole dataset (~12M rows).
- Probably `get_data_stdev_across_locations` method in Database, should be updated to only account for rows of latest issue date for any (time-series key, time_value) combination?

Around 60% of the time in the setting profiled is spent in computing the direction, so probably this where we should look next for performance gain.

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,0 @@
-<component name="ProjectCodeStyleConfiguration">
-  <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
-  </state>
-</component>
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -162,6 +162,102 @@ class Database:
 
     self._cursor.execute(sql, args)
 
+  def get_all_record_values_of_timeseries_with_potentially_stale_direction(self, temporary_table=None):
+    """
+    [TODO]
+    """
+
+    stale_ts_key_sql = '''
+      SELECT
+        `source`,
+        `signal`,
+        `time_type`,
+        `geo_type`,
+        `geo_value`
+      FROM
+        `covidcast`
+      WHERE
+        `time_type` = 'day'
+      GROUP BY
+        `source`,
+        `signal`,
+        `time_type`,
+        `geo_type`,
+        `geo_value`
+      HAVING
+        MAX(`timestamp1`) > MIN(`timestamp2`)
+    '''
+
+    stale_ts_record_keys_with_latest_issues_sql = f'''
+      SELECT
+        `source`,
+        `signal`,
+        `time_type`,
+        `geo_type`,
+        `geo_value`,
+        `time_value`,
+        MAX(`issue`) AS `issue`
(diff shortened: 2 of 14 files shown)


AlSaeed:
This is the time measurements for the entire dataset (12786416 Rows, 139887 time-series):
| Implementation | Retrieval Queries | Updating `direction` Queries | Updating `timestamp2` Queries | Direction Computation | Other  | Total Time |
|----------------|-------------------|------------------------------|-------------------------------|-----------------------|--------|------------|
| Existing       |            317.59 s |                      3600.79 s |                       152.56 s |               2834.59 s | 44.82 s | 6950.36 s    |
| Optimized      |            868.32 s |                       341.64 s |                        267.12 s |               2785.32 s | 859.63 s | 5122.02 s    |

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
krivard, line 1: (IDE configuration files like this should not be added to the repository)

src/acquisition/covidcast/database.py
```diff
@@ -158,6 +163,179 @@ def update_direction(
...
+      SELECT *
+      FROM ({stale_ts_records_with_latest_issues_sql}) AS c;
+      '''
+    self._cursor.execute(final_sql)
+    self._cursor.execute(f"SELECT * FROM `{temporary_table}`;")
+    return list(self._cursor)
+
+  def batched_update_direction(self, new_direction_value, id_list, batch_size=1024):
+    """Updates the `direction` column of rows with ids in `id_list` to `new_direction_value`,
+    the update query is executed in batches of `bath_size` rows at a time.
```
krivard, line 294:
```suggestion
    the update query is executed in batches of `batch_size` rows at a time.
```

src/acquisition/covidcast/database.py
```diff
@@ -158,6 +163,179 @@ def update_direction(
...
+
+    `new_direction_value`: the new value to be stored in `direction` column.
+    `id_list`: a list of ids of rows that will change to `new_direction_value`.
+    `batch_size`: batch size of the update query.
+    """
```
krivard, line 299: Can you add a brief explanation here on why batch_size is necessary, and guidance on how to decide whether to use the default or specify a higher or lower value?

.idea/codeStyles/codeStyleConfig.xml
AlSaeed: Noted, this was added by mistake.

src/acquisition/covidcast/database.py
AlSaeed: The reason why I added `batch_size` is that there is a maximum limit on the length of an SQL statement, and so in the case where the list size is 4M this limit will be certainly exceeded. As for choosing the `batch_size` there is a tradeoff, as with larger `batch_size` the update query will take longer to execute but there will be fewer queries. The only way I can think of to choose `batch_size` is experimentally. I see your point that it might be more appropriate for it to be hardcoded than to be exposed as a parameter.

krivard:
Replaced by #149

PR #135, opened 2020-06-17
Make metadata requests use the cached endpoint automatically

capnrefsmmat:
The overhead for the non-cached endpoint is substantial.

(I'm going to check if David can push a new version of `delphi-epidata` to PyPI for us, so we can release a Python package that has fast access to the metadata without monkey-patching the current version.)

### DIFF
--- a/src/client/delphi_epidata.R
+++ b/src/client/delphi_epidata.R
@@ -524,7 +524,7 @@ Epidata <- (function() {
 
   # Fetch Delphi's COVID-19 Surveillance Streams metadata
   covidcast_meta <- function() {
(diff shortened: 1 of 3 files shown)


capnrefsmmat:
I don't know if it's worth taking an optional argument to force skipping the cache; do we imagine clients needing to do that?

krivard:
> I don't know if it's worth taking an optional argument to force skipping the cache; do we imagine clients needing to do that?

No, and we can always add it later if someone comes up with a use case. Cached data is only returned if the cache is less than 75 minutes old, otherwise it runs the full aggregation-based query.

PR #136, opened 2020-06-20
Fix/cvc meta timeout

krivard:
This PR addresses the metadata cache update timeout problem by switching how the cache updater obtains live metadata computations. The cache updater used to call the python API for this; it now runs its own SQL query.

All integration and unit tests pass -- they have been updated with appropriate MagicMock settings for the covidcast database utility, and to call the cache updater when needed for testing the API `covidcast_meta` call (which now only supports pulling cached metadata)

If you ask the API server for cached metadata, and the metadata is stale, it will log an error on the server, and then give you the stale metadata anyway.

### DIFF
--- a/integrations/server/test_covidcast_meta.py
+++ b/integrations/server/test_covidcast_meta.py
@@ -7,6 +7,8 @@ import unittest
 import mysql.connector
 import requests
 
+#first party
+from delphi.epidata.acquisition.covidcast.covidcast_meta_cache_updater import main as update_cache
 
 # use the local instance of the Epidata API
 BASE_URL = 'http://delphi_web_epidata/epidata/api.php'
@@ -69,6 +71,7 @@ class CovidcastMetaTests(unittest.TestCase):
               for gv, v in zip(('geo1', 'geo2'), (10, 20)):
                 self.cur.execute(template % (src, sig, tt, gt, tv, gv, v))
     self.cnx.commit()
+    update_cache(args=None)
 
     # make the request
     response = requests.get(BASE_URL, params={'source': 'covidcast_meta'})
@@ -114,6 +117,7 @@ class CovidcastMetaTests(unittest.TestCase):
               for gv, v in zip(('geo1', 'geo2'), (10, 20)):
                 self.cur.execute(template % (src, sig, tt, gt, tv, gv, v))
     self.cnx.commit()
+    update_cache(args=None)
 
     # make the request
     response = requests.get(BASE_URL, params={'source': 'covidcast_meta'})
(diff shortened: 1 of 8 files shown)


src/server/api.php
```diff
@@ -1005,18 +1004,14 @@ function get_covidcast_meta_cached() {
   $epidata = null;
   $result = mysqli_query($dbh, $query);
   if($row = mysqli_fetch_array($result)) {
+    // parse and use the cached response
+    $epidata = json_decode($row['epidata'], true);
+
     if (intval($row['age']) < $max_age && strlen($row['epidata']) > 0) {
```
capnrefsmmat, line 1010: `$max_age` is no longer defined, since line 998 was deleted above. Also, this seems to issue an error if the row is too *new*, not too *old*

src/acquisition/covidcast/database.py
```diff
@@ -251,6 +251,42 @@ def update_timeseries_timestamp2(
...
+        t.`geo_type`
+      ORDER BY
+        t.`source` ASC,
+        t.`signal` ASC,
+        t.`time_type` ASC,
+        t.`geo_type` ASC
+'''
+    FIELDS="data_source signal time_type geo_type min_time max_time num_locations min_value max_value mean_value stdev_value last_update".split()
+    self._cursor.execute(sql)
+    return list(dict(zip(FIELDS,x)) for x in self._cursor)
```
capnrefsmmat, line 288: I believe you can use `self._cursor.column_names` instead of having to write `FIELDS` manually, provided the column names in the query are the names you want.

https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlcursor-column-names.html

krivard:
Implemented all changes suggested by @capnrefsmmat; all tests pass.

integrations/server/test_covidcast_meta.py
```diff
@@ -69,6 +71,7 @@ def test_round_trip(self):
               for gv, v in zip(('geo1', 'geo2'), (10, 20)):
                 self.cur.execute(template % (src, sig, tt, gt, tv, gv, v))
     self.cnx.commit()
+    update_cache(args=None)
```
capnrefsmmat, line 74: Will this try to connect to the production database? I don't see any provision in `update_cache` for connecting to a mock database; regardless of what connection the test makes, `update_cache` uses the database details in the secrets file.

Unless the test environment just has identical secrets but a different database?

krivard: My understanding is that the test environment is isolated from the outside internet via the Docker container, and could not contact the production database even if it wanted to. 

That said, we can be confident that this particular test writes to the `covidcast` table on host `delphi_web_epidata`, which is the Docker container. The test then asks for the metadata from host `delphi_web_epidata`, and receives the expected result. So *someone* must be writing to the `covidcast_meta_cache` table on host `delphi_web_epidata`, so if it's not `update_cache` using a secrets file modified by the testing machinery to point to `delphi_web_epidata` instead of the production server, then we have a whole nother set of problems.

PR #143, opened 2020-06-29
Implement data versioning for COVIDcast in server and client

krivard:
This PR includes a migration assistance script -- it seems like a good idea to track these in source control, since even if we only use them once, we may regret losing track of what was done.


src/server/api_helpers.php
```diff
@@ -157,6 +157,11 @@ function execute_query($query, &$epidata, $fields_string, $fields_int, $fields_f
   global $dbh;
   global $MAX_RESULTS;
   $result = mysqli_query($dbh, $query . " LIMIT {$MAX_RESULTS}");
+  error_log("Query: ".$query);
```
krivard, line 160:
```suggestion
```

krivard:
@korlaxxalrok This PR includes Eu Jing's script that converts a set of .sql backups into a full record of issue dates. It is a bit of a memory hog, so it's probably worth trying a couple of test runs before setting it loose on the full history. His notes:

> Just in case you want to run the migration script, it requires a 3rd party package csvdiff.
> I have been running it with [the week's worth of compressed backup files](http://rinkitink.ml.cmu.edu/delphi/misc/issues_from_backups/) in the same dir as:
> ```
> python proc_db_backups.py --input-files just_covidcast_*_database.sql.gz --tmp-dir ./tmp  \
>    --out-dir ./out --max-insert-chunk 1000 --skip jhu-csse usa-facts safegraph --debug
> ```
> But I have also observed sources like doctor-visits and fb-survey consume around 10+ GB of ram during the diffing, 
> so skipping those to save memory might be good too just to see how it runs.
>
> It should produce one SQL file per source in the ./out directory, and also several intermediate CSVs in the ./tmp directory.

korlaxxalrok:
@krivard I see this now 👁️ Thanks for the reminder 🙇 I'll check it out asap.

src/server/api.php
```diff
@@ -931,7 +931,13 @@ function get_dengue_nowcast($locations, $epiweeks) {
...
 //   $geo_value (required): location identifier or `*` as a wildcard for all
 //     locations (specific to `$geo_type`)
-function get_covidcast($source, $signal, $time_type, $geo_type, $time_values, $geo_value) {
+//   $issues (optional): array of time values/ranges
+//     overrides $lag
+//     default: most recent issue
+//   $lag (optional): number of time units between each time value and its issue
+//     overridden by $issues
+//     default: most recent issue
+function get_covidcast($source, $signal, $time_type, $geo_type, $time_values, $geo_value, $issues, $lag) {
```
AlSaeed, line 940: For consistency with issue_date, should lag be also filtered by a list?

src/server/api.php
```diff
@@ -941,27 +947,47 @@ function get_covidcast($source, $signal, $time_type, $geo_type, $time_values, $g
...
+    $condition_issue = filter_integers('t.`issue`', $issues);
+    $condition_version = $condition_issue;
+  } else if($lag !== null) {
+    //build the lag filter
+    $condition_lag = "(t.`lag` = {$lag})";
+    $condition_version = $condition_lag;
+  } else {
+    //fetch most recent issues
+    $sub_fields = "max(`issue`) `max_issue`, `time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
+    $sub_group = "`time_value`, `source`, `signal`, `geo_value`";
```
AlSaeed, line 984:
```suggestion
    $sub_group = "`time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
```

src/server/covidcast_issues_migration/proc_db_backups.py
```diff
@@ -0,0 +1,441 @@
...
+    for _, sub_f in open_file_writers.items():
+        sub_f.close()
+
+    return created_files
+
+def date_to_int(date: datetime.date) -> int:
+    '''
+    Convert input datetime.date to date integer format YYYYmmdd
+    '''
+    return date.day + date.month * 100 + date.year * 10000
```
AlSaeed, line 247:
```suggestion
    return int(date.strftime("%Y%m%d"))
```

src/server/covidcast_issues_migration/proc_db_backups.py
```diff
@@ -0,0 +1,441 @@
...
+
+def int_to_date(date_int: int) -> datetime.date:
+    '''
+    Convert input date integer format YYYYmmdd to datetime.date
+    '''
+    day = date_int % 100
+    month = (date_int // 100) % 100
+    year = date_int // 10000
+    return datetime.date(year, month, day)
```
AlSaeed, line 256:
```suggestion
    return datetime.datetime.strptime(str(date_int),"%Y%m%d").date()
```

src/server/covidcast_issues_migration/proc_db_backups.py
```diff
@@ -0,0 +1,441 @@
...
+
+# Ensure the hard-coded index indices are right
+assert INDEX_COLS[2] == "time_type"
+assert INDEX_COLS[4] == "time_value"
+
+def create_issue_from_change(
+        latest_issue: dict, change: dict,
+        index: tuple, issue_date: datetime.date) -> dict:
+    '''
+    Creates an issue, i.e. the values that may chaneg across time from the latest issue
```
AlSaeed, line 309:
```suggestion
    Creates an issue, i.e. the values that may change across time from the latest issue
```

AlSaeed:
By only inserting diffs into the database the semantics of the API is changed. For example a query that filters by issue_date will only retrieve new or changed rows in these issue_dates rather than the entire issues.

capnrefsmmat:
> By only inserting diffs into the database the semantics of the API is changed. For example a query that filters by issue_date will only retrieve new or changed rows in these issue_dates rather than the entire issues.

If we only store diffs, we probably want a query for issue date 2020-06-01 to find the most recent data for each location that was issued on *or before* 2020-06-01.

That does leave open the question of how to handle a new issue that removes a particular geo_id from reporting, e.g. if a data revision brought it below a sample size threshold. A query that finds the most recent issue for each geo_id would fail to notice that, unless we issue an NA in this case.

There's also the question of how to implement such a query so that finding the most recent issue for 3000 counties can be done very quickly.

AlSaeed:
> > By only inserting diffs into the database the semantics of the API is changed. For example a query that filters by issue_date will only retrieve new or changed rows in these issue_dates rather than the entire issues.
> 
> If we only store diffs, we probably want a query for issue date 2020-06-01 to find the most recent data for each location that was issued on _or before_ 2020-06-01.
> 
> That does leave open the question of how to handle a new issue that removes a particular geo_id from reporting, e.g. if a data revision brought it below a sample size threshold. A query that finds the most recent issue for each geo_id would fail to notice that, unless we issue an NA in this case.
> 
> There's also the question of how to implement such a query so that finding the most recent issue for 3000 counties can be done very quickly.

In the private API a parameter with the name `as_of_issue_date` is used to satisfy this query for a single `issue_date`, though it will report the latest issue of a key even if the key was removed in the last issue. If version filters were limited to this parameter then, ignoring removed keys, the data will be returned as expected but the `issue_date` and `lag` values returned will correspond with the latest diff rather than the latest issue per key.

krivard:
When I asked Jacob about Forecasting's needs for the data versioning interface a few weeks ago, we decided three things:
* We only store diffs. Data sets like cases/deaths are too dense to store a full issue for each update. (that said, we are currently uploading the full calendar of cases/deaths each day in automation, so this should be revisited)
* If you don't specify an issue date in the query, covidcast should return the most recent issue for each time_value requested
* If you do specify an issue date or a range of issue dates in the query, covidcast should return all matching issues for each time_value requested, which may include multiple entries for a single time_value. "On-or-before" functionality is specified using a range of issue dates.

This design does not handle deletions, but then again, neither does the current unversioned data.

krivard:
Update from team leads meeting:

* The terminology "as-of" is probably more useful than "issue" for sensors that only update a small portion of the calendar each day (as most covidcast sensors do). "As-of" would mean "Most recent data issued on or before" some time_value. 
* It is useful to report different kinds of missingness. Currently covidcast reports -2 No Results whenever the requested combination of source, signal, time_type, time_value, geo_type, geo_value is not found in the database. This includes malformed sources, signals, time_types, time_values, geo_types, and geo_values, *as well as* properly-formed queries that do not report a value due to privacy/sample size concerns. If we had a way to specify the latter, this would solve the deletions problem.

I do not think either of the above are blockers to merging this pull request and proceeding with the data migration, but they should be resolved before we publish updates to the clients.

krivard:
@eujing, would you check my work on adding `as_of` to the clients?

krivard:
Replaced by #149

PR #147, opened 2020-07-06
Document missingness of each signal

capnrefsmmat:
So far, this is just an update to the source template to encourage new sources to document when they are missing and what kinds of missingness are possible.

Before merging, we should add documentation for each existing source when applicable.


krivard (approved):
truncating this work to just the template

PR #149, opened 2020-07-09
Feature/data versioning and optimization

krivard:
This PR combines the changes from #133 and #143, rebased for legibility and to incorporate the changes in meta behavior on main.


krivard:
@korlaxxalrok this is the PR we'll want to merge on Migration Day simultaneously with the data loading, since it will not insert a spurious issue for each direction recomputation.

src/acquisition/covidcast/direction_updater.py
```diff
@@ -144,17 +145,135 @@ def get_direction_impl(x, y):
...
+
+    # A DataFrame holding rows that potentially changed direction value
+    ts_pot_changed = ts_rows.set_index('time_value').loc[days]
+    ts_pot_changed['new_direction'] = np.array(directions, np.float64)
+
+    # is_eq_nan = ts_pot_changed.direction.isnull() & ts_pot_changed.new_direction.isnull()
+    # is_eq_num = ts_pot_changed.direction == ts_pot_changed.new_direction
+    # changed_mask = ~(is_eq_nan | is_eq_num)
+    # ts_changed = ts_pot_changed[changed_mask]
+
```
eujing, line 256: Should we remove these if not needed anymore?

src/acquisition/covidcast/direction_updater.py
```diff
@@ -144,17 +145,135 @@ def get_direction_impl(x, y):
...
+      print(msg % args)
+
+    offsets = ts_rows.offsets.values.astype(np.int64)
+    days = ts_rows.time_value.values.astype(np.int64)
+    values = ts_rows.value.values.astype(np.float64)
+    timestamp1s = ts_rows.timestamp1.values.astype(np.int64)
+    timestamp2s = ts_rows.timestamp2.values.astype(np.int64)
+
+    # create a direction classifier for this signal
+    data_stdev = data_stdevs[source][signal][geo_type]
```
AlSaeed, line 234:
```suggestion
    data_stdev = data_stdevs[source][signal][geo_type] if (source in data_stdevs) and (signal in data_stdevs[source]) and (geo_type in data_stdevs[source][signal]) else 0
```
To avoid crashing in case some `(source,signal,geo_type)` tuple doesn't have data on or before `Constants.SIGNAL_STDEV_MAX_DAY`.

src/acquisition/covidcast/direction_updater.py
AlSaeed: I agree, I considered updating only the rows where the value actually changed, it turned out that this increases overall runtime, so I didn't end up using it.

AlSaeed:
My only additional note is that `get_data_stdev_across_locations` in `database.py` is still computed regardless of issue dates,  I am not sure if this is the intended behavior or if it also should be reimplemented to compute standard deviation on only data with latest issue per key.

PR #150, opened 2020-07-09
Improvements to acquisition / CSV file loading.

melange396:
SQL batching improves ingestion speed significantly.
Loads files in name-sorted order for determinism and saner progress tracking.
Individual transactions for each file (instead of one transaction for the full collection of files) results in:
-an UPSERT failure means the entirety of the file will not be loaded (before a single line failure would preserve the preceding lines and attempt the following lines of the same file).
-lock should be released more frequently, meaning read access should not block as much while acquisition is in progress.


melange396:
i started but then resisted the urge to rewrite more of this stuff in lieu of a smaller changelist...  there is some dead code that could be removed but i thought leaving it might provide context as to why the call structure is the way it is.  i also began to look into better handling of import failures, but decided that should go into its own issue, and i wasnt sure how representative my sample data files are.  can we try this out on the staging server?  if so, how can i do that?

melange396:
some TODOs i didnt put in the code:
--do more speed testing with UPDATEs instead of clean INSERTs (UPDATE will be slower than INSERT in any case, but its unclear how much batching improves this)
--better deal with failure files - save error message somewhere, alert someone that errors happened, perhaps preserve a summary of what files were/werent loaded in each run, etc...
--ON DUPLICATE KEY - update (zero out) timestamp2 & direction ?? [ database.py:104 -ish ]
--make sure directories exist before writing files?  it might do this already?

melange396:
im seeing 2-4x speed up for acquisition depending on which subsets of sample data i provide it with, though that is on a beefy but shared test machine.  @AlSaeed is there a particular tool used to get this profiling output, or was that just something you did by hand?  https://github.com/cmu-delphi/delphi-epidata/pull/133#issuecomment-645358029

AlSaeed:
> im seeing 2-4x speed up for acquisition depending on which subsets of sample data i provide it with, though that is on a beefy but shared test machine. @AlSaeed is there a particular tool used to get this profiling output, or was that just something you did by hand? [#133 (comment)](https://github.com/cmu-delphi/delphi-epidata/pull/133#issuecomment-645358029)

This was done by hand via inserting boilerplate timing code around specific code blocks.

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -73,27 +73,21 @@ def archive_as_successful(path_src, filename, source):
...
-            row_values.geo_value,
-            row_values.value,
-            row_values.stderr,
-            row_values.sample_size)
-      except Exception as e:
+    csv_rows = csv_importer_impl.load_csv(path, geo_type)
+
+    if not csv_rows:
+      all_rows_valid = False
+      continue
```
krivard, line 81: This continue will skip to the next file without archiving the busted one as failed

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -73,27 +73,21 @@ def archive_as_successful(path_src, filename, source):
...
-            geo_type,
-            time_value,
-            row_values.geo_value,
-            row_values.value,
-            row_values.stderr,
-            row_values.sample_size)
-      except Exception as e:
+    csv_rows = csv_importer_impl.load_csv(path, geo_type)
+
+    if not csv_rows:
```
krivard, line 79: `csv_rows` is a generator and will always be truthy; you probably want `any()` or `all()` here instead

src/acquisition/covidcast/database.py
```diff
@@ -251,41 +339,6 @@ def update_timeseries_timestamp2(
     args = (source, signal, time_type, geo_type, geo_value)
     self._cursor.execute(sql, args)
 
-  def get_covidcast_meta(self):
```
krivard, line 254: Why was this removed?

src/acquisition/covidcast/csv_to_database.py
melange396: good catch

src/acquisition/covidcast/csv_to_database.py
melange396: good catch

src/acquisition/covidcast/database.py
melange396: whoops!  looks like it was from a merge conflict...  i dont know how i missed that

PR #152, opened 2020-07-15
First draft of detailed Google Survey documentation

capnrefsmmat:
Based on text provided by Natalia and Justin.


robohyun66:
Looks great! I found no mistakes, and the bootstrap methodology should be added as the next step.

PR #153, opened 2020-07-16
Bug-Fix/Substituting WITH statement

AlSaeed:
* WITH statement isn't defined in MariaDB 5.5.67.
* The bug fix is to use the sql statement directly instead of the CTE.
* Cannot use temporary table, as it cannot be opened twice while evaluating the same query.


krivard:
Resolved by upgrading MariaDB on prod instead

PR #161, opened 2020-07-24
Partitioning Direction Update

AlSaeed:


AlSaeed:
The new direction updater takes an argument `partitions`: a list of integers in the range [0, 23] representing the indexes of the partitions to be processed. It will then run the direction updater for each one of these partitions in sequence. The default value is the list of all partition indexes.
Example:
`python3 -m delphi.epidata.acquisition.covidcast.direction_updater --partitions 0 1` will process the first two partition in sequence.

AlSaeed:
To add the `is_latest_issue` column, run the two following commands:
1. ```ALTER TABLE `covidcast` ADD `is_latest_issue` BINARY(1) NOT NULL DEFAULT (0);```
2. ```python3 -m delphi.epidata.acquisition.covidcast.fill_is_latest_issue```

We can drop the `fill_is_latest_issue` and its corresponding integration test once they are run successfully.

korlaxxalrok:
@AlSaeed The new column is added. Once merged we can do the second step.

src/acquisition/covidcast/fill_is_latest_issue.py
```diff
@@ -0,0 +1,86 @@
...
+# third party
+import mysql.connector
+
+# first party
+import delphi.operations.secrets as secrets
+
+
+# partition configuration
+PARTITION_VARIABLE = 'geo_value'
+PARTITION_SPLITS = ["'05101'", "'101'", "'13071'", "'15007'", "'17161'", "'19039'", "'20123'", "'21213'", "'24035'",
```
melange396, line 15: how did you come up with these split values?

src/acquisition/covidcast/direction_updater.py
```diff
@@ -39,11 +39,26 @@ class Constants:
...
+PARTITION_SPLITS = ["'05101'", "'101'", "'13071'", "'15007'", "'17161'", "'19039'", "'20123'", "'21213'", "'24035'",
+                    "'27005'", "'28115'", "'29510'", "'31161'", "'35100'", "'37117'", "'39081'", "'41013'", "'44140'",
+                    "'47027'", "'48140'", "'48461'", "'51169'", "'55033'"]
+
+
+
 def get_argument_parser():
   """Define command line arguments."""
 
   # there are no flags, but --help will still work
```
melange396, line 53: delete outdated comment

src/acquisition/covidcast/direction_updater.py
```diff
@@ -278,21 +302,22 @@ def main(
...
+      update_loop_impl(database, partition_index)
+      # only commit on success so that directions are consistent with respect
+      # to methodology
+      commit = True
+    except Exception as e:
+      raise e
+    finally:
+      # no catch block so that an exception above will cause the program to
+      # fail after the following cleanup
+      database.disconnect(commit)
```
melange396, line 319: possibly consider moving this out of the loop so the db is consistent?  i guess having partially updated directions isnt such a bad thing

src/acquisition/covidcast/direction_updater.py
AlSaeed: Done

src/acquisition/covidcast/direction_updater.py
AlSaeed: I guess there is an argument for both sides. The other side being this allows to investigate the failure and only update the remaining partitions instead of running the entire job again.

src/acquisition/covidcast/fill_is_latest_issue.py
AlSaeed: They were chosen so that they split a recent full issue into roughly equal partitions.

src/acquisition/covidcast/database.py
```diff
@@ -92,110 +92,141 @@ def count_all_rows(self):
...
+
+        if result is None:
+          # the SQL connector does not support returning number of rows affected
+          total = None
+        else:
+          total += result
+        if commit_partial:
+          self._connection.commit()
+    except Exception as e:
+      print('AAA')
```
melange396: you can get rid of this, but its really not necessary

AlSaeed: It was removed in the previous commit

src/acquisition/covidcast/direction_updater.py
melange396: ¯\_(ツ)_/¯ its probably fine to leave it

krivard: Either way someone has to go in and fix it manually, so as long as we document what the behavior is and what should be done in a failure, it's fine.

PR #162, opened 2020-07-24
Support for is_wip flag in DB

melange396:


melange396:
will we need to run this SQL in production, or does the database get truncated and repopulated?

```
ALTER TABLE `covidcast` ADD `is_wip` BINARY(1) DEFAULT (NULL);
UPDATE TABLE `covidcast` SET `is_wip` = 1 WHERE `is_wip` IS NULL and `signal` LIKE 'wip_%';
UPDATE TABLE `covidcast` SET `is_wip` = 0 WHERE `is_wip` IS NULL and `signal` NOT LIKE 'wip_%';
```

krivard:
That SQL will probably be more efficient than trying to run the full data load again (iirc that had to run overnight last time)

melange396:
i guess we can close this as the commits are included in https://github.com/cmu-delphi/delphi-epidata/pull/167 ?

PR #167, opened 2020-07-29
Meta cache breakdown

melange396:
1 SQL query per distinct (src, sig) instead of 1 global

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -94,113 +94,141 @@ class Database:
   def insert_or_update_bulk(self, cc_rows):
     return self.insert_or_update_batch(cc_rows)
 
-  def insert_or_update_batch(self, cc_rows, batch_size=0, commit_partial=False):
+  def insert_or_update_batch(self, cc_rows, batch_size=2**20, commit_partial=False):
     """
     Insert new rows (or update existing) in the `covidcast` table.
 
     This has the intentional side effect of updating the primary timestamp.
+
+
     """
-    sql = '''
-      INSERT INTO `covidcast`
-        (`id`, `source`, `signal`, `time_type`, `geo_type`, `time_value`, `geo_value`,
-        `value_updated_timestamp`, `value`, `stderr`, `sample_size`,
-        `direction_updated_timestamp`, `direction`,
-        `issue`, `lag`, `is_wip`)
+
+    tmp_table_name = 'tmp_insert_update_table'
+
+    create_tmp_table_sql = f'''
+      CREATE TEMPORARY TABLE `{tmp_table_name}` (
+        `source` varchar(32) NOT NULL,
+        `signal` varchar(64) NOT NULL,
+        `time_type` varchar(12) NOT NULL,
+        `geo_type` varchar(12) NOT NULL,
+        `time_value` int(11) NOT NULL,
+        `geo_value` varchar(12) NOT NULL,
+        `value_updated_timestamp` int(11) NOT NULL,
+        `value` double NOT NULL,
+        `stderr` double,
+        `sample_size` double,
+        `direction_updated_timestamp` int(11) NOT NULL,
+        `direction` int(11),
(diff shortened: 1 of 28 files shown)


melange396:
so in sum total, the applying this change to the db would either require wiping/recreating/repopulating from scratch, or applying this SQL:

```
ALTER TABLE `covidcast` RENAME COLUMN `timestamp1` to `value_updated_timestamp`;
ALTER TABLE `covidcast` RENAME COLUMN `timestamp2` to `direction_updated_timestamp`;

ALTER TABLE `covidcast` ADD `is_wip` BINARY(1) DEFAULT (NULL);

-- and then either these (should be faster)
UPDATE TABLE `covidcast` SET `is_wip` = 1 WHERE `signal` LIKE 'wip_%';
UPDATE TABLE `covidcast` SET `is_wip` = 0 WHERE `is_wip` IS NULL;

-- or these (at least slightly safer)
UPDATE TABLE `covidcast` SET `is_wip` = 0 WHERE `is_wip` IS NULL and `signal` NOT LIKE 'wip_%';
UPDATE TABLE `covidcast` SET `is_wip` = 1 WHERE `is_wip` IS NULL and `signal` LIKE 'wip_%';
```

korlaxxalrok:
@melange396 The necessary changes should be done.

src/acquisition/covidcast/database.py
```diff
@@ -539,7 +544,13 @@ def update_timeseries_timestamp2(
   def get_covidcast_meta(self):
     """Compute and return metadata on all non-WIP COVIDcast signals."""
 
-    sql = '''
+    meta = []
+
+    sql = 'SELECT `source`, `signal` FROM covidcast GROUP BY `source`, `signal` ORDER BY `source` ASC, `signal` ASC;'
```
AlSaeed, line 549: I think it is more appropriate to filter by `is_wip` in this query

src/acquisition/covidcast/database.py
```diff
@@ -570,39 +581,38 @@ def get_covidcast_meta(self):
...
           x.`time_value` = t.`time_value` AND
-          x.`source` = t.`source` AND
-          x.`signal` = t.`signal` AND
           x.`geo_type` = t.`geo_type` AND
           x.`geo_value` = t.`geo_value`
       WHERE
-        t.`signal` NOT LIKE 'wip_%'
+        NOT t.`is_wip` AND
+        t.`source` = %s AND
+        t.`signal` = %s
```
AlSaeed, line 602: I suggest to keep merging on `source`, and `signal` in order to avoid filtering by them twice.

src/acquisition/covidcast/database.py
```diff
@@ -612,5 +622,21 @@ def update_covidcast_meta_cache(self, epidata_json):
         `timestamp` = UNIX_TIMESTAMP(NOW()),
         `epidata` = %s
     '''
+    epidata_json = json.dumps(metadata)
 
     self._cursor.execute(sql, (epidata_json,))
+
+  def retrieve_covidcast_meta_cache(self):
```
AlSaeed, line 629: A docstring should probably be added. Additionally this method doesn't seem to be used.

melange396: i used it in debugging...  i can remove it if you think thats best practices or whatever.

src/acquisition/covidcast/database.py
melange396: good call

src/acquisition/covidcast/database.py
melange396: i didnt like seeing the repetition, but thats a good point

src/acquisition/covidcast/database.py
AlSaeed: This seems like a useful function. Probably we only need to add a docstring and keep it.

AlSaeed:
Merge Conflict Resolution PR:
https://github.com/melange396/delphi-epidata/pull/1

PR #168, opened 2020-07-30
Increase maximum signal length for COVIDcast to 64 characters

krivard:
* Database schema
* Signal name length guard in CSV uploader
* Integration test verifies signal length 32<x<64 succeeds, x>64 fails

Unit tests and integration tests pass.

To apply this to the live DB, use the following:

```
ALTER TABLE `covidcast`
MODIFY `signal` varchar(64) NOT NULL;
```


korlaxxalrok:
The column type has been updated 👍

PR #170, opened 2020-07-30
Add support for querying multiple signals

krivard:
Fixes cmu-delphi/covidcast-indicators#116

* backwards-compatible signal/signals parameter
* optional 'format' parameter for retrieving results as a tree (preferred by Viz) or as a flat list (backwards-compatible); defaults to flat list
* client support: coffee, js, R, py
* integration tests to check multi-signal queries and queries for tree format

### DIFF
--- a/integrations/acquisition/covidcast/test_covidcast_meta_caching.py
+++ b/integrations/acquisition/covidcast/test_covidcast_meta_caching.py
@@ -1,6 +1,7 @@
 """Integration tests for covidcast's metadata caching."""
 
 # standard library
+import json
 import unittest
 
 # third party
@@ -100,15 +101,15 @@ class CovidcastMetaCacheTests(unittest.TestCase):
         'stdev_value': 0,
         'max_issue': 20200423,
         'min_lag': 0,
-        'max_lag': 1
+        'max_lag': 1,
       }
     ])
-    epidata1={'result':1,'message':'success','epidata':epidata1}
+    epidata1={'result':1, 'message':'success', 'epidata':epidata1}
 
     # make sure the API covidcast_meta is still blank, since it only serves
     # the cached version and we haven't cached anything yet
     epidata2 = Epidata.covidcast_meta()
-    self.assertEqual(epidata2['result'], -2)
+    self.assertEqual(epidata2['result'], -2, json.dumps(epidata2))
 
     # update the cache
     args = None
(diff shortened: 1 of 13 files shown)


krivard:
@capnrefsmmat can you check my work in the raw R client? Backwards-compatibility is not something I've had to deal with before in R.

sgratzl:
looking good, just to clarify: the tree format is

```js
{
  signal1: [...],
  signal2: [...],
}
```

for me, I would call it more a group by signal. One could do a similar things for every parameter that allows multiple values. 

general question: this PR allows me to query multiple signals of the same data source? how about multiple signals distributed in different data sources?

capnrefsmmat:
The R client change looks fine -- it should support the old syntax and anyone using named arguments. I'm not sure how jsonlite will parse the tree format, but as long as the default format is unchanged from before, current users should not be surprised.

krivard:
@sgratzl 
> how about multiple signals distributed in different data sources?

Since we don't enforce unique signal names, what we'd really want for that is to accept (source, signal) pairs, and that's a larger architectural shift than we're really prepared for just now.

sgratzl:
why not enforcing that a certain character such as `:` doesn't appear and that you have a combined unique signal id: `datasource:signal`

krivard:
That would probably be a fine way to handle it in the clients, but I'm more concerned about [the query architecture](https://github.com/krivard/delphi-epidata/blob/28b9f8afaa61253f6add469a189b42c112cd5efa/src/server/api.php#L940-L999) just now. How important is this to you, should we wait to merge this PR until we get full (source, signal)* support working?

sgratzl:
as soon as we show small multiples of different signals in production, it will become handy.

krivard:
Followed up with @sgratzl offline: Small multiples sends 12 API requests (one query per source-signal pair) in the default view, but network time for these queries is less than 10% of the page render time, largely due to browsers parallelizing network requests. 

Since small multiples is the major use case for source-signal pairs, and it's not a responsiveness issue, we'll hold off on source-signal pair support for now.

sgratzl:
another use case which involves more data is when querying for average / count / average cumulative / count cumulative of cases / deaths. since they are all coming from the same data source this PR would save a lot

krivard:
Yep, that was the original plan for this feature. Once #187 merges we'll
apply this one.

krivard:
@melange396 check my work on [a57b61a](https://github.com/cmu-delphi/delphi-epidata/pull/170/commits/a57b61aec8a4d4d78bf194841a8c7520ef22c255)? tests pass but I want to make sure there's not a gotcha I didn't see

melange396 (approved):
LGTM!

Issue #160, opened 2020-07-23
Compress COVIDcast API responses


krivard:
Especially since county-level responses are super large.

sgratzl:
this is very important in my opinion since it significantly down the COVIDast map. 

see also e.g., https://medium.com/@ali.dev/how-to-setup-caching-gzip-compression-in-apache-web-server-with-htaccess-292a0f689553

https://httpd.apache.org/docs/2.4/mod/mod_brotli.html

in this case https://github.com/cmu-delphi/delphi-epidata/blob/main/src/server/.htaccess would be a start to configure that

capnrefsmmat:
Surprise! delphi.cmu.edu seems to be served by nginx 1.16.1, not Apache, but the map API endpoints are on Apache. I submitted #171 to fix the htaccess for those.

In the nginx config on delphi.cmu.edu we just need to add

```
gzip on;
gzip_types application/json;
```

to the appropriate server block. Pinging @korlaxxalrok. I think this will speed up queries to signals like DV, which can be 200+ kB per day of data. According to a random online checker tool I found, one example day can be compressed to only 14 kB. (Seems implausible, but if random website tools can't be trusted, who can?)

capnrefsmmat:
Fixed in #191.

PR #171, opened 2020-08-02
Enable gzip/brotli compression and caching of API responses

capnrefsmmat:
See #160 -- this will only help if deployed to the AWS API endpoints at api.covidcast.cmu.edu, since those run Apache and the others run nginx.

This assumes we have mod_deflate and mod_brotli available on those servers.

### DIFF
--- a/src/server/.htaccess
+++ b/src/server/.htaccess
@@ -1,2 +1,6 @@
 # Open up the API to all origins.
 Header set Access-Control-Allow-Origin "*"
+
+# Compress API responses
+AddOutputFilterByType DEFLATE application/json
+AddOutputFilterByType BROTLI_COMPRESS application/json
--- a/src/server/.htaccess
+++ b/src/server/.htaccess
@@ -2,8 +2,15 @@
 Header set Access-Control-Allow-Origin "*"
 
 # Compress API responses
-AddOutputFilterByType DEFLATE application/json
-AddOutputFilterByType BROTLI_COMPRESS application/json
+<IfModule mod_deflate.c>
+    AddOutputFilterByType DEFLATE application/json
+</IfModule>
+<IfModule mod_brotli.c>
+    AddOutputFilterByType BROTLI_COMPRESS application/json
+</IfModule>
 
 # Allow brief caching of API responses
-ExpiresByType application/json "access plus 2 hours"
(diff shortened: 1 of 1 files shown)


capnrefsmmat:
Just also added API response caching for 2 hours. This depends on mod_expires and will only apply to the AWS API endpoints, since they run Apache.

korlaxxalrok:
This makes sense, @capnrefsmmat. I need to check to make sure those Apache modules are available.

Also, everything runs Apache, but we have a bit of nginx sitting in front of what we serve on the Delphi primary server. Might see some benefit there as well, but we may need to possibly look at an additional nginx config.

capnrefsmmat:
I can also wrap everything up in the right `IfModule` blocks if that would be useful.

I'd assume that if nginx is serving as reverse proxy, it would pass through any compression and caching headers, but I guess if it doesn't we can fix it.

korlaxxalrok:
I think we'd need to get `brotli`, which should not be a problem. I will do that and then we can proceed 👍

capnrefsmmat:
I don't think mod_brotli is a necessity; DEFLATE compression should be good enough for most purposes. Brotli is just a slightly-faster slightly-smaller compression.

So if it'd take much work to add mod_brotli, we can just delete that line instead.

sgratzl:
any estimate when this one will be merged and available in production?

korlaxxalrok:
Ah! Sorry, I keep falling out of the loop on this one. Let me look quickly this morning to see about the brotli piece and then we can move forward.

korlaxxalrok:
@capnrefsmmat I think it'll take some effort to work mod_brotli into the picture. The best option will be to upgrade Apache to a version that includes it. I'd probably want to take some time to make sure we can do this easily and safely for all the servers that are running it.

So if it isn't a deal breaker maybe we move forward without mod_brotli for now, then follow on with the future task to upgrade Apache.

Reasonable?

capnrefsmmat:
Sounds reasonable. I just pushed a change that uses IfModule around the new directives, so it should be possible to use the htaccess as-is. Then if we install mod_brotli we don't have to change anything in the config, just restart Apache.

I didn't actually test this change, so worth checking the syntax is right, but otherwise I think it should work fine and get gzip compression working.

PR #187, opened 2020-08-12
Apply "quick fix" data ingestion performance improvements

krivard:
Fixes cmu-delphi/covidcast-indicators#158

* Add flag columns for wip and latest-issue status
* Partition direction updates x24 for memory and running time
* Partition meta cache update xN(signal, source) for memory
* Give timestamp columns more descriptive names
* Extend signal name from 32 to 64 chars

Results:

* Direction update runs in 35-40 minutes each partition, compatible with an hourly ingestion cycle
* Meta cache update runs in 90 minutes total, compatible with once/twice daily cycle

Requires the following setup in prod before merging:

* MariaDB version update
* Table schema edits, as above
* One-time computation of `is_wip` and `is_latest_issue` for existing rows


sgratzl:
If I understand the `is_latest_issue` correctly -> doesn't that mean we could get rid of the sub query:

https://github.com/cmu-delphi/delphi-epidata/blob/dff64d0822159ed1d321256414112e9715b71bdb/src/server/api.php#L982-L991

krivard:
@sgratzl Yes, though I'm going to call that out of scope for this PR, mostly because staging this batch of changes has been more work than expected. Once it's merged we can do a followup to drop the subquery from the API server code.

PR #191, opened 2020-08-12
Enable compression but not caching

capnrefsmmat:
I think enabling expires will require `AllowOverride Index`, or something like it, in the main Apache conf. So let's settle for enabling compression alone.

It's a hack to only have the Expires directives commented-out, but hopefully we can soon figure out how to set the Apache config appropriately.

I've tested on staging and the compression seems to work.

### DIFF
--- a/src/server/.htaccess
+++ b/src/server/.htaccess
@@ -1,2 +1,6 @@
 # Open up the API to all origins.
 Header set Access-Control-Allow-Origin "*"
+
+# Compress API responses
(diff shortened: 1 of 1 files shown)


capnrefsmmat:
On staging, this takes the metadata query from 153 kB to 18 kB. A day of JHU case data goes from 409 kB to 13 kB. This should dramatically reduce bandwidth and improve performance.

PR #193, opened 2020-08-15
Edits throughout, added appendix

ryantibs:


ryantibs:
I wasn't able to compile the math, but Alex showed me a preview and I have a few things to fix.

PR #194, opened 2020-08-15
FB signals doc update

ryantibs:


capnrefsmmat:
I pushed some copyedits, but I need to check on this paragraph:

> When our geographical mapping data indicates that a ZIP code is part of multiple geographical units in a single aggregation, we assign weights to each of these units and proceed as described below, but with uniform participation weights ($$w^{\text{part}}_i=1$$ for all $$i$$).

This is false for the new pipeline (we don't use uniform participation weights; we use the true weights), but I don't know if it's true for the old pipeline.

capnrefsmmat:
I've pushed a commit correcting the error. @ryantibs This is ready to merge if you're happy with my changes.

ryantibs:
Look good.  Thank you!

ryantibs:
(I'll just note the obvious, which is that, when the new survey version goes out, we'll have to update this to reflect the new questions in the first section.)

PR #207, opened 2020-09-04
Support 'nation' as geo type with alpha values

krivard:
This was remarkably straightforward?

API ingestion hookups to support cmu-delphi/covidcast-indicators#192

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -225,7 +225,7 @@ class CsvImporter:
 
     elif geo_type == 'nation':
       # geo_id is lowercase
-      if not 'a' <= geo_id <= 'z':
+      if len(geo_id) != 2 or not 'aa' <= geo_id <= 'zz':
         return (None, 'geo_id')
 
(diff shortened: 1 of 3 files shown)


src/acquisition/covidcast/csv_importer.py
```diff
@@ -223,6 +223,11 @@ def extract_and_check_row(row, geo_type):
       if len(geo_id) != 2 or not 'aa' <= geo_id <= 'zz':
         return (None, 'geo_id')
 
+    elif geo_type == 'nation':
+      # geo_id is lowercase
+      if not 'a' <= geo_id <= 'z':
```
melange396, line 228: would we want this in 'aa' -> 'zz' (im presuming using standard international country codes (eg: us, ca, mx, gb, etc)

melange396 (approved):
small point, otherwise LGTM

src/acquisition/covidcast/csv_importer.py
krivard: good idea, done

PR #210, opened 2020-09-08
csv and raw json

sgratzl:
add support for `format=csv` and `format=json`. The former creates a CSV file the later a raw JSON array.

To support the different messages, http status codes (500, 401, 400) are used which will in an error case also send the json object back

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -79,6 +79,76 @@ class CovidcastTests(unittest.TestCase):
       'message': 'success',
     })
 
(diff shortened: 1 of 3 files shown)


krivard:
Tests pass, tests look reasonable, code looks good. Needs a documentation update to match. I'm looping in some folks to decide whether we want to do a provisional merge now and fill in documentation later.

krivard:
Merging now; we'll tell some of our users about it as a beta and then add it to the public documentation in a future release.

PR #213, opened 2020-09-09
bugfix

sgratzl:
bugfix array_filter returns an object instead of a plain array

appears when filtering the covid cast meta data

### DIFF
--- a/src/server/api.php
+++ b/src/server/api.php
@@ -1028,25 +1028,31 @@ function get_covidcast_meta() {
     $geo_types = extract_values($_REQUEST['geo_types'], 'str');
 
     if ($time_types !== null || $signals !== null || $geo_types !== null) {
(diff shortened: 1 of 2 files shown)


src/server/api.php
```diff
@@ -1028,25 +1028,31 @@ function get_covidcast_meta() {
     $geo_types = extract_values($_REQUEST['geo_types'], 'str');
 
     if ($time_types !== null || $signals !== null || $geo_types !== null) {
-      $epidata = array_filter($epidata, function($row) use(&$time_types, &$signals, &$geo_types) {
```
krivard, line 1031: This seems overly complicated -- why not use `$epidata = array_values($epidata)` instead?

krivard (approved):
fabulous

PR #216, opened 2020-09-13
Update hospital-admissions.md for claims averaging

mariajahja:
Include claims averaging in documentation. Not sure the last equation is necessary, I just wanted to be clear about the averaging (we are using a weighted average of the two streams, where the weight is given by the denominators/total admissions). If you think it's confusing, please feel free to remove.


capnrefsmmat:
Looks good, thanks.

PR #222, opened 2020-09-21
Set up a Jenkins pipeline and add the ability to deploy to the AWS app nodes

korlaxxalrok:
This PR adds the initial ability for Jenkins to trigger an Ansible playbook that
deploys the application code to the AWS nodes.

There are two main components:

- A Jenkinsfile (defines what Jenkins will do and when).
  - For now a merge to main triggers Jenkins to simply echo text. We can do
    more here later, but this initial setup is a workaround because there seems
    to be a bug in triggering another Jenkins job from a declarative pipeline.
- An independently configured Jenkins job that runs some Ansible.
  - This job is triggered whenever the declarative pipeline configured in the
    Jenkinsfiles runs.

A quick follow on task to reconfigure the independent job will be necessary
after this is merged.


capnrefsmmat:
I see a review request for me. Is there something specific you'd like me to check? I don't know too much about Jenkins and can't be terribly helpful.

korlaxxalrok:
Hi @capnrefsmmat Yep, not much to look at, just wanted to pass it by another set of eyes. I will remove you as reviewer and merge it 👍

PR #226, opened 2020-09-28
grant read permissions in Dockerfiles to various internal libraries

### LINKED ISSUE #225: Fix problem with host permissions in Docker definition
While debugging this error:

```
$ sudo docker run --rm -p 13306:3306   --network delphi-net --name delphi_database_epidata   delphi_database_epidata
[...]
2020-09-24 14:25:44-04:00 [Note] [Entrypoint]: /usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/api_analytics.sql
/usr/local/bin/docker-entrypoint.sh: line 72: /docker-entrypoint-initdb.d/api_analytics.sql: Permission denied
```

@brookslogan and Mike discovered that something seems to be amiss in the Docker definition for epidata that causes world to be denied all access to .sql files, like so:

```
$ sudo docker start delphi_database_epidata
delphi_database_epidata
$ sudo docker exec -it delphi_database_epidata /bin/bash
root@157020b36924:/# whoami
root
root@157020b36924:/#  ls -al docker-entrypoint-initdb.d
total 88
drwxr-xr-x 1 root root  4096 Sep 24 14:25 .
drwxr-xr-x 1 root root  4096 Sep 25 14:30 ..
-rwxr-x--- 1 root root  1262 Sep 21 15:25 api_analytics.sql
-rwxr-x--- 1 root root  3410 Sep 21 15:25 cdc.sql
-rwxr-x--- 1 root root  6435 Sep 21 15:25 covidcast.sql
-rwxr-x--- 1 root root  1193 Sep 21 15:25 ecdc_ili.sql
-rwxr-x--- 1 root root 15422 Sep 21 15:25 fluview.sql
-rwxr-x--- 1 root root   878 Sep 21 15:25 forecasts.sql
-rwxr-x--- 1 root root   887 Sep 21 15:25 gft.sql
-rwxr-x--- 1 root root  1016 Sep 21 15:25 ght.sql
-rwxr-x--- 1 root root  1160 Sep 21 15:25 kcdc_ili.sql
-rwxr-x--- 1 root root  2404 Sep 21 15:25 nidss.sql
-rwxr-x--- 1 root root   984 Sep 21 15:25 nowcasts.sql
-rwxr-x--- 1 root root  1727 Sep 21 15:25 paho_dengue.sql
-rwxr-x--- 1 root root  1093 Sep 21 15:25 quidel.sql
-rwxr-x--- 1 root root  1017 Sep 21 15:25 sensors.sql
-rwxr-x--- 1 root root   889 Sep 21 15:25 twitter.sql
-rwxr-x--- 1 root root  3984 Sep 21 15:25 wiki.sql
root@157020b36924:/# 
```

The correct permissions for world should be read-only.

Running `chmod o+r` on all the .sql files works, so fixing this permanently will require something like adding the following to the Dockerfile:

```
RUN chmod o+r /docker-entrypoint-initdb.d/*.sql
```

sgsmob:
Should resolve #225

### DIFF
--- a/dev/docker/web/epidata/Dockerfile
+++ b/dev/docker/web/epidata/Dockerfile
@@ -7,3 +7,6 @@ COPY repos/delphi/delphi-epidata/src/server/*.php /var/www/html/epidata/
 
 # point to the development database (overwrites the production config)
 COPY repos/delphi/delphi-epidata/dev/docker/web/epidata/assets/database_config.php /var/www/html/epidata/
+
+RUN chmod o+r /var/www/html/secrets.php
+RUN chmod o+r /var/www/html/epidata/*.php
--- a/dev/docker/web/epidata/Dockerfile
+++ b/dev/docker/web/epidata/Dockerfile
@@ -8,5 +8,4 @@ COPY repos/delphi/delphi-epidata/src/server/*.php /var/www/html/epidata/
 # point to the development database (overwrites the production config)
 COPY repos/delphi/delphi-epidata/dev/docker/web/epidata/assets/database_config.php /var/www/html/epidata/
 
-RUN chmod o+r /var/www/html/secrets.php
 RUN chmod o+r /var/www/html/epidata/*.php
--- a/dev/docker/web/epidata/Dockerfile
+++ b/dev/docker/web/epidata/Dockerfile
@@ -8,4 +8,4 @@ COPY repos/delphi/delphi-epidata/src/server/*.php /var/www/html/epidata/
 # point to the development database (overwrites the production config)
 COPY repos/delphi/delphi-epidata/dev/docker/web/epidata/assets/database_config.php /var/www/html/epidata/
 
-RUN chmod o+r /var/www/html/epidata/*.php
+RUN chmod o+r /var/www/html/epidata/*
--- a/dev/docker/database/epidata/Dockerfile
+++ b/dev/docker/database/epidata/Dockerfile
@@ -10,3 +10,6 @@ ENV MYSQL_PASSWORD pass
 
 # provide DDL which will create empty tables at container startup
 COPY repos/delphi/delphi-epidata/src/ddl/*.sql /docker-entrypoint-initdb.d/
+
+# grant access to SQL scripts
+RUN chmod o+r /docker-entrypoint-initdb.d/*.sql


krivard:
@korlaxxalrok check the read access for world on the secrets.php file -- is that needed? bad?

dev/docker/web/epidata/Dockerfile
```diff
@@ -7,3 +7,6 @@ COPY repos/delphi/delphi-epidata/src/server/*.php /var/www/html/epidata/
 
 # point to the development database (overwrites the production config)
 COPY repos/delphi/delphi-epidata/dev/docker/web/epidata/assets/database_config.php /var/www/html/epidata/
+
+RUN chmod o+r /var/www/html/secrets.php
+RUN chmod o+r /var/www/html/epidata/*.php
```
dfarrow0, line 12: do the other files under `/var/www/html/epidata/` need to be updated also? (html files are copied there too on line 5)

dev/docker/web/epidata/Dockerfile
```diff
@@ -7,3 +7,6 @@ COPY repos/delphi/delphi-epidata/src/server/*.php /var/www/html/epidata/
 
 # point to the development database (overwrites the production config)
 COPY repos/delphi/delphi-epidata/dev/docker/web/epidata/assets/database_config.php /var/www/html/epidata/
+
+RUN chmod o+r /var/www/html/secrets.php
```
dfarrow0, line 11: the `secrets.php` file is added by a separate dockerfile here: https://github.com/cmu-delphi/operations/blob/main/dev/docker/web/Dockerfile

i recommend updating permissions there instead of here, because there are other docker images besides this one that are built on top of that.

dfarrow0 (changes requested):
unfortunately my local install of docker is broken, so i can't reproduce the problem or test the fix. however, this looks reasonable to me.

dfarrow0:
> @korlaxxalrok check the read access for world on the secrets.php file -- is that needed? bad?

it would be needed if you build as one user (e.g. root) and run as another (e.g. your local user account), which is probably the case, but i can't check until i fix my docker.

it's not bad from a local development perspective -- it's a copy of a file in your local repo.

@korlaxxalrok are you using these images server-side?

korlaxxalrok:
@dfarrow0 @krivard Nope, not using them server-side so this should be ok.

dev/docker/web/epidata/Dockerfile
krivard: If we update the permissions in operations, will that propagate here?

dfarrow0 (approved):
lgtm, thanks!

dfarrow0:
@krivard @korlaxxalrok 

requesting merge when you get a chance please. hoping to unblock others who are onboarding.

PR #230, opened 2020-10-01
csv file loading with explicit 'issue' dates

melange396:
still needs some tests and documentation, but this should do the trick

### DIFF
--- a/src/acquisition/covidcast/csv_to_database.py
+++ b/src/acquisition/covidcast/csv_to_database.py
@@ -17,9 +17,78 @@ def get_argument_parser():
   parser.add_argument(
     '--data_dir',
     help='top-level directory where CSVs are stored')
+  parser.add_argument(
+    '--specific_issue_date',
+    action='store_true',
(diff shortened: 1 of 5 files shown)


src/acquisition/covidcast/csv_to_database.py
```diff
@@ -17,9 +17,78 @@ def get_argument_parser():
...
+    action='store_true',
+    help='overrides all signals to mark them as WIP.  NOTE: specify neither or only one of --is_wip_override and --not_wip_override.')
+  parser.add_argument(
+    '--not_wip_override',
+    action='store_true',
+    help='overrides all signals to mark them as *not* WIP.  NOTE: specify neither or only one of --is_wip_override and --not_wip_override.')
   return parser
 
+def scan_issue_specific(data_dir, database, is_wip_override=None, csv_importer_impl=CsvImporter, file_archiver_impl=FileArchiver):
+  """  is_wip_override: None (default) for standard 'wip_' prefix processing, True for forcing all as WIP, False for forcing all as not WIP"
```
krivard, line 36: this is glorious, bless you for thinking of it

PR #234, opened 2020-10-05
Wave 4 survey signal documentation

capnrefsmmat:
Document the signals that are ready for public release.


krivard (approved):
the lack of raw behavioral signals threw me for a sec, but it all appears correct

capnrefsmmat:
Yes, I made the executive decision that the raw signals aren't very useful, but we can revisit and add them if we want. But I expect sample size will not increase, only decrease, so the raw signals aren't going to get *more* useful.

PR #236, opened 2020-10-07
feat: add min_issue to covidcast meta data

### LINKED ISSUE #232: Add min_issue to covidcast metadata endpoint
The endpoint contains max_issue, but does not report min_issue, so users don't know what range they can do `as_of` queries for. We should add this to the metadata calculation and the API endpoint, then update the clients to ensure they return it in correct format (as a date object rather than a string).

sgratzl:
closes #232

+ removes some unneeded whitespace


krivard:
Fails integration tests:

```
======================================================================
FAIL: test_round_trip (repos.delphi.delphi-epidata.integrations.server.test_covidcast_meta.CovidcastMetaTests)
Make a simple round-trip with some sample data.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/app/repos/delphi/delphi-epidata/integrations/server/test_covidcast_meta.py", line 87, in test_round_trip
    self.assertEqual(response, {
AssertionError: {'res[250 chars]ue': 1, 'max_issue': 2, 'min_lag': 0, 'max_lag[4262 chars]ess'} != {'res[250 chars]ue': 2, 'max_issue': 2, 'min_lag': 0, 'max_lag[4262 chars]ess'}
Diff is 22152 characters long. Set self.maxDiff to None to see it.

======================================================================
FAIL: test_suppress_work_in_progress (repos.delphi.delphi-epidata.integrations.server.test_covidcast_meta.CovidcastMetaTests)
Don't surface signals that are a work-in-progress.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/app/repos/delphi/delphi-epidata/integrations/server/test_covidcast_meta.py", line 257, in test_suppress_work_in_progress
    self.assertEqual(response, {
AssertionError: {'res[250 chars]ue': 1, 'max_issue': 2, 'min_lag': 0, 'max_lag[4262 chars]ess'} != {'res[250 chars]ue': 2, 'max_issue': 2, 'min_lag': 0, 'max_lag[4262 chars]ess'}
Diff is 22152 characters long. Set self.maxDiff to None to see it.
```

src/acquisition/covidcast/database.py
```diff
@@ -576,6 +576,7 @@ def get_covidcast_meta(self):
         ROUND(AVG(`value`),7) AS `mean_value`,
         ROUND(STD(`value`),7) AS `stdev_value`,
         MAX(`value_updated_timestamp`) AS `last_update`,
+        MIN(`issue`) as `min_issue`,
```
melange396, line 579: i believe this is not going to work as intended, and instead will show the issue date of the least-recently-updated of the issues in a group, which i would argue is meaningless (plz correct me if im wrong). i think it will require a MIN(issue) inside the sub-SELECT and appropriate handling outside of that.

also, some non-trivial test cases would be nice, with different issue values.

melange396 (changes requested):
i believe this is not going to work as intended, and instead will show the issue date of the least-recently-updated of the issues in a group, which i would argue is meaningless (plz correct me if im wrong). i think it will require a MIN(issue) inside the sub-SELECT and appropriate handling outside of that.

also, some non-trivial test cases would be nice, with different issue values.

src/acquisition/covidcast/database.py
sgratzl: true, I will think about it how to fix it. Moreover, we can think of using the `is_latest_issue` to find the max issue

sgratzl: @melange396 can you take a look at the new statements. I split it into two, since I couldn't think of a fast way to handle both cases at the same time. Moreover, I switched the nested statement to the `is_latest_issue` flag. It is not that easy to test for me, since I don't have access to the real database but just local test data

melange396: i suppose this will work, but the old SQL query would have done the trick if you added `min(`issue`) `min_issue`` to the nested select...  i think that would be preferable to adding a bunch of extra lines unless you can think of another benefit that justifies the extra complexity

src/acquisition/covidcast/database.py
```diff
@@ -560,68 +560,69 @@ def get_covidcast_meta(self):
...
+        `covidcast`
+      WHERE
+        `source` = %s AND
+        `signal` = %s
+      GROUP BY
+        `time_type`,
+        `geo_type`
+      '''
+      self._cursor.execute(sql, (source, signal))
+      min_issue_lookup = {f'{x[0]}:{x[1]}': x[2] for x in self._cursor}
```
melange396, line 582: you can use a tuple as the key for a dict.  it probably wouldnt make much (if any) of a performance difference here, but it would make for a shorter comprehension.  its really just a style judgement call either way ¯\_(ツ)_/¯

sgratzl: done

src/acquisition/covidcast/database.py
sgratzl: yes but I think this version should be more efficient using `is_latest_issue` instead of a join

melange396: after wrestling with and waiting for long-duration queries for the past week, i think it might be wise to test the performance of the one-query vs two-query approaches on the staging server.  i can certainly help do this if necessary.

krivard:
Putting this back on your radar @melange396 now that the JHU fires are out

src/acquisition/covidcast/database.py
krivard: Apologies for being so late to this conversation. 

The use case for `min_issue` is briefly described in [the source issue for this PR](https://github.com/cmu-delphi/delphi-epidata/issues/232), but the jist is this: A researcher is constructing `as_of` queries and wants to know the earliest date they can pass in and still expect a non-nil result. This means that `min_issue` _must not_ be restricted to rows with `is_latest_issue=1`, since that would hide issues which are not the most recent but nevertheless accessible using `as_of`.

melange396: i will go ahead and try this and other variants out on the staging copy of the database to check performance and runtimes

sgratzl:
closing this PR since meta generation has changed since then

Issue #195, opened 2020-08-23
Allow encoding of non-numbers (e.g. missingness, deletions, ranges)


krivard:
Since moving to a data versioning scheme, there is no longer any way to remove a row from COVIDcast without removing all previous versions of that row as well (so that it's as if it was never published at all). This is hazardous -- leaving the row in is inaccurate, and removing the row gives forecasters access to future-privileged information that will not match realtime usage.

We are developing a survey of different kinds of missingness and deletions that occur in the different COVIDcast sources [here](https://docs.google.com/spreadsheets/d/1Q8sNBeYwvdBhvEPFkGRsqJHiNveGK8eyEy-Fv91JSFk/edit?usp=sharing) to help spec out an encoding system.

Some additional conversation on this is in [a thread on the first set of performance fixes](https://github.com/cmu-delphi/covidcast-indicators/issues/158#issuecomment-663167217), but it looks like the column additions mentioned there didn't actually make it into staging this time around.

RoniRos:
Suppressed/censored values are a common facet of epidemiological data, so we need to build support for it into Epidata.  This means a variety of non-numeric values.  In the case of HHS's Hospitalization By Facility data stream, it will be a value that means "a number in the range [0-3]".  There are cases where zero could be singled out, e.g. in Change Healthcare data we have a value for "a number in the range [1-3]".  In addition to supporting numerical ranges, there are a variety of other possible values ("censored", "unreported", "N/A", etc.).

Implementation:

My suggestion is that we maintain an open-ended and ever growing list of special values.  To start with, we'll have all the types of missingness and deletions that we have encountered, as well as all types of numeric ranges we encountered or expect to encounter (e.g. [0-3], [1-3] etc.).  Our code would then be able to ask about specific values and do the right thing with them.

As far as how to represent these values in a database, there are likely known solutions that have been worked out for any self-respecting DB that supports missingness.  But since I don't know what they are, let me suggest something very basic:  We probably don't want to use a separate non-numeric stream of values, because it will double the size of the database.  Instead, we should expropriate a range of numerical values.  With an integer stream, it would have been ideal to expropriate the 1000 largest negative or positive numbers represented by a 64 bit integer, but this is too tied to low level architecture.  Instead, we can expropriate the smallest 1000 negative numbers, namely -1 thru -1000.  This will require that if we actually use negative integers in our streams, we substract 1000 from them before storage, and add 1000 upon retrieval.   This could be generalized to floating point data as well.

[...]

PR #237, opened 2020-10-07
feat: support geo_values (single geo_type)

### LINKED ISSUE #221: Accept a list of geo_values instead of only a single value
Currently, the covidcast endpoint accepts a single `geo_value` that can either be `*` (all locations) or the identifier of a single location. That means that if an API client wants to fetch, say, the top 500 counties by case counts, they either have to send one query and filter it or send 500 separate queries. The R and Python clients currently do the latter, but it's slow. (Fetching one day of data for 500 counties would take as long as fetching all state data for 500 days -- making it unusably slow if you want more than one or two days of data.)

Can `geo_value` take a comma-separated list, like the issues and time_values parameters?

For new people, the code lives here: https://github.com/cmu-delphi/delphi-epidata/blob/main/src/server/api.php#L940

sgratzl:
closes #221 (single geo type only for now)

add support for `geo_values`.

valid variants:
 * `geo_value=*`
 * `geo_value=CA`
 * `geo_values=CA,NY`

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -311,6 +311,100 @@ class CovidcastTests(unittest.TestCase):
       'message': 'success',
     })
 
+  def test_geo_value(self):
+    """test different variants of geo types: single, *, multi."""
+
+    # insert dummy data
+    self.cur.execute('''
+      insert into covidcast values
+        (0, 'src', 'sig', 'day', 'county', 20200414, '11111',
+          123, 10, 11, 12, 456, 13, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'county', 20200414, '22222',
+          123, 20, 21, 22, 456, 23, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'county', 20200414, '33333',
+          123, 30, 31, 32, 456, 33, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'msa', 20200414, '11111',
+          123, 40, 41, 42, 456, 43, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'msa', 20200414, '22222',
+          123, 50, 51, 52, 456, 53, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'msa', 20200414, '33333',
+          123, 60, 61, 62, 456, 634, 20200414, 0, 1, False)
+    ''')
+    self.cnx.commit()
+
+    def fetch(geo_value):
+      # make the request
+      params = {
+        'source': 'covidcast',
+        'data_source': 'src',
+        'signal': 'sig',
+        'time_type': 'day',
+        'geo_type': 'county',
+        'time_values': 20200414,
+      }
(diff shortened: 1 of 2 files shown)


src/server/api.php
```diff
@@ -959,12 +958,16 @@ function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $
...
 
-  if ($geo_value === '*') {
+  if ($geo_values === '*') {
     // the wildcard query should return data for all locations in `geo_type`
     $condition_geo_value = 'TRUE';
+  } else if (is_array($geo_values)) {
+    // return data for multiple location
+    $condition_geo_value = filter_strings('t.`geo_value`', $geo_values);
   } else {
     // return data for a particular location
```
sgsmob, line 968: If a `'*'` is present in the list will it be interpreted literally (matching `geo_values` that are `'*'`)?

integrations/server/test_covidcast.py
```diff
@@ -311,6 +311,100 @@ def test_location_wildcard(self):
...
+    self.assertEqual(r['message'], 'no results')
+    r = fetch(['11111', '22222'])
+    self.assertEqual(r['message'], 'success')
+    self.assertEqual(r['epidata'], [counties[0], counties[1]])
+    r = fetch(['11111', '33333'])
+    self.assertEqual(r['message'], 'success')
+    self.assertEqual(r['epidata'], [counties[0], counties[2]])
+    r = fetch(['11111', '55555'])
+    self.assertEqual(r['message'], 'success')
+    self.assertEqual(r['epidata'], [counties[0]])
```
krivard, line 403: This is probably the right choice (silently omit bogus geo values from the response) but we should revisit it when we figure out deletion encoding. One of the goals of deletion encoding is to start giving better feedback about "you asked for a bogus county" vs "that information was withheld for privacy reasons", and an error on *parts* of a query might have to be treated carefully.

krivard (approved):
Everything builds, tests pass, just added note so we don't forget about partial-failures when we fix #195

integrations/server/test_covidcast.py
```diff
@@ -311,6 +311,100 @@ def test_location_wildcard(self):
...
+      'value': 30,
+      'stderr': 31,
+      'sample_size': 32,
+      'direction': 33,
+      'issue': 20200414,
+      'lag': 0,
+      'signal': 'sig',
+    }]
+
+    r = fetch('*')
```
sgsmob, line 387: Please add comments to each call to `fetch()` indicating the specific behavior being tested.

sgsmob (changes requested):

integrations/server/test_covidcast.py
sgratzl: I think it behaves the same way for all kind of parameters that are not existing, e.g., if you specify a wrong data source you just get no results back but it doesn't throw an error. That is the current nature of just generating a where clause

src/server/api.php
sgratzl: atm. yes, there is the clear distinction see PR comment about the variants

sgsmob (approved):

PR #238, opened 2020-10-07
Add 127.0.0.1: to beginning of docker run commands

benjaminysmith:
Add 127.0.0.1: to beginning of docker run commands to only listen on localhost (not externally) which enhances security.


krivard (approved):
Works!

PR #245, opened 2020-10-12
add geo_values support in raw clients

### LINKED ISSUE #241: Support multiple geo_values in raw clients
* Python: https://github.com/cmu-delphi/delphi-epidata/blob/e24ae513a8106d39e540e5f6d0f2cef4324f002a/src/client/delphi_epidata.py#L557
* R: https://github.com/cmu-delphi/delphi-epidata/blob/e24ae513a8106d39e540e5f6d0f2cef4324f002a/src/client/delphi_epidata.R#L506
* Coffee: https://github.com/cmu-delphi/delphi-epidata/blob/e24ae513a8106d39e540e5f6d0f2cef4324f002a/src/client/delphi_epidata.coffee#L359

The JavaScript client is generated using the following procedure:
```
coffee --compile --transpile delphi_epidata.coffee
```

`--transpile` is necessary because the delphi backend doesn't believe in ES6.

sgratzl:
closes #241

### DIFF
--- a/src/client/delphi_epidata.js
+++ b/src/client/delphi_epidata.js
@@ -573,9 +573,14 @@ function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _d
             'signals': signals,
             'time_type': time_type,
             'geo_type': geo_type,
-            'time_values': _list(time_values),
-            'geo_value': geo_value
+            'time_values': _list(time_values)
           };
+
+          if (Array.isArray(geo_value)) {
+            params.geo_values = geo_value.join(',');
+          } else {
+            params.geo_value = geo_value;
+          }
 
           if (as_of != null) {
             params.as_of = as_of;
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -573,10 +573,13 @@ class Epidata:
       'signals': Epidata._list(signals),
       'time_type': time_type,
       'geo_type': geo_type,
-      'time_values': Epidata._list(time_values),
-      'geo_value': geo_value,
+      'time_values': Epidata._list(time_values)
     }
 
+    if isinstance(geo_value, (list, tuple)):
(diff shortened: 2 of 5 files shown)


src/client/delphi_epidata.js
```diff
@@ -573,10 +573,15 @@ function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _d
...
-            'geo_value': geo_value
+            'time_values': _list(time_values)
           };
 
+          if (Array.isArray(geo_value)) {
+            params.geo_values = geo_value.join(',');
+          } else {
+            params.geo_value = geo_value;
+          }
+
```
krivard, line 584: Would `_list()` be able to handle this, and save some duplication?

src/client/delphi_epidata.py
```diff
@@ -573,10 +573,13 @@ def covidcast(
...
       'geo_type': geo_type,
-      'time_values': Epidata._list(time_values),
-      'geo_value': geo_value,
+      'time_values': Epidata._list(time_values)
     }
 
+    if isinstance(geo_value, (list, tuple)):
+      params['geo_values'] = ','.join(geo_value)
+    else:
+      params['geo_value'] = geo_value
```
krivard, line 582: Same question -- why not `Epidata._list()`?

krivard (commented):
Clarify handling choices

src/client/delphi_epidata.py
sgratzl: in the current version changing it won't change much, since the array and plain version needs another parameter name `geo_value` vs `geo_values`. Thus, the `if` will remain.

src/client/delphi_epidata.js
sgratzl: in the current version changing it won't change much, since the array and plain version needs another parameter name `geo_value` vs `geo_values`. Thus, the `if` will remain.

PR #246, opened 2020-10-12
feat: use is_latest_issue in PHP

### LINKED ISSUE #223: Refactor api.php to use latest scheme features: is_latest_issue
atm. there is still a slow sub query that checks that manually

sgratzl:
closes #223 

in case no special flag is set (issues, as_of, lag) it will simplify the query to `is_latest_issue is TRUE` instead of a complex sub query

I would be curious how much faster this query is compared to the previous one. However, I cannot test it since my local database doesn't have enough data to be able to observe an effect

### DIFF
--- a/src/server/api.php
+++ b/src/server/api.php
@@ -976,21 +976,21 @@ function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $
     //build the issue filter
     $condition_issue = filter_integers('t.`issue`', $issues);
     $condition_version = $condition_issue;
-  } else if($lag !== null) {
+  } else if ($lag !== null) {
     //build the lag filter
     $condition_lag = "(t.`lag` = {$lag})";
     $condition_version = $condition_lag;
-  } else {
-    //fetch most recent issues
-    $sub_condition_asof = "TRUE";
-    if ($as_of !== null) {
-      $sub_condition_asof = "(`issue` <= {$as_of})";
-    }
+  } else if ($as_of !== null) {
+    // fetch most recent issues with as of
+    $sub_condition_asof = "(`issue` <= {$as_of})";
     $sub_fields = "max(`issue`) `max_issue`, `time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
     $sub_group = "`time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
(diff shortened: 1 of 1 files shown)


src/server/api.php
```diff
@@ -976,21 +976,21 @@ function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $
...
+    // fetch most recent issues with as of
+    $sub_condition_asof = "(`issue` <= {$as_of})";
     $sub_fields = "max(`issue`) `max_issue`, `time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
     $sub_group = "`time_type`, `time_value`, `source`, `signal`, `geo_type`, `geo_value`";
     $sub_condition = "x.`max_issue` = t.`issue` AND x.`time_type` = t.`time_type` AND x.`time_value` = t.`time_value` AND x.`source` = t.`source` AND x.`signal` = t.`signal` AND x.`geo_type` = t.`geo_type` AND x.`geo_value` = t.`geo_value`";
     $subquery = "JOIN (SELECT {$sub_fields} FROM {$table} WHERE ({$conditions} AND {$sub_condition_asof}) GROUP BY {$sub_group}) x ON {$sub_condition}";
     $condition_version = 'TRUE';
+  } else {
+    // fetch most recent issue fast
+    $condition_version = '(t.`is_latest_issue` IS TRUE)';
```
benjaminysmith, line 993: It looks like this is new query logic -- are there tests for this to make sure this works correctly? (the automated test link in the conversation tab fails to load for me)

sgratzl: the query should behave like the old one but just simpler and faster. Thus, the existing unit tests should already cover this change.

benjaminysmith: The logic makes sense. If there are existing tests that's great, was wondering if needed to follow up with any new/additional tests.  I'm still trying to understand the code structure and test layout and didn't see anything obvious under delphi-epidata/tests/ -- is there another location I should check?

dfarrow0: Quick notes on testing:

- `tests/` contains unit tests for python code
- `integrations/` contains integration tests (differentiated from unit tests by involvement of external services, like database and web server) for python, sql, and php

Both types of tests are written in python and are run by `py3tester`, which is a thin wrapper around python's built-in unit testing framework. In addition to pass/fail, for unit tests specifically, `py3tester` provides line-based coverage and profiling.

For this particular file, `api.php`, the only test coverage is in the form of integration tests which live at [`integrations/server/`](https://github.com/cmu-delphi/delphi-epidata/tree/main/integrations/server).

melange396 (approved):
LGTM.  i imagine this will save a lot of DB thrashing

src/server/api.php
benjaminysmith: Thanks David -- it looks like this changes the default case so is exercised by integrations/server/test_covidcast.py.

benjaminysmith (approved):

PR #249, opened 2020-10-14
Make sure that exceptions are reraised if the SQL insert fails.

benjaminysmith:
Currently the return statement is inside of the `finally` block, this means that if an exception is caught and re-raised by the `except` block, it will be cached but then immediately discarded -- i.e., the function will never re-raise the exception. This will cause inserts that result in an exception to appear successful (though I believe with 0 lines inserted).

See https://stackoverflow.com/questions/43830803/re-raising-the-exception-doesnt-work-with-finally-clause for more context on the try/except/finally behavior.

This is an additional fix for #203.

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -229,7 +229,7 @@ class Database:
       raise e
     finally:
       self._cursor.execute(drop_tmp_table_sql)
-      return total
+    return total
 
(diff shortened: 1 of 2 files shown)


melange396 (approved):
LGTM.  nice catch!  (pun was not intended, but i will just go with it)

PR #250, opened 2020-10-14
deprecating server-side direction computation

melange396:


src/acquisition/covidcast/database.py
```diff
@@ -106,6 +106,7 @@ def insert_or_update_batch(self, cc_rows, batch_size=2**20, commit_partial=False
 
     tmp_table_name = 'tmp_insert_update_table'
 
+    # TODO: this heavily copypastas src/ddl/covidcast.sql -- theres got to be a better way
```
benjaminysmith, line 109: Is it worth creating an issue for tracking this or do  you typically leave as todo?

benjaminysmith (approved):
I love seeing deleted code!

It looks like this changes the API -- I'm guessing we leave the data in the DB?

krivard:
@benjaminysmith Yep, we'll leave the data in the DB until we're ready to do a database migration. We're not as good at doing that without substantial service interruptions, so it will probably need to wait until we're less in the PR spotlight from the mask data release from Monday.

krivard (approved):
Looks great!

src/acquisition/covidcast/database.py
krivard: Leaving a TODO essentially creates a lowest-priority task. If it becomes more obviously important we can escalate to an issue.

PR #252, opened 2020-10-16
update dev guides

dfarrow0:
- mainly adding documentation around bind-mounting local sources to prevent the need to rebuild images all the time
- also update some links


krivard (approved):
nice work

Issue #255, opened 2020-10-19
Document DV's known holiday effects in its signal Limitations section


capnrefsmmat:
The Doctor Visits data is known to have spikes around major holidays (July 4th, Labor Day) because of different medical-seeking behavior on those days. If we can't get rid of this, we should at least document it in the Limitations section of the docs: https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/api/covidcast-signals/doctor-visits.md#limitations

PR #256, opened 2020-10-19
Document DV holiday effects

rumackaaron:
Addresses #255

### DIFF
--- a/docs/api/covidcast-signals/doctor-visits.md
+++ b/docs/api/covidcast-signals/doctor-visits.md
@@ -52,9 +52,10 @@ States.
 Standard errors are not available for this data source.
 
 Due to changes in medical-seeking behavior on holidays, this data source has
-upward spikes in CLI around major holidays (i.e. Memorial Day, July 4, Labor
-Day, etc.). These spikes are not necessarily indicative of a true increase of
-COVID-like illness in a location.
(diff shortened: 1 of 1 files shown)


docs/api/covidcast-signals/doctor-visits.md
```diff
@@ -51,6 +51,11 @@ States.
 
 Standard errors are not available for this data source.
 
+Due to changes in medical-seeking behavior on holidays, this data source has
+upward spikes in CLI around major holidays (i.e. Memorial Day, July 4, Labor
```
capnrefsmmat, line 55:
```suggestion
upward spikes in the fraction of doctor's visits that are COVID-related around major holidays (e.g. Memorial Day, July 4, Labor
```

1. i.e. is "that is", i.e. to rephrase something another way, whereas e.g. is for examples, e.g. a list of holidays
2. looks like this doc doesn't define CLI by this point, so may as well be more explicit

capnrefsmmat (changes requested):
Looks good, just one suggested change you can apply and then I'll merge

docs/api/covidcast-signals/doctor-visits.md
rumackaaron: You should be an English professor!

capnrefsmmat (approved):

PR #260, opened 2020-10-22
JHU is_latest fix w/ join

melange396:
generates raw SQL that for JHU data, by month and by signal, finds latest issues' ids and saves them in a temp table.  then by month and by signal, clears is_latest_issue flags.  then sets is_latest_issue flag for all saved ids.  then drops temp table.


krivard (approved):
lgtm, epic work on this

PR #265, opened 2020-10-28
TEMPORARY TABLE is evil

melange396:
in practice, we will probably not see issues with this unless we get some insanely large file to process, but with recent issues, i think this prudent.  there may be some performance drawbacks to this, but seeing as how the failure mode is silent, i think this is worth it.

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -108,7 +108,7 @@ class Database:
 
     # TODO: this heavily copypastas src/ddl/covidcast.sql -- theres got to be a better way
     create_tmp_table_sql = f'''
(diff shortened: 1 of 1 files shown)


melange396:
there are also TEMPORARY TABLE lines in `src/acquisition/norostat/norostat_sql.py` but i dont know if that stuff is in the scope of our development effort right now.  also, it would probably necessitate adding TRUNCATE or DROP statements to keep things clean.

krivard:
`norostat` is a substantially smaller dataset, so it's probably fine to leave it out

krivard (approved):
Unit and integration tests pass; all good

PR #266, opened 2020-10-30
Reorganize Epidata API docs to make hierarchy clearer

capnrefsmmat:
Based on comments by @RoniRos about the hierarchy being confusing here, with the site being named Epidata API, containing separate sections for COVIDcast (which is part of the Epidata API) and the Epidata API again.

Preview: http://rosmarus.refsmmat.com/delphi-epidata/

cc @dfarrow0


Akvannortwick (approved):
The edits are useful in creating distinctions between API content

PR #272, opened 2020-11-06
Mention smoothed fb-survey signals closer to raw signal discussion

capnrefsmmat:
Raw signals are not listed for the other indicators, so we only need this comment for the CLI/ILI signals.

cc @dshemetov, who couldn't find the information he needed because of this issue.

### DIFF
--- a/docs/api/covidcast-signals/fb-survey.md
+++ b/docs/api/covidcast-signals/fb-survey.md
@@ -77,14 +77,14 @@ influenza-like illness (fever, along with cough or sore throat). Using this
 survey data, we estimate the percentage of people who have a COVID-like illness,
 or influenza-like illness, in a given location, on a given day.
 
-| Signal | Description |
+| Signals | Description |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


Akvannortwick (commented):
This looks good, but I would keep the Smoothing sub-header.

capnrefsmmat:
@Akvannortwick Do you mean on the paragraph beginning "The signals beginning with `smoothed_`"?

Akvannortwick:
Yes, I feel it gives a nice separation in the text that doesn't affect flow. It's more personal preference, though there is no issue without it.

docs/api/covidcast-signals/fb-survey.md
```diff
@@ -115,6 +108,12 @@ can be asymptomatic. Instead, we expect these indicators to be useful for
 comparison across the United States and across time, to determine where symptoms
 appear to be increasing.
 
+The signals beginning with `smoothed_` estimate the same quantities as their
```
capnrefsmmat, line 111:
```suggestion
**Smoothing.** The signals beginning with `smoothed_` estimate the same quantities as their
```

PR #277, opened 2020-11-11
Revise changehc signal names

krivard:
Roni had a change to the signal names that didn't make it into the last PR


krivard:
Superseded by #278

PR #282, opened 2020-11-12
Update Changelog for DV

Akvannortwick:


krivard (changes requested):
DV receives FIPS (county) data from the source, not zip

krivard (approved):
👍

PR #283, opened 2020-11-13
Improve data source template docs

capnrefsmmat:
Most importantly, add a licensing section with options for different
circumstances.

This is a draft. Once we agree the language here is good, I'll add licensing details to all our current data sources.

### DIFF
--- a/docs/api/covidcast-signals/_source-template.md
+++ b/docs/api/covidcast-signals/_source-template.md
@@ -8,7 +8,7 @@ grand_parent: COVIDcast Epidata API
 {: .no_toc}
 
 * **Source name:** `SOURCE-API-NAME`
-* **First issued:** DATE RELEASED TO API
+* **First issued:** FIRST ISSUE DATE IN API
 * **Number of data revisions since 19 May 2020:** 0
 * **Date of last change:** Never
 * **Available for:** county, hrr, msa, state (see [geography coding docs](../covidcast_geography.md))
@@ -72,3 +72,30 @@ backfill here.
 If the signal has specific licensing or sourcing that should be acknowledged,
 describe it here. Also, include links to source websites for data that is
 scraped or received from another source.
+
+## Licensing
+
+### For signals constructed by Delphi:
+
+This data is provided under the terms of the [Creative Commons Attribution
+license](https://creativecommons.org/licenses/by/4.0/). If you use this data,
+you must cite [Delphi](https://delphi.cmu.edu/) COVIDcast API, preferably by
+providing a link to this page.
+
+### For signals republished from other sources:
+
+This data is originally published by SOURCE NAME under the terms LICENSE. Cite
+the source as
+
+> SOURCE'S PREFERRED ATTRIBUTION.
+
+If you use the Delphi COVIDcast API to access this data, we additionally request
+you acknowledge us in your product or publication.
+
+### For signals from the United States government:
+
+This data is originally published by AGENCY NAME. As a work of the United States
+Federal Government, it is in the public domain.
+
+If you use the Delphi COVIDcast API to access this data, we request you
+acknowledge us in your product or publication.


docs/api/covidcast-signals/_source-template.md
```diff
@@ -72,3 +72,30 @@ backfill here.
...
 scraped or received from another source.
+
+## Licensing
+
+### For signals constructed by Delphi:
+
+This data is provided under the terms of the [Creative Commons Attribution
+license](https://creativecommons.org/licenses/by/4.0/). If you use this data,
+you must cite [Delphi](https://delphi.cmu.edu/) COVIDcast API, preferably by
+providing a link to this page.
```
dfarrow0, line 83: Tiny nit around linking: with Delphi being linked earlier in this sentence, it initially reads like "you need to link back to the delphi website". But if you keep reading, then you realize that the Delphi website is incidental and really "you need to link to the current page". It might be more straightforward to remove the link to Delphi.

capnrefsmmat: I think I changed my mind about linking halfway through writing that. Do you think linking to the current page is the best idea? It seemed sound to me (so you're always providing a link to the detailed information about the data), but maybe if we publish a paper about the API we can change the text to point there.

krivard: Optimal approach would be to just provide boilerplate citation text.

capnrefsmmat: How about this?

> If you use this data, you must cite Delphi's COVIDcast API. For example:
>
> > Data from the COVIDcast API, by the Delphi group at Carnegie Mellon University. <https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/SOURCENAME.html>

docs/api/covidcast-signals/_source-template.md
```diff
@@ -72,3 +72,30 @@ backfill here.
...
 describe it here. Also, include links to source websites for data that is
 scraped or received from another source.
+
+## Licensing
+
+### For signals constructed by Delphi:
+
+This data is provided under the terms of the [Creative Commons Attribution
+license](https://creativecommons.org/licenses/by/4.0/). If you use this data,
+you must cite [Delphi](https://delphi.cmu.edu/) COVIDcast API, preferably by
```
dfarrow0, line 82: I think we should sync as a group on our approach to branding this. (Or maybe you already have and I just missed it.) I've seen the `covidcast` API endpoint referred to by a couple of different names, and I think it would be good to settle on one.

Is there a plan to separate COVIDcast from Epidata? I assume not, but correct me if I'm wrong. In my mind, "the COVIDcast API" is really just one particular endpoint (i.e. `covidcast`) of the Epidata API, and COVIDcast isn't an API in itself. I think it would be most accurate in this sentence to replace "Delphi COVIDcast API" with "Delphi Epidata API", or if that's not specific enough then maybe "Delphi Epidata API for COVIDcast".

There's also the matter of COVIDcast R/Python clients which use the Epidata API behind the scenes. I would call those "SDKs" or "libraries" interchangeably, and might suggest citing them separately as the "COVIDcast SDK" or "COVIDcast client library". As in:

> The `COVIDcast client library` provides a lot of functionality, including, in part, the ability to fetch COVID-19-related data from the `Delphi Epidata API`, which is also surfaced though interactive maps and charts in the `COVIDcast` web application.

@ryantibs @RoniRos @capnrefsmmat interested to hear your thoughts.

krivard: Recently merged PR #266 uses "COVIDcast Epidata API" and "Epidata API (other epidemics)"

capnrefsmmat: Roni is, I think, a fan of branding it the "COVIDcast Epidata API", as done in #266. "COVIDcast" is a pretty good brand name, so I'd be happy to centralize on that language.

I don't think we need to mention client package citations on each data source page -- we can put that in their documentation.

docs/api/covidcast-signals/_source-template.md
Akvannortwick: Everything seems pretty straight forward to me. Adding the example helps clarify at least one way of citing the use of Covidcast data.

docs/api/covidcast-signals/_source-template.md
RoniRos: Actually, @capnrefsmmat, we may have misunderstood each other.  I strongly support unified branding of the EpiData database and API.  It has a bright future for multiple diseases.  The only way for the rest of the world to make sense of it is as a unified DB & API for all epidemiological time series, with COVID-related time series being one component.   
I agree that "COVIDcast" is a very good brand name for COVID properties (if I should say so myself!).  We can still use it as an endpoint, and as the name for COVID-related maps, apps, tools, etc.  But I think the database and API should be understood to be EpiData.  So I support David's initiative to fix all references, I think it's long overdue.

docs/api/covidcast-signals/_source-template.md
dfarrow0: It feels again like a divergence is happening, and I'd like to understand better.

In this case, the API has citation text in the top-level [readme](https://github.com/cmu-delphi/delphi-epidata#citing), which is linked to by the COVIDcast [docs](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html#covidcast-epidata-api). Can you clarify why a different citation would be desired here? Hypothetically, if someone uses a `covidcast` signal, would we want them to cite Epidata and COVIDcast separately?

More thoughts to come.

dfarrow0: Some stray thoughts around terminology and citations.

I'd like to see unified terminology around the api and covidcast, and I think that means replacing "COVIDcast API" with "Epidata API".

COVIDcast encompasses many efforts, such as: the interactive website, acquisition/curation/generation of various signals, the `covidcast`-specific client libraries, the blog (some posts at least), etc. But I wouldn't consider the API to be a COVIDcast project.

However, I think it's important to recognize the huge contributions made by a large number of people to the various COVIDcast efforts. So I don't think it's optimal to cite the Epidata API as a proxy for the separate body of COVIDcast work. Where the two projects touch — the `covidcast` endpoint of the Epidata API — I think maybe a dual citation is not a terrible thing?

So for example, if there's a signal that Delphi generates (i.e. we're not just mirroring someone else) — like the indicator combination — then maybe we ought to have users cite the Epidata API for hosting that signal, and also cite the "COVIDcast Project" to satisfy the signal's CCBY license?

Which leads to another point, which is that we should have a COVIDcast Project page, with a list of contributors and a copy-pasta citation. I _think_ this is in the works as part of the site redesign.

capnrefsmmat: We do have [a public list of contributors](https://covidcast.cmu.edu/covid19-response-team.html). It was initially the "Delphi COVID-19 Response Team", but a few months ago it was decided that everyone is just part of Delphi now, and there is no separate COVID-19 team. I believe the plan is for this to be merged into a general list of Delphi members on the new site (but I might be wrong).

The practical concerns are that

* we already named the API clients "covidcast"
* we refer to the "COVIDcast API" in various presentations and published posts/docs, and changing the name will cause confusion
* people can download data without using the API at all, via the CSV download feature on the COVIDcast site. [Technically the API is still involved, but the user doesn't see it; they just know they're on COVIDcast]
* people have a hard enough time consistently giving credit with one source to cite; having two citations invites confusion

I'm not sure how to balance these and pick one option. But we clearly need a solution soonish so people can cite our work.

RoniRos: * @capnrefsmmat : I now understand better what you meant by me being a fan of "COVIDcast EpiData API" -- you mean vs. other forms of tying the two brand names.  I had thought you meant vs. using Epidata.  My bad.
* "Delphi EpiData" is a known brand name among long-term participants in the CDC forecasting challenges, as well as CDC itself.
* I can live with either "COVIDcast EpiData API" or "COVIDcast endpoint of the EpiData API", or even using both with the former being an informal shorthand for the latter.  What I prefer to avoid is "COVIDcast API", because at some point soon will will go back to revamping and expanding the other endpoints of Epidata, and it's important for users to understand that conceptually it's all one thing.  If we retroactively replace all "COVIDcast API" with "COVIDcast EpiData API" in our blogs and presentations, I don't think anyone will be confused.
* EpiData is not just an API, it is also the database itself, including lots of flu-related signals, going back 20 years, dengue signals, norovirus  signals, etc.  At some point we may want to distinguish the API from the DB (for example if there are multiple DBs managed by different organizations, all speaking EpiData), but I guess we don't need to worry about it now.

capnrefsmmat:
To push this forward, let me make a concrete proposal for citation text. I don't expect this to be the best proposal, just something specific we can work from. @dfarrow0 and @RoniRos, I welcome suggestions for changes to this text.

For signals constructed by Delphi:

> This data is provided under the terms of the [Creative Commons Attribution license](https://creativecommons.org/licenses/by/4.0/). If you use this data, please cite us with the following text and link:
>
>> Data from COVIDcast, part of the Delphi Epidata API. <https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/SOURCE-NAME.html

For signals republished from other sources, we'd say:

> This data is originally published by SOURCE NAME under the LICENSE. If you obtain this data from Delphi, such as via our API clients or the COVIDcast website, please cite both the original source and Delphi:
>
>> SOURCE'S PREFERRED ATTRIBUTION (e.g. data from JHU, link to their GitHub)
>> Obtained via COVIDcast, part of the Delphi Epidata API. <https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/SOURCE-NAME.html

Open questions:

* I've mentioned both COVIDcast and Epidata, positioning COVIDcast as "part of the Delphi Epidata API", so we both include the COVIDcast name (known to the general public) and the Epidata name (as the "real" source that's known to the epi-forecasting community). Do we want to do it this way?
* Is it "EpiData" or "Epidata"? All the existing documentation says "Epidata", but I see "EpiData" on delphi.cmu.edu, and Roni seems to use it as well.
* Do we want the attribution link to be to the technical docs? I say yes, because it means anyone seeing the data "in the wild" can immediately find out what it is. But we could also point to delphi.cmu.edu (or include both links).

ryantibs:
Chiming in here on just the most trivial part.  Ignoring the rest because you didn't ask my opinion 😄 

I prefer "Epidata", I generally don't like camel caps, looks like we're selling something on an informercial.

dfarrow0:
Thanks Alex, that looks great to me!

Response to questions:
- sounds ok to me
- hah, naming is hard. I think it started as "EpiData" and evolved to "Epidata". My slight preference is "Epidata", but either is fine as long as we're consistent.
- I think linking to the docs is most helpful for people who come across the citation. But it puts some pressure on us to not break the link. I'm not sure how to handle that for signals that we deprecate or rename, etc. And we won't be able to reorganize the docs without setting up a bunch of redirects. Having said that, I slightly prefer linking to the doc as you suggested since that is best for end-users.

---

Some thoughts on COVIDcast, Epidata, APIs, and citations. Maybe this isn't the best forum for this, but we're here now and it's convenient. Please feel free to disagree or suggest alternatives. I mainly just want to make sure we're all aligned in our view of these things, otherwise it really complicates communication.

- The Epidata API is a service and set of client libraries which together provide a variety of epi-related data across many times, locations, and diseases. A lot of data acquisition code lives in the same repo, but technically I wouldn't call that part of the API.
- The COVIDcast moniker encompasses Delphi's various responses to the COVID-19 pandemic. The "flagship" product is arguably the eponymous map. But COVIDcast also includes the efforts which produce the data ("indicators") that are surfaced through the `covidcast` endpoint of the Epidata API — which is shown on the map. Another part of the COVIDcast effort is producing client libraries which (1) fetch COVIDcast indicators from the Epidata API and (2) perform analysis/plotting on those data streams. (Open question: do the various nowcasting/forecasting efforts fall under the COVIDcast umbrella? I think probably not, but I could be convinced otherwise.)
- COVIDcast doesn't have an API. There's the Epidata API, of which `covidcast` is a component, and there's the COVIDcast clients, which call the Epidata API, but aren't themselves an API. (In the most technical sense, the call surface of the client library is _an_ API, but it's not what would be meant by a term like "COVIDcast API".)
- COVIDcast, as in Delphi's collective response to COVID-19, isn't a part of (read: "subset of") the Epidata API. While the Epidata API has the `covidcast` endpoint specific to the COVIDcast effort, the Epidata API doesn't encompass COVIDcast.
- We gently ask users to cite the Delphi Epidata API when they use the Epidata API as a service, regardless of whether they use one of our client libraries, and regardless of what data source they request (e.g. they might just `curl` some public domain government data). Users aren't _required_ to cite the Epidata API because it's a service, and users aren't currently bound by any terms of service. Open question: can terms of service require citation? If so, is that even desirable? Personal opinion: In the interest of lowering the barrier to entry as far as possible, I'm fine with the current gentle request rather than imposing a terms of service. (But that's just a dream, and we'll eventually need a ToS if we get popular enough.)
- We require users to cite Delphi when they use Delphi data. I think this is fair and reasonable. This is done by the CC BY-SA license which we apply to the data. As you've written above: "Data from COVIDcast, part of the Delphi Epidata API.". Given the lengthy novel I've written prior to this sentence, something like "Data from Delphi COVIDcast. Obtained via the Delphi Epidata API." is more in line with my thinking.

capnrefsmmat:
I also prefer "Epidata", unless we name it "Delphi EpiData COVIDcast ProData TurboStats Plus®™", which as you know [I'd totally approve of](https://delphi.cmu.edu/csvomatic).

I'd be happy with the language "Data from Delphi COVIDcast. Obtained via the Delphi Epidata API" rather than my suggestion above. In fact, here are three options -- pick your favorite:

1. Data from COVIDcast, part of the Delphi Epidata API. 
2. Data from Delphi COVIDcast. Obtained via the Delphi Epidata API
3. Data from Delphi's COVIDcast, via the Epidata API. [just removing some redundancy]

Regarding the other points above, I agree, apart from a couple notes:

* There's an ongoing effort to make the [COVIDcast website terms of use](https://covidcast.cmu.edu/terms-of-use.html) apply to the `covidcast` API endpoint as well, and there's currently draft language that we're supposed to approve and post. This is mainly to adjust the licensing information in the TOU to account for #286.
* We can retire citation for data insofar as it's covered by CC-BY, which excludes public domain government data, as you mention. I suppose terms of use *could* require citation anyway, but that would be considered anti-social behavior by the kind of people who advocate for open scientific data. Citation is a widely accepted social norm, enforced by social pressure rather than law, and open access to data shouldn't be impeded by needing to sign forms or read legal documents.

dfarrow0:
I'm on-board with all three. Ever-so-slight preference is in order: 3, 2, 1. Maybe @RoniRos has a stronger preference.

Re last point, I couldn't agree more.
> open access to data shouldn't be impeded by needing to sign forms or read legal documents

RoniRos:
The Epi[Dd]ata API is David’s brainchild, so I would tend to yield to his preference.  But I’d like to offer some context:

At least for me, the use of EpiData was an attempt to create a unique name that people can recognize as a proper noun rather than a generic term for epidemiological data (‘epi’ is a very common shorthand for epidemiology, and ‘epi data’ is a very common bigram in spoken and written exchanges).  A while back we even looked into making it a registered name, for that purpose, not because we are trying to sell a used car, but because we wanted it to become a recognized standard to emulate.  The registration has failed, predictably, because ‘epidata’ is such a generic combination that it is already in use by a variety of organizations around the world.

My fallback solution, and the one I’d like to suggest here, is to always preface Epi[dD]ata with a modifier, either ‘Delphi Epidata’ or ‘CMU Epidata’ or ‘Carnegie Mellon Epidata’.  Let’s choose one and stick with it.

This raises the question of how ‘COVIDcast’ fits in.  To re-iterate what I have said in other forums, in spite of COVIDcast being obviously very prominent this year, the idea of Epidata is much broader, started well before the current pandemic, and will hopefully become even more prominent much after the pandemic hopefully recedes.  So my preference would be to use “COVIDcast” as a modifier or endpoint of the more general concept, which is, e.g. “Delphi Epidata”.
Best,
Roni

For COVID-19 indicators in the U.S., see covidcast.cmu.edu<https://covidcast.cmu.edu/> by CMU’s Delphi group<https://delphi.cmu.edu/>.

RoniRos:
> I'm on-board with all three. Ever-so-slight preference is in order: 3, 2, 1. Maybe @RoniRos has a stronger preference.

I do have a preference to avoid #3, for the reason I mentioned above: 'Epidata' shouldn't be used anywhere with the 'Delphi' prefix.  This actually goes beyond the currently discussed terminology.

capnrefsmmat:
I'd be happy adopting option 2 as the default citation text. I can adapt the template later and update this pull request with new language.

capnrefsmmat:
Replaced by #299.

PR #292, opened 2020-11-18
new data source: covid hospitalization

### LINKED ISSUE #288: add healthdata.gov signals on hospitalizations
We are hoping to migrate the teams doing the hospitalization forecasting challenge at the COVID-19 Forecast Hub over to using the datasets recently (in the last few days) made publicly available at the link below.

Note API links on left-hand panel down the page a bit:
https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-state-timeseries

dfarrow0:
Adds a new data source to the API, start to finish.

"COVID-19 Reported Patient Impact and Hospital Capacity by State Timeseries" dataset provided by the US Department of
Health & Human Services via healthdata.gov.

https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-state-timeseries

Fixes #288

---

See individual commits for more details on what happens at each step.

- add test data and code readme
- implement network interface
- implement database interface, including ddl for new tables
- implement acquisition code
- add new endpoint `covid_hosp` to api server
- add same endpoint to all epidata clients
- write integration tests for acquisition and serving
- provide api documentation

---

repo-wide unit tests: ✔ All 52 tests passed!
repo-wide integration tests: ✔ All 29 tests passed!


integrations/acquisition/covid_hosp/test_scenarios.py
```diff
@@ -0,0 +1,87 @@
...
+      self.assertEqual(response['result'], 1)
+      self.assertEqual(len(response['epidata']), 1)
+      row = response['epidata'][0]
+      self.assertEqual(row['state'], 'MA')
+      self.assertEqual(row['date'], 20200510)
+      self.assertEqual(row['issue'], 20201116)
+      self.assertEqual(row['hospital_onset_covid'], 53)
+      actual = row['inpatient_bed_covid_utilization']
+      expected = 0.21056656682174496
+      self.assertTrue(abs(actual - expected) < 1e-5)
```
krivard, line 72:
```suggestion
      self.assertAlmostEqual(actual, expected, delta=1e-5)
```

integrations/server/test_covid_hosp.py
```diff
@@ -0,0 +1,78 @@
...
+      with db.new_cursor() as cur:
+        cur.execute('truncate table covid_hosp')
+        cur.execute('truncate table covid_hosp_meta')
+
+  def test_query_by_issue(self):
+    """Query with and without specifying an issue."""
+
+    # insert dummy data
+    def insert_issue(cur, issue, value):
+      so_many_nulls = ', '.join(['null'] * 51)
```
krivard, line 36: > so_many_nulls

lol

src/ddl/covid_hosp.sql
```diff
@@ -0,0 +1,408 @@
...
+
+NOTE: Names have been shortened to 64 characters, as this is a technical
+limitation of the database. Affected names are:
+
+staffed_icu_adult_patients_confirmed_and_suspected_covid ->
+  staffed_icu_adult_patients_confirmed_suspected_covid
+staffed_icu_adult_patients_confirmed_and_suspected_covid_coverage ->
+  staffed_icu_adult_patients_confirmed_suspected_covid_coverage
+staffed_icu_adult_patients_confirmed_covid ->
+  staffed_icu_adult_patients_confirmed_covid
```
krivard, line 136: these are identical

src/ddl/covid_hosp.sql
```diff
@@ -0,0 +1,408 @@
...
+staffed_icu_adult_patients_confirmed_covid_coverage ->
+  staffed_icu_adult_patients_confirmed_covid_coverage
```
krivard, line 138: these are identical

krivard (commented):
unfinished review; will hit the other files later

integrations/acquisition/covid_hosp/test_scenarios.py
dfarrow0: Oh wow, that's handy! I think I had another one of these makeshift comparisons around here somewhere. I'll update them all, and default precision is fine with me, nothing special about 1e-5.

src/ddl/covid_hosp.sql
dfarrow0: wow nice catch. i double-checked both lists again. these two weren't changed, and none of the changed ones are missing from this list. updated to remove these.

src/ddl/covid_hosp.sql
dfarrow0: fixed, per comment above

src/client/delphi_epidata.js
```diff
@@ -1,693 +1,621 @@
...
-    var BASE_URL, _list, _listitem, _request; // Because the API is stateless, the Epidata class only contains static methods
-
+  Epidata = (function() {
+    var BASE_URL, _list, _listitem, _request;
 
-    var Epidata = /*#__PURE__*/function () {
-      function Epidata() {
-        _classCallCheck(this, Epidata);
+    // Because the API is stateless, the Epidata class only contains static methods
+    class Epidata {
```
krivard, line 24: The Delphi primary doesn't speak ES6; you'll need to transpile to ES5. This will do it:

```
babel delphi_epidata.es6.js --presets=@babel/env > delphi_epidata.js
```

src/acquisition/covid_hosp/update.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+
+    match = Update.REVISION_PATTERN.match(revision)
+    if not match:
+      raise UpdateException(f'unable to extract issue from "{revision}"')
+    y, m, d = match.group(3), match.group(1), match.group(2)
+    return int(y) * 10000 + int(m) * 100 + int(d)
+
+  def get_date_as_int(date):
+    """Convert a YYYY-MM-DD date from a string to a YYYYMMDD int.
```
krivard, line 86: I notice this dataset uses multiple date formats (MM/DD/YYYY and YYYY-MM-DD). How stable do we think those formats are, and would it be wise to use something like `dateutil`?

krivard (changes requested):
Needs a new javascript client (otherwise it will fail deployment on prod) but everything else is probably optional

src/client/delphi_epidata.js
dfarrow0: I'm confused, can you elaborate? AFAIK the server just hands this file out like any other static web content. Was there an issue with ES6 at some point that I missed?

src/acquisition/covid_hosp/update.py
dfarrow0: `YYYY-MM-DD` should be stable as that comes from the dataset, which had better have a stable format.

`MM/DD/YYYY` comes from a field in the metadata which is not described. I treat it as a free-form text field, but in practice I would bet that the format won't change.

I'm not familiar with `dateutil`. It could be wise, but I recommend deferring that upgrade until later since it's sortof speculative at this point.

src/acquisition/covid_hosp/update.py
```diff
@@ -0,0 +1,176 @@
...
+  """Exception raised exclusively by the Update class."""
+
+
+class Update:
+
+  # regex to extract issue date from revision field
+  # example revision: "Mon, 11/16/2020 - 00:55"
+  REVISION_PATTERN = re.compile(r'^.*\s(\d+)/(\d+)/(\d+)\s.*$')
+
+  def get_entry(obj, *path):
```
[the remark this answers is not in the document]
dfarrow0: I'm not sure what you're referring to here.

(But yes, overkill is most likely an accurate description in any case :) )

src/acquisition/covid_hosp/database.py
```diff
@@ -0,0 +1,208 @@
...
+    issue : int
+      Issue of the dataset in YYYYMMDD format.
+    dataframe : pandas.DataFrame
+      The dataset.
+    """
+
+    # the database requires `nan` to be converted to `None` for `NULL` values
+    dataframe = dataframe.replace({math.nan: None})
+
+    value_placeholders = ', '.join(['%s'] * 56)
```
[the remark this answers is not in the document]
dfarrow0: I like it.

src/client/delphi_epidata.js
krivard: The server runs uglifyjs on this file. If it's in ES6, we get an error like this during deployment:

```
  [uglifyjs '/home/automation/driver/github_deploy_repo__tmp/src/client/delphi_epidata.js' -c -m -o '/home/automation/driver/github_deploy_repo__tmp/src/client/delphi_e$
idata.min.js']
Parse error at /home/automation/driver/github_deploy_repo__tmp/src/client/delphi_epidata.js:24,10
SyntaxError: Unexpected token: name (Epidata)
```

src/acquisition/covid_hosp/update.py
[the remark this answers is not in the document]
krivard: I meant writing a utility to dig through a deep tree, instead of just accessing it directly. It _does_ give us more sensical error messages, should they ever occur, and helps centralize the error handling implementation, but `get_entry` is only ever called six times, and that's including the tests. So... borderline, yeah. We'll leave it in.

src/client/delphi_epidata.js
dfarrow0: Ohhh that's right... /facepalm. Thanks for the reminder.

At risk of starting another side quest from my current side quest, I'd like to update deployment so that the minify action uses something more modern than uglify, like babel. But I'll def leave that alone in this PR. In the meantime, I'll just recompile this targeting ES5.

src/acquisition/covid_hosp/update.py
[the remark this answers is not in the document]
dfarrow0: Yeah so I normally wouldn't go _quite_ to this extent, but in a recent PR in another repo I had some feedback asking for more graceful handling when things aren't found — so I went extra safe-mode here as well.

src/client/delphi_epidata.js
dfarrow0: Created https://github.com/cmu-delphi/github-deploy-repo/issues/16 to track the ES6 issue.

dfarrow0:
@krivard thanks for the review, should be good to go!

krivard (approved):
🎉

PR #296, opened 2020-11-20
Fb survey fix changelog

Akvannortwick:


docs/api/covidcast_changelog.md
```diff
@@ -1,4 +1,4 @@
----
+        ---
```
capnrefsmmat, line 1: This change is probably unintentional and should be un-done, since I think it'll mess up the Markdown formatting

docs/api/covidcast_changelog.md
```diff
@@ -40,6 +40,14 @@ Duplicate survey weights had corrupted historical figures for the following sign
   * `hrr`: 20200406-20200415, 20200430-20200506
   * `msa`: 20200408-20200414, 20200430-20200506
   * `state`: 20200408-20200416, 20200430-20200506
+  
+#### 20 November 2020  
+
+Due to a bug in our data processing system, estimates of the percentage of people reporting that they have been tested for COVID-19 calculated before October 8th were incorrect. 
+We incorrectly treated an answer of “no” as a missing response, which affected the smoothed_tested_14d and smoothed_wtested_14d signals from the fb-survey source.
```
capnrefsmmat, line 47:
```suggestion
We incorrectly treated an answer of “no” as a missing response, which affected the `smoothed_tested_14d` and `smoothed_wtested_14d` signals from the `fb-survey` source. 
```

capnrefsmmat: We typically put source and signal names in monospace font, which the backticks do

capnrefsmmat (changes requested):
Mostly good, just a couple suggested minor fixes

docs/api/covidcast_changelog.md
Akvannortwick: Oops sorry about that, will do.

capnrefsmmat (approved):
Sorry for the delay

Issue #302, opened 2020-12-02
Better document meaning of SafeGraph Patterns signals


capnrefsmmat:
Currently the signal docs say things like:

* The number of daily visits to bar-related POIs in a certain region
* The number of daily visits to bar-related POIs in a certain region, per 100,000 population

I don't think this is accurate. SafeGraph does not count all visits to bars or restaurants, but rather visits by members of their "panel", i.e. people who have whatever apps do the tracking. They provide separate [normalization data](https://docs.safegraph.com/docs/weekly-patterns#section-normalization-stats) to indicate how many people are in their panel in each state.

The documentation should make clear (in a Limitations section) that the counts are subject to bias if some regions have more SafeGraph panel members, and do not represent absolute counts, just counts in the panel. Also, unless we already use the normalization data (do we?) we should consider using it to adjust the numbers correctly.

cc @Akvannortwick, @jingjtang

jingjtang:
@capnrefsmmat. We DO NOT use the normalization data currently. Actually it does not provide what we want. We want the number of unique visitors （or devices） per day per zip to be our denominators. However:

- normalization_stats.csv has the total number of visitors considered across US per day but no data for specific regions.
- home_summary_file has weekly number of devices per zip but no daily data

krivard:
Fixed in #304

PR #303, opened 2020-12-02
Add limitation description for geo coverage problem

jingjtang:
Bar visits data is missing for 9-12 states/regions. 
Add description in docs for the geo coverage problem.


capnrefsmmat:
It looks like something weird happened with the Git branching here; could you try to get it down to the one change without conflicts?

Also, it'd be good to address #302 while we're here.

jingjtang:
I will close this PR and make another one.

PR #309, opened 2020-12-04
Change requests from GET to POST

chinandrew:
See https://github.com/cmu-delphi/covidcast/issues/362. I've verified that the post works on the server, but that was just a single query and not what the best way to do more intensive stuff is.

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -79,6 +79,41 @@ class CovidcastTests(unittest.TestCase):
       'message': 'success',
     })
 
+  def test_uri_too_long(self):
+    """Make a simple round-trip with some sample data."""
+
+    # insert dummy data
+    self.cur.execute('''
+      insert into covidcast values
+        (0, 'src', 'sig', 'day', 'county', 20200414, '01234',
+          123, 1.5, 2.5, 3.5, 456, 4, 20200414, 0, 1, False)
+    ''')
+    self.cnx.commit()
+
+    # make the request with GET
+    response = requests.get(BASE_URL, {
+      'source': 'covidcast',
+      'data_source': 'src'*10000,
+      'signal': 'sig',
+      'time_type': 'day',
+      'geo_type': 'county',
+      'time_values': 20200414,
+      'geo_value': '01234',
+    })
+    self.assertEqual(response.status_code, 414)
+
+    # make request with POST
+    response = requests.post(BASE_URL, {
+      'source': 'covidcast',
+      'data_source': 'src'*10000,
+      'signal': 'sig',
+      'time_type': 'day',
+      'geo_type': 'county',
+      'time_values': 20200414,
+      'geo_value': '01234',
+    })
+
+    self.assertEqual(response.status_code, 200)
 
   def test_csv_format(self):
     """Test generate csv data."""
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -80,7 +80,7 @@ class CovidcastTests(unittest.TestCase):
     })
 
   def test_uri_too_long(self):
-    """Make a simple round-trip with some sample data."""
+    """Test that a long request yields a 414 with GET but works with POST."""
 
     # insert dummy data
     self.cur.execute('''
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -55,7 +55,7 @@ class Epidata:
     """Request and parse epidata."""
     try:
       # API call
-      return requests.get(Epidata.BASE_URL, params, headers=_HEADERS).json()
+      return requests.post(Epidata.BASE_URL, params, headers=_HEADERS).json()
     except Exception as e:
       # Something broke
       return {'result': 0, 'message': 'error: ' + str(e)}
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -52,7 +52,12 @@ class Epidata:
   # Helper function to request and parse epidata
   @staticmethod
   def _request(params):
-    """Request and parse epidata."""
+    """Request and parse epidata.
+
+    We default to GET since it has better caching and logging
+    capabilities, but fall back to POST if the request is too
+    long and returns a 414.
+    """
     try:
       # API call
       req = requests.get(Epidata.BASE_URL, params, headers=_HEADERS)
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -55,7 +55,10 @@ class Epidata:
     """Request and parse epidata."""
     try:
       # API call
-      return requests.post(Epidata.BASE_URL, params, headers=_HEADERS).json()
+      req = requests.get(Epidata.BASE_URL, params, headers=_HEADERS)
+      if req.status_code == 414:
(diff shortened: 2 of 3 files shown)


dfarrow0:
Nice! Did you by chance run the integration tests?

chinandrew:
I've run it on the `repos/delphi/delphi-epidata/tests/` folder and it passed, but are there tests written for the client? I only see this file: https://github.com/cmu-delphi/delphi-epidata/blob/main/tests/client/test_delphi_epidata.py

dfarrow0:
Gotcha. `tests/` contains all the unit tests, while `integrations/` contains the integration tests. Unlike the unit tests, the integration tests involve local database and api servers, and the tests there make real requests against those servers. The integration analogue of the unit test for the client is at https://github.com/cmu-delphi/delphi-epidata/blob/main/integrations/client/test_delphi_epidata.py, but it would be nice to verify that all the other integration tests pass too since many of them use the client in their testing.

Running integration tests requires a few extra steps, like building and running containers for the database and api. The process is documented in the [epidata dev guide](https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/epidata_development.md). Let me know if you have any questions with that, I'd be happy to help navigate.

chinandrew:
:+1: thanks for the link, just ran integration tests and they passed.

dfarrow0:
Awesome, thanks!

Just to see if there are any surprise limitations (e.g. query length) in the backend, how do you feel about adding a short test that:
1. requests all counties via GET, and get back 414 status code
2. requests all counties via POST, and get back some data

Probably fine to merge without the extra test, but if you have the cycles it make a great addition imo.

chinandrew:
Good idea, I'll look into it.

chinandrew:
Hmm looks like the post is also getting a 414?

Running on the docker container and using  `'src'*3000` to make a really long request
```
# long request
>>> requests.get("http://delphi_web_epidata:80/epidata/api.php", params={'source': 'covidcast','data_source': 'src'*3000,'signal': 'sig','time_type': 
'day','geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
414
>>> requests.post("http://delphi_web_epidata:80/epidata/api.php", params={'source': 'covidcast','data_source': 'src'*3000,'signal': 'sig','time_type':
 'day','geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
414

# normal request
>>> requests.get("http://delphi_web_epidata:80/epidata/api.php", params={'source': 'covidcast','data_source': 'src','signal': 'sig','time_type': 'day'
,'geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
200
>>> requests.post("http://delphi_web_epidata:80/epidata/api.php", params={'source': 'covidcast','data_source': 'src','signal': 'sig','time_type': 'day
','geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
200
```

dfarrow0:
Oh! I think you need to use `params` with `get` and `data` with `post`. `params` always feeds into the URL, even with `post`. [docs](https://requests.readthedocs.io/en/latest/api/#requests.post)

Does this work? (s/params/data/)
```
requests.post("http://delphi_web_epidata:80/epidata/api.php", data={'source': 'covidcast','data_source': 'src'*3000,'signal': 'sig','time_type':
 'day','geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
```

chinandrew:
> Oh! I think you need to use `params` with `get` and `data` with `post`. `params` always feeds into the URL, even with `post`. [docs](https://requests.readthedocs.io/en/latest/api/#requests.post)
> 
> Does this work? (s/params/data/)
> 
> ```
> requests.post("http://delphi_web_epidata:80/epidata/api.php", data={'source': 'covidcast','data_source': 'src'*3000,'signal': 'sig','time_type':
>  'day','geo_type': 'county','time_values': 20200414,'geo_value': '01234'}).status_code
> ```

Good catch, that worked. Thanks!

chinandrew:
Tests added. Two to just post/get the api, and one that uses the _request() method to ensure a post returns a valid response

krivard:
Should this PR update the Coffee and JS clients as well? Having them not-match is gonna make me itch

dfarrow0:
Thanks for the tests. I didn't see an update to the client though to use `data` instead of `params`. Did you mean to include that change too?

dfarrow0 (approved):

sgratzl:
regarding GET vs POST. One advantage of using GET when possible is that the response can be cached by the browser/proxies, POST requests are usually not cacheable

dfarrow0:
That's a good point. We could improve this (in the future) by attempting GET, and falling back to POST if status is 414. Or, we could try to inspect the request and switch to POST if it looks like a GET would likely fail.

chinandrew:
> That's a good point. We could improve this (in the future) by attempting GET, and falling back to POST if status is 414. Or, we could try to inspect the request and switch to POST if it looks like a GET would likely fail.

Is there a reason to do anything more complicated than 
```
req = requests.get(...)
if req.status_code == 414:
    req = requests.post(...)
```

or some rough approximation based on len(str(params)) as you suggested

dfarrow0:
looks good to me

sgratzl:
another advantage of using GET: we get better access logs

chinandrew:
Added conditional for post statement

integrations/server/test_covidcast.py
```diff
@@ -79,6 +79,41 @@ def test_round_trip(self):
       'message': 'success',
     })
 
+  def test_uri_too_long(self):
+    """Make a simple round-trip with some sample data."""
```
dfarrow0, line 83: small thing, but do you mind updating this description to say we're testing get/post behavior?

src/client/delphi_epidata.py
```diff
@@ -55,7 +55,10 @@ def _request(params):
     """Request and parse epidata."""
     try:
       # API call
-      return requests.get(Epidata.BASE_URL, params, headers=_HEADERS).json()
+      req = requests.get(Epidata.BASE_URL, params, headers=_HEADERS)
```
dfarrow0, line 58: maybe worth a comment saying that we prefer get for caching/logging, but fall back to post if get fails

dfarrow0 (approved):

integrations/server/test_covidcast.py
chinandrew: oops this was copy pasted from another test, I've updated the docstrings.

src/client/delphi_epidata.py
chinandrew: Added

krivard:
> Should this PR update the Coffee and JS clients as well? Having them not-match is gonna make me itch

@chinandrew

chinandrew:
> > Should this PR update the Coffee and JS clients as well? Having them not-match is gonna make me itch
> 
> @chinandrew

That's a good point, agree they should be consistent. It'll probably take me some time to learn the relevant JS, if there's someone fluent they could probably make the change 10x faster. Otherwise I can work on it (after the geo work? or before?).

There's also an R client provided that we should update, though could probably repurpose some of the code thats being written for the covidcast client (not sure why it doesnt use these clients).

dfarrow0:
just a recommendation: suggest merging this one now since @JedGrabman needs it, and let someone update the other clients in a follow-up

PR #307, opened 2020-12-03
Chc & dv api updates

Akvannortwick:
Addresses the discussion on the backfill and spatial bias for the Change Healthcare and Doctor's Visits indicators. The content is preliminary and can be revised as necessary. 

- [x] Review and add suggestions for edits @rumackaaron @huisaddison

Issue #298, opened 2020-11-21
Various suggested doc updates


ryantibs:
Hi all, cc @Akvannortwick @dfarrow0 @capnrefsmmat @krivard @RoniRos 

Been peeking at our [documentation site](https://cmu-delphi.github.io/delphi-epidata/) now, and I have various suggested updates.  Wanted to bring them here for discussion and if we converge on them, then perhaps @Akvannortwick can tackle them.  These are in "visitation order" as I navigate down the menu bar on the left.

1. On the ["Home" page](https://cmu-delphi.github.io/delphi-epidata/), the title should be "Delphi Epidata API".  And the first sentence should begin "Delphi's Epidata API provides ...".

2. On this same page, the documentation link is broken ("... for all his help with the Epidata API documentation.")

3. On the ["COVIDcast Epidata API" page](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html), we should remove "Delphi’s COVID-19 Surveillance Streams".  This is old language, AFAIK, and we scrapped the word "surveillance" back in April.  I count two uses of the word "surveillance" on this page.  Should use "Delphi's COVID-19 indicators", which has been our standard terminology for a while.

4. On this same page, we should change all examples that use the `raw_cli` signal to instead use the `smoothed_cli` signal.  The latter is much more widely available (especially significant in the example that fetches available data for all counties).  In general, would do a `grep` to find all usages of `raw_cli` and replace it with `smoothed_cli`, in our documentation site.

5. On the ["API Clients" subpage](https://cmu-delphi.github.io/delphi-epidata/api/covidcast_clients.html) under the COVIDcast Epidata API section, we shouldn't document the generic Epidata clients.  This is confusing, in my opinion.  I think we should move everything from the "Generic Epidata Clients" header and below to a new subpage under the "Epidata API (Other Epidemics)" section, and the link to this from the current "API Clients" subpage.

6. On the ["Epidata API (Other Epidemics)" page](https://cmu-delphi.github.io/delphi-epidata/api/README.html), I think it's confusing that this is titled "Delphi Epidata API".  Given that the "Home" page also should have the same title.  I would rename this section something like "Epidata API (General Epidemics)" and then title the page the same way.
https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html).

[...]

PR #312, opened 2020-12-07
Api document revisions

Akvannortwick:
I believe I tackled all the points made in #298, if there is something I missed I can easily address them.

[krivard edit: this incorporates changes from #307, so should merge after that]


krivard:
I think you collided with #311; lmk if you want a buddy for resolving conflicts

Akvannortwick:
Yes, that would definitely help @krivard

krivard (approved):
for future travelers: title of README is being changed to be distinct from the title of docs/index.md, which is where the actual home is located

PR #325, opened 2020-12-11
feat: rename api parameter source -> endpoint

### LINKED ISSUE #321: Change name of first parameter from 'source' to 'endpoint'
This has been confusing to outsiders (e.g. email from Nick Reich, but probably just about everyone) with the 'signal source' parameter.

The same 'thing' should be called by the same name...

@undefx told me he agrees and suggested maybe @krivard or @capnrefsmmat could do or else oversee it?

No real rush, but maybe a good intro task for someone?

sgratzl:
closes #321

### DIFF
--- a/src/server/api.php
+++ b/src/server/api.php
@@ -924,7 +924,7 @@ function get_dengue_nowcast($locations, $epiweeks) {
 }
 
 // queries the `covidcast` table.
-//   $source (required): name of upstream data source
+//   $endpoint (required): name of data endpoint; 'source' also accepted here for backwards-compatibility
 //   $signals (required): array of names for signals derived from upstream data
 //   $time_type (required): temporal resolution (e.g. day, week)
 //   $geo_type (required): spatial resolution (e.g. county, msa, state)
--- a/src/server/api.php
+++ b/src/server/api.php
@@ -924,7 +924,7 @@ function get_dengue_nowcast($locations, $epiweeks) {
 }
 
 // queries the `covidcast` table.
-//   $source (required): name of upstream data souce
+//   $source (required): name of upstream data source
 //   $signals (required): array of names for signals derived from upstream data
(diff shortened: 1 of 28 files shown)


src/server/api.php
```diff
@@ -924,7 +924,7 @@ function get_dengue_nowcast($locations, $epiweeks) {
 }
 
 // queries the `covidcast` table.
-//   $source (required): name of upstream data souce
+//   $source (required): name of upstream data source
```
krivard, line 927:
```suggestion
//   $endpoint (required): name of data endpoint; 'source' also accepted here for backwards-compatibility
```

krivard (approved):
minor docs nit but otherwise fabulous, all unit and integration tests pass

krivard:
Ready to merge, but waiting until Monday just in case to appease the launch curse gremlins

sgratzl:
> Ready to merge, but waiting until Monday just in case to appease the launch curse gremlins

regarding your small change as far as I can is about the `function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $geo_values, $as_of, $issues, $lag) {` function which remains to have the `$source` parameter, since it is actually the `data_source` API parameter.

krivard:
augh you're right, sorry about that. fixing

src/server/api.php
```diff
@@ -924,7 +924,7 @@ function get_dengue_nowcast($locations, $epiweeks) {
 }
 
 // queries the `covidcast` table.
-//   $source (required): name of upstream data souce
+//   $endpoint (required): name of data endpoint; 'source' also accepted here for backwards-compatibility
```
krivard, line 927:
```suggestion
//   $source (required): name of upstream data source
```

Reverting 40f63f9; this was an unrelated typo fix

PR #328, opened 2020-12-14
R/JS api clients to support POST requests

### LINKED ISSUE #317: Switch Coffee, JS, and R Epidata clients to use POST instead of GET for long queries
We expedited https://github.com/cmu-delphi/delphi-epidata/pull/309, but now the Python Epidata client (yet unreleased) doesn't match the Coffee, JS, and R client behavior.

Use the Python implementation of POST as a template to bring the others back in line.

sgratzl:
closes #317 

are there any automatic test setups which will test that?


dfarrow0:
> are there any automatic test setups which will test that?

if you're meaning integration tests for js/r clients, then sadly no (at least that i'm aware of)

dfarrow0:
thanks for taking this on! i noticed that you added js/html files to test the js client. how do you run this locally? i'd be happy to add instruction to the dev guide and possibly automate this.

sgratzl:
atm they are simple smoke tests. You open the HTML file in the browser and it should show two success messages. Same for the node script that you just run. We could think of using a proper testing framework like jest or simple based on `console.assert`

src/client/delphi_epidata.coffee
```diff
@@ -42,9 +42,19 @@ class Epidata
       else
         callback(0, 'unknown error', null)
     # Request data from the server
-    if $?.getJSON?
+    isBrowser = $?.ajax?
+    data = if isBrowser then $.param(params) else querystring.stringify(params)
+    fullURL = "#{BASE_URL}?#{data}"
+    method = if fullURL.length < 2048 then 'GET' else 'POST'
```
krivard, line 48: add a comment here about avoiding 414

integrations/client/test_delphi_epidata.html
```diff
@@ -0,0 +1,9 @@
+<script src="https://unpkg.com/jquery"></script>
+<script src="../../src/client/delphi_epidata.js"></script>
+<script>
+    function cb(r, msg, data) {
+        console.log(r, msg, data);
+    }
+    Epidata.covidcast(cb, 'fb-survey', Array(1).fill('raw_cli').join(','), 'day', 'state', '2020-12-01', 'ca');
+    Epidata.covidcast(cb, 'fb-survey', Array(1000).fill('raw_cli').join(','), 'day', 'state', '2020-12-01', 'ca');
+</script>
```
krivard, line 9: add newline at EOF

integrations/client/test_delphi_epidata.js
```diff
@@ -0,0 +1,7 @@
+const Epidata = require('../../src/client/delphi_epidata').Epidata;
+
+function cb(r, msg, data) {
+    console.log(r, msg, data);
+}
+Epidata.covidcast(cb, 'fb-survey', Array(1).fill('raw_cli').join(','), 'day', 'state', '2020-12-01', 'ca');
+Epidata.covidcast(cb, 'fb-survey', Array(1000).fill('raw_cli').join(','), 'day', 'state', '2020-12-01', 'ca');
```
krivard, line 7: add newline at EOF

krivard (approved):
nice work! 

this lays the groundwork for adding a test runner; something that will output pass or fail with a docker command. out of scope for this PR but I'll make an issue for it.

krivard (changes requested):
[wrong radio button]

src/client/delphi_epidata.coffee
sgratzl: done

integrations/client/test_delphi_epidata.html
sgratzl: done

integrations/client/test_delphi_epidata.js
sgratzl: done

krivard (approved):

PR #329, opened 2020-12-14
covid hosp: finish refactoring

dfarrow0:
- done in preparation of adding a new dataset to the "covid hosp" family
- rename table `covid_hosp` to `covid_hosp_state_timeseries`
- add column `dataset_name` to `covid_hosp_meta`
- make api endpoint `covid_hosp` a synonym for 
`covid_hosp_state_timeseries`
- this is a no-op change, next PR will add the new `facility` scraper

all unit and integration tests pass. will need to update database according to new DDL when this is merged; also need to update automation step to match changes. otherwise, this is a functional no-op.

adding the `facility` scraper will require no changes beyond this point -- only additions.


dfarrow0:
noting that a bunch of files/functions were moved around, so the actual delta of this PR is significantly smaller than what's shown on github

dfarrow0:
https://github.com/cmu-delphi/delphi-epidata/pull/330 follows this with the implementation of the `facility` scraper

tests/acquisition/covid_hosp/state_timeseries/test_network.py
```diff
@@ -2,7 +2,7 @@
 
 # standard library
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import patch
```
krivard, line 5: ❤️

krivard (approved):
lgtm

PR #330, opened 2020-12-14
covid hosp: facility scraper

dfarrow0:
- adds the new `facility` scraper
- makes extensive reuse of the common `covid hosp` code; most additions 
are trivial
- adds new database table `covid_hosp_facility` which needs to be 
created prior to automation
- dataset is **not** surfaced by api yet, will happen in a later PR
- can add step to automation prior to surfacing through api, and after 
creating table mentioned above

all unit and integration tests pass


krivard (approved):
lgtm but in theory this PR should contain exactly one commit. github is also showing commits that were merged as part of #329. the commit hashes match, so in theory we should be okay?

PR #331, opened 2020-12-14
feat: support language parameter in wiki source

### LINKED ISSUE #50: R, js, and coffee clients missing support for wiki source's required language parameter
**problem:** the api for the `wiki` source now requires the `language` parameter:

https://github.com/cmu-delphi/delphi-epidata/blob/c48af8f126e935306eaebb79fcf1b22a201a3dfa/src/server/api.php#L1204-L1205

however, while the python client supports it, none of the other 3 clients (R, js, coffee) do. for example:

https://github.com/cmu-delphi/delphi-epidata/blob/c48af8f126e935306eaebb79fcf1b22a201a3dfa/src/client/delphi_epidata.R#L240

**impact:** customers won't be able to use these three clients for accessing the `wiki` source.

**proposal:** add the `language` parameter as a required parameter for the `wiki` source in these three clients.

sgratzl:
closes #50

### DIFF
--- a/src/client/delphi_epidata.R
+++ b/src/client/delphi_epidata.R
@@ -247,7 +247,7 @@ Epidata <- (function() {
   }
 
   # Fetch Wikipedia access data
-  wiki <- function(articles, dates, epiweeks, hours) {
+  wiki <- function(articles, dates, epiweeks, hours, language='en') {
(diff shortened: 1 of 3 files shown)


krivard (approved):
lgtm

PR #334, opened 2020-12-15
Clarify and expand NCHS documentation

capnrefsmmat:
I wanted to explain how it's different from the other death data, so I stated that explicitly, and fixed a few formatting bugs.


Akvannortwick (commented):
LGTM

PR #335, opened 2020-12-15
Document available time types and their formats more explicitly

capnrefsmmat:
List the formats and how to interpret them. List available time types for every signal so users know what to request (since they *must* specify a time type to fetch a signal).

### DIFF
--- a/docs/api/covidcast_times.md
+++ b/docs/api/covidcast_times.md
@@ -20,7 +20,7 @@ For example, consider using our [doctor visits signal](covidcast-signals/doctor-
 which estimates the percentage of outpatient doctor visits that are
 COVID-related, and consider a result row with `time_value = "2020-05-01"` for
 `geo_value = "pa"`. This is an estimate for the percentage in Pennsylvania on
-May 1, 2020. That estimate was issued on May 5, 2020, the delay being due to the
+May 1, 2020, which was issued on May 5, 2020. The delay is due to the
 aggregation of data by our source and the time taken by the API to ingest the
 data provided. Later, the estimate for May 1st could be updated, perhaps because
 additional visit data from May 1st arrived at our source and was reported to us.
--- a/docs/api/covidcast_times.md
+++ b/docs/api/covidcast_times.md
@@ -1,0 +1,44 @@
+---
+title: Date Coding and Revisions
+parent: COVIDcast Epidata API
+nav_order: 5
+---
+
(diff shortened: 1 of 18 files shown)


docs/api/covidcast_times.md
```diff
@@ -0,0 +1,47 @@
...
+nav_order: 5
+---
+
+# Date Coding and Revisions
+
+Every observation in the COVIDcast Epidata API has two dates attached:
+
+* `time_value`: The time the underlying events happened. For example, for a data
+  source that reports on COVID test results, the time value is the date the
+  results were recorded by the testing provider.
```
Akvannortwick, line 13:
```suggestion
* `time_value`: The time the underlying events happened. For example, when a data
  source reports on COVID test results, the time value is the date the
  results were recorded by the testing provider.
```

docs/api/covidcast_times.md
```diff
@@ -0,0 +1,47 @@
...
+* `issue`: The date the estimates were *issued*. For example, a COVID test
+  result might be recorded on October 1st, but it may take several days for
+  that report to be collected, aggregated, received by Delphi, and added to our
+  database. The date Delphi makes the data available is the *issue date*.
```
Akvannortwick, line 17:
```suggestion
  database.  The *issue date* is when Delphi makes the data available.
```

docs/api/covidcast_times.md
```diff
@@ -0,0 +1,47 @@
...
+
+For example, consider using our [doctor visits signal](covidcast-signals/doctor-visits.md),
+which estimates the percentage of outpatient doctor visits that are
+COVID-related, and consider a result row with `time_value = "2020-05-01"` for
+`geo_value = "pa"`. This is an estimate for the percentage in Pennsylvania on
+May 1, 2020. That estimate was issued on May 5, 2020, the delay being due to the
```
Akvannortwick, line 23:
```suggestion
May 1, 2020. That estimate was issued on May 5, 2020, the delay is due to the
```

Akvannortwick (commented):
Some minor suggestions otherwise it looks good.

docs/api/covidcast_times.md
capnrefsmmat: This change creates a comma splice; we'd either have to make it two sentences or leave it as-is

krivard: how about

```suggestion
May 1, 2020, which was issued on May 5, 2020. The delay is due to the
```

PR #339, opened 2020-12-16
Document monthly rollup CSVs

capnrefsmmat:
To be merged once the monthly rollup generator is implemented and monthly files are on the SFTP server.

(Yes, the survey loading functions are copypasta-d here; not sure how to do a better job of code reuse, save having people install all of `delphiFacebook` or making a new survey loading package)


docs/symptom-survey/survey-files.md
```diff
@@ -26,49 +26,75 @@ where the data is hosted.
...
+daily files, by taking the data with the most recent `recordedby` date for each
+day of the month. They are compressed with gzip; the standard `gunzip` command
+on macOS or Linux can decompress them. (macOS can also decompress these files
+through Finder automatically; on Windows, free programs like
+[7-zip](https://www.7-zip.org/) can decompress gzip files.) Users doing
+historical analyses of the survey data should start with these files, since they
+provide the easiest way to get all the necessary data, without accidentally
+including duplicate results.
+
+Second, we produce monthly tarballs containing all the daily `.csv.gz` files for
```
krivard, line 78: (monthly tarballs are just the most recent recordedby file for each date)

capnrefsmmat: fixed

docs/symptom-survey/survey-files.md
```diff
@@ -26,49 +26,77 @@ where the data is hosted.
...
+standard `tar` command. The unpacked files are described in [Daily
+Files](#daily-files) above.
+
 ## Conditions Responses are Recorded
 
-The survey was configured to record responses under two sets of circumstances:
+The survey is configured to record responses under two sets of circumstances:
 
-1. The user taking the survey clicked submit, or
+1. The user taking the survey reached the end of the survey, or
```
krivard, line 92: Not accurate. The "Progress" column records how much of the survey has been seen by the user; the "Finished" column records whether the user clicked submit. Responses can have Progress=100 (reached the end of the survey) and Finished=0 (did not click submit). Recorded dates for such responses are 4 hours after the last activity.

capnrefsmmat: Well, now I'm confused, because there is no submit button. You just proceed through successive pages until you click the arrow button and are thanked for participating. So... what does the Finished column actually measure?

capnrefsmmat: [Ah](https://www.qualtrics.com/support/survey-platform/data-and-analysis-module/data/download-data/understanding-your-dataset/):

> The Finished column details whether the response was submitted or closed. A “1” or “TRUE” indicates the respondent reached an end point in their survey (hitting the last Next/Submit button, being screened-out with Skip or Branch Logic, etc.). A “0” or “FALSE” indicates the respondent left their survey before reaching an end point and the response was instead closed manually or due to session expiration.
>
> The Progress column shows the progress a respondent made in the survey before finishing. For those marked as “1” or “TRUE” in the Finished column, the Progress is marked 100, regardless of whether they were screened out. For those whose responses are marked “0” or “FALSE,” you will get an exact percentage of how far they got in the survey based on what question they left off on.

I don't know how you get Progress=100 with Finished=0, though, since you'd have to answer all questions but  still have it closed after 4 hours.

capnrefsmmat: I'm going to interpret "Reached the end of the survey" here to mean "made it to the thank-you page", which is all they can do. Since it's not clear what Progress=100 with Finished=0 means, and it's an edge case, I don't see how else to change this text

krivard: I would bet Progress=100, Finished=0 means they reached the final page of questions, but did not hit the final next button. Since the end screen is configured from survey settings and not as actual question text in the main editor, it would make sense that the end screen is not considered part of Progress.

PR #340, opened 2020-12-16
micro-delta to relieve main pain point of metacache update performance.

melange396:
this is not the 'right way' to fix it, but it should be able to improve the timing of the computation in the meanwhile
(all tests pass)

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -266,9 +266,17 @@ class Database:
 
     meta = []
 
-    sql = 'SELECT `source`, `signal` FROM covidcast WHERE NOT `is_wip` GROUP BY `source`, `signal` ORDER BY `source` ASC, `signal` ASC;'
+    signal_list = []
+    sql = 'SELECT `source`, `signal` FROM `covidcast` GROUP BY `source`, `signal` ORDER BY `source` ASC, `signal` ASC;'
     self._cursor.execute(sql)
-    for source, signal in  [ss for ss in self._cursor]: #NOTE: this obfuscation protects the integrity of the cursor; using the cursor as a generator will cause contention w/ subsequent queries
+    for source, signal in list(self._cursor):
+      sql = "SELECT `is_wip` FROM `covidcast` WHERE `source`=%s AND `signal`=%s LIMIT 1"
+      self._cursor.execute(sql, (source, signal))
+      is_wip = int(self._cursor.fetchone()[0]) # casting to int as it comes out as bytearray (essentially str)
+      if not is_wip:
+        signal_list.append((source, signal))
+
+    for source, signal in signal_list:
 
       sql = '''
       SELECT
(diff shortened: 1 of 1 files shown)


src/acquisition/covidcast/database.py
```diff
@@ -266,9 +266,17 @@ def get_covidcast_meta(self):
...
 
-    sql = 'SELECT `source`, `signal` FROM covidcast WHERE NOT `is_wip` GROUP BY `source`, `signal` ORDER BY `source` ASC, `signal` ASC;'
+    signal_list = []
+    sql = 'SELECT `source`, `signal` FROM `covidcast` GROUP BY `source`, `signal` ORDER BY `source` ASC, `signal` ASC;'
     self._cursor.execute(sql)
-    for source, signal in  [ss for ss in self._cursor]: #NOTE: this obfuscation protects the integrity of the cursor; using the cursor as a generator will cause contention w/ subsequent queries
+    for source, signal in list(self._cursor):
+      sql = "SELECT `is_wip` FROM `covidcast` WHERE `source`=%s AND `signal`=%s LIMIT 1"
+      self._cursor.execute(sql, (source, signal))
+      is_wip = int(self._cursor.fetchone()[0]) # casting to int as it comes out as bytearray (essentially str)
```
benjaminysmith, line 275: Does it return a bytearray 1/0? Why not cast to a boolean instead?

benjaminysmith (approved):
LGTM -- iiuc you are basically implementing a hacky any_value aggregation function. That's sneaky. I think there is a way to define arbitrary aggregation functions, but this is simpler.

src/acquisition/covidcast/database.py
melange396: it does indeed return '0'/'1'.  only the empty string becomes false when casting str to bool :(

melange396:
i actually tried very hard to make/find an "any" aggregator, but could not make it happen. ¯\_(ツ)_/¯

dfarrow0:
oh wow, this is very surprising to me. i would have expected that filtering in python would be much slower than filtering in sql, especially with the new nested queries.

i'd like to test some ideas. how are you benchmarking for comparison?

benjaminysmith:
@dfarrow0 -- this is still filtering in SQL though, right? It's just moving the is_wip check out of the top query (where we check every row) to an inner query with a limit 1 (so we only check one row). Mariadb doesn't have an any_value aggregation function which would do the whole thing in one line.

melange396:
this might become untenable if the cardinality of the (source, signal) pairs gets too high, but for now it is quite speedy.

one more caveat is that we will have unpredictable behavior if any (s,s) pair has both True and False values for is_wip...  that "shouldn't" happen and it would probably indicate deeper problems.

dfarrow0:
@benjaminysmith yes technically filtering is done in sql, but the part that sticks out to me is that there are an additional `N` (number of source-signals pairs) new queries done from python. it would be preferable imo if the whole thing was just a single sql query.

i recommend leaving the original query as-is and adding an index to the table on (`is_wip`, `source`, `signal`). that should solve the full table scan and wouldn't require any code changes.

(i'm not meaning to downplay the value of this pr in any way -- good work -- i just wanted to point out that the root of the problem is a table scan, and that can be solved with an index.)

krivard:
I'm hesitant to add indexes to a table with 100s of millions of rows, especially since we already find ourselves provisioning more disk space every 6 weeks

PR #342, opened 2020-12-17
add endpoint `covid_hosp_facility`

dfarrow0:
- api udpated
- all clients updated
- added integration test which exercises the new endpoint in server and 
python client

next pr will add another endpoint for hospital discovery. (e.g. answer "what hospitals are in PA?". then use that info to query `covid_hosp_facility` with a specific `hospital_pk` of interest.)

all unit and integrations pass

### DIFF
--- a/integrations/acquisition/covid_hosp/facility/test_scenarios.py
+++ b/integrations/acquisition/covid_hosp/facility/test_scenarios.py
@@ -1,0 +1,87 @@
+"""Integration tests for acquisition of COVID hospitalization facility."""
+
+# standard library
+import unittest
+from unittest.mock import MagicMock
+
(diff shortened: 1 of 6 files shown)


krivard (approved):
really wish we didn't have to duplicate 90+ column names twice, but I don't see any way around it

krivard:
@dfarrow0  Does this need any prep in the DB before merging?

dfarrow0:
yes, the duplication hurts me too :(

no prep necessary, should be safe to merge -- and endpoint should go live once auto-deployed

we'll need to update pypi at some point, but i still have a new endpoint on the way, so recommend waiting for that

PR #343, opened 2020-12-17
facility fix and new endpoint

dfarrow0:
- `bool('false')` is `True`
- add a helper function to parse bool from str, allowing missing and 
empty values

note that prod database needs to be backfilled since previous 
acquisitions incorrectly casted 'false' and 'true' both to `True`

---

- the hhs covid hosp dataset is keyed by `hospital_pk`, which is a 
unique identifier for healthcare facilities.
- the key is not generally well-known, but querying the 
`covid_hosp_facility` dataset requires key(s)
- the new endpoint returns facility information (including key) for 
given regions (e.g. city, state, zip, ccn, fips)
- users would be expected to lookup hospitals of interest e.g. by state 
or city. then query `covid_hosp_facility` with the key for those 
hospitals.
- server and clients updated
- all unit and integrations pass

note that new indexes are added to the database to facilitate fast 
lookups. otherwise it would be a full table scan each time, which does 
not scale well. indexes are expected to be very small.

### DIFF
--- a/integrations/acquisition/covid_hosp/facility/test_scenarios.py
+++ b/integrations/acquisition/covid_hosp/facility/test_scenarios.py
@@ -85,3 +85,79 @@ class AcquisitionTests(unittest.TestCase):
           '450822', Epidata.range(20200101, 20210101))
       self.assertEqual(response['result'], 1)
       self.assertEqual(len(response['epidata']), 1)
+
+  def test_facility_lookup(self):
+    """Lookup facilities using various filters."""
+
+    # only mock out network calls to external hosts
+    mock_network = MagicMock()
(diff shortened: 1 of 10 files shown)


chinandrew (approved):
:shipit::shipit::shipit:

chinandrew:
does a list of states work here or do they have to be individual?

dfarrow0:
here they would be individual states. i'm not opposed to making that a list of states if it's useful, but prefer to defer that as a followup

chinandrew:
> here they would be individual states. i'm not opposed to making that a list of states if it's useful, but prefer to defer that as a followup

makes sense, sounds fine to leave as is

PR #337, opened 2020-12-15
streaming support

sgratzl:
closes #51 

streams results back to the user instead of caching it locally and then send it in a bulk.

It is a bigger change since I introduced the notion of a printer which is responsible for printing in different formats. There is no change for clients. 

Following formats are supported:
 * classic = classic epidata format
 * tree = the epidata format which the first element grouped by `signal` ... no streaming
 * json = flat json array
 * csv = csv file
 * jsonl = jsonlines format (https://jsonlines.org/) which is a nice format for streaming

don't know why
```
delphi.delphi-epidata.integrations.acquisition.covidcast.test_csv_uploading.CsvUploadingTests.test_uploading: fail
```

is failing. I didn't touch those files

PR #350, opened 2020-12-30
Proof of Concept: Flask based API

### LINKED ISSUE #177: consider using sqlalchemy
https://www.sqlalchemy.org/ provides a nice level on top of any python DB connection and there are several libraries on top of it.

### LINKED ISSUE #178: consider switching from PHP to Python
e.g. using https://fastapi.tiangolo.com/ which is a nice layer, ready for production and good integrations as well as testing capabilities.

sgratzl:
based on #337 

closes #178 and closes #177 

proof of concept that we could also use Python for the API and thus get rid of PHP for hosting the API. The public interface remains the same but allows way more flexibility if we want to. Moreover, it is python thus we can unit test/lint/format it properly. 

There is also a Dockerfile that will create a docker image to run it. 

for the 414 failing unit test, see also https://github.com/benoitc/gunicorn/issues/2487


chinandrew:
Took a high level look through this (did not look into all the lines in much detail given the size of this PR), but looks reasonable to me at first glance. Have you run it through the docker integration tests?

sgratzl:
> Have you run it through the docker integration tests?

see https://github.com/sgratzl/delphi-epidata/runs/1645645816?check_suite_focus=true as noted in the comments before the 414 test failing is due to benoitc/gunicorn#2487

krivard:
You mention this is a proof of concept -- what are the next steps? Are there enough decisions that need to be made that we should schedule some discussion time?

sgratzl:
> You mention this is a proof of concept -- what are the next steps? Are there enough decisions that need to be made that we should schedule some discussion time?

since it is a fundamental change there are some open tasks and questions:
 * is the current structure/approach fine, or are they any ideas for improvements
 * besides python+flask changes include
   * pass secrets via environment variables which is docker friendly
   * use of SQLAlchemy as a wrapper for the MySQL connection
   * have the `api.php` as a fallback but also provide direct endpoints `/covidcast`

tasks:
 * since I don't have much data (especially for the protected one) I cannot be certain that all the endpoints remain working
 * performance/scalability testing
 * ?

sgratzl:
closing in favor of #386 (with a branch in the same repo for CI purposes)

PR #355, opened 2021-01-06
Acquisition code for to-be-developed sensor files

chinandrew:
In preparation for the nowcaster, adding an acquisition step for loading and storing sensorized data into a database table. I'm having it follow a similar pattern to the covidcast data, although the exact fields will be different. I'm also using the pandas to_sql method to do inserts which seemed convenient here, and requires some dependencies to be added (https://github.com/cmu-delphi/operations/pull/53). This is different than how covidcast does it (which inserts individual rows in a loop), and the archive logic is similar but not exactly the same as well, so let me know if that's not preferred. My intention was just to make this as lightweight as possible for now.

The table itself will be developed separately and has not been completely specified yet.

Will add more README content in a later PR

Summary of Changes:
- Add module for loading and inserting CSV files into a table. Exact table definition is TBD but is easy to change.

TODO
- [x] Tests: unit tests in this PR, integration tests in a follow up PR once the database is implemented. I think the main function ill do as an integration test since it'd basically just 99% mocked in a unit test.
- [x] Add last update timestamp column

@krivard since I can't tag reviewers. 

cc: @mariajahja


chinandrew:
So I ran unit tests which passed, but integration tests are failing on main for me and thus are also failing for this branch. looking into it

chinandrew:
> So I ran unit tests which passed, but integration tests are failing on main for me and thus are also failing for this branch. looking into it

I lied, was running an old docker container :(. Tests are fine on main and this PR

src/acquisition/covidcast_nowcast/README.md
```diff
@@ -0,0 +1,3 @@
+# COVIDCast Nowcasting Sensor Data
+
+The Nowcasting indicator uses sensorized indicator data to generate and update nowcasts.
```
krivard, line 3:
```suggestion
The Nowcasting indicator uses sensorized indicator data to generate and update COVID nowcasts.
```

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -0,0 +1,79 @@
...
+
+SENSOR_CSV_PATH = "/common/covidcast-nowcast/receiving/"
+TABLE_NAME = "covidcast_nowcast"
+DB_NAME = "epidata"
+CSV_DTYPES = {"geo_value": str, "value": float}
+
+
+def main(csv_path: str = SENSOR_CSV_PATH) -> None:
+    """
+    Parse all files in a given directory and insert them into database sensor table.
```
krivard, line 17:
```suggestion
    Parse all files in a given directory and insert them into the sensor table in the database.
```

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -0,0 +1,79 @@
...
+
+    For all the files found recursively in csv_path that match the naming scheme specified by
+    CsvImporter.find_csv_files(), attempt to load the insert them into the database. Files which do
```
krivard, line 20:
```suggestion
    CsvImporter.find_csv_files(), attempt to load and insert them into the database. Files which do
```

krivard (approved):
lgtm, just docs fixes

PR #356, opened 2021-01-07
Nowcast sensor endpoint

chinandrew:
Based off #355, only new commits are f3c84f2 and onward. Basically steals a bunch of the covidcast code for the PHP and uses a s lightly different table definition (which is still to be finalized).

TODO:
- [x] more integration tests, especially testing the acquisition flow.
- [x] update client

@krivard for review (let me know if theres anyone else I should tag)

cc: @mariajahja

### DIFF
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -448,7 +448,7 @@ class DelphiEpidataPythonClientTests(unittest.TestCase):
       'message': 'success',
     })
 
-    # request as_od
+    # request as_of
     response = Epidata.covidcast_nowcast(
       'src', 'sig1', 'day', 'county', 20200101, '01001',
       as_of=20200101)
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -30,6 +30,7 @@ class DelphiEpidataPythonClientTests(unittest.TestCase):
         database='epidata')
     cur = cnx.cursor()
     cur.execute('truncate table covidcast')
+    cur.execute('truncate table covidcast_nowcast')
     cnx.commit()
     cur.close()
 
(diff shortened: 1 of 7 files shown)


krivard (approved):
lgtm

src/server/api.php
```diff
@@ -2028,6 +2092,44 @@ function meta_delphi() {
...
+        foreach ($epidata as $row) {
+          if ($key != $row['signal']) {
+            $key = $row['signal'];
+            $epi_tree[$key] = array();
+          }
+          unset($row['signal']);
+          array_push($epi_tree[$key],$row);
+        }
+        $epidata = array($epi_tree);
+      }
```
krivard, line 2130: recommend dropping this unless you have an actual use case for multi-signal queries with results grouped by signal

krivard (commented):
Whoops, github lied to me about how many changes were in here. Not done yet, I'll be able to take a closer look tomorrow.

chinandrew:
> Whoops, github lied to me about how many changes were in here. Not done yet, I'll be able to take a closer look tomorrow.

No worries, it's certainly on the longer side. Let me know if I can make the commits more clear (the first one, f3c84f2, is the big one.)

src/server/api.php
chinandrew: good call, will remove.

integrations/client/test_delphi_epidata.py
```diff
@@ -385,3 +386,88 @@ def test_covidcast_meta(self):
...
+        'geo_value': '01001',
+        'value': 1.5,
+        'issue': 20200102,
+        'lag': 2,
+        'signal': 'sig1',
+      }],
+      'message': 'success',
+    })
+
+    # request as_od
```
krivard, line 451:
```suggestion
    # request as_of
```

integrations/server/test_covidcast_nowcast.py
```diff
@@ -0,0 +1,127 @@
...
+    self.cnx = cnx
+    self.cur = cnx.cursor()
+
+  def tearDown(self):
+    """Perform per-test teardown."""
+    self.cur.close()
+    self.cnx.close()
+
+  def test_query_without(self):
+    """Query without specifying an issue."""
```
krivard, line 42:
```suggestion
  def test_query(self):
    """Query nowcasts using default and specified issue."""
```

integrations/server/test_covidcast_nowcast.py
```diff
@@ -0,0 +1,127 @@
...
+      'epidata': [{
+        'signal': 'sig',
+        'time_value': 20200101,
+        'geo_value': '01001',
+        'value': 3.5,
+        'issue': 20200101,
+        'lag': 2,
+       }],
+      'message': 'success',
+    })
```
krivard, line 127: newline?

src/client/delphi_epidata.py
```diff
@@ -665,3 +665,44 @@ def covid_hosp_facility_lookup(
       raise Exception('one of `state`, `ccn`, `city`, `zip`, or `fips_code` is required')
     # Make the API call
     return Epidata._request(params)
+
+  # Fetch Delphi's COVID-19 Surveillance Streams
```
krivard, line 669: No idea why we comment these twice, but they should both be correct

```suggestion
  # Fetch Delphi's COVID-19 Nowcast sensors
```

src/client/delphi_epidata.py
```diff
@@ -665,3 +665,44 @@ def covid_hosp_facility_lookup(
...
+  @staticmethod
+  def covidcast_nowcast(
+          data_source, signals, time_type, geo_type,
+          time_values, geo_value, as_of=None, issues=None, lag=None, **kwargs):
+    """Fetch Delphi's COVID-19 Nowcast sensors"""
+    # also support old parameter name
+    if signals is None and 'signal' in kwargs:
+      signals=kwargs['signal']
```
krivard, line 677: This is a new endpoint so there is no need to make it backwards-compatible with anything
```suggestion
```

src/client/delphi_epidata.py
```diff
@@ -665,3 +665,44 @@ def covid_hosp_facility_lookup(
...
+    if issues is not None:
+      params['issues'] = Epidata._list(issues)
+    if lag is not None:
+      params['lag'] = lag
+
+    if 'format' in kwargs:
+      params['format'] = kwargs['format']
+
+    # Make the API call
+    return Epidata._request(params)
```
krivard, line 708: newline?

src/server/api.php
```diff
@@ -2028,6 +2092,30 @@ function meta_delphi() {
       $epidata = get_covid_hosp_facility_lookup($state, $ccn, $city, $zip, $fips_code);
       store_result($data, $epidata);
     }
+  } else if($source === 'covidcast_nowcast') {
+    if(require_all($data, array('data_source', 'time_type', 'geo_type', 'time_values'))
+       && require_any($data, array('signal', 'signals'))
```
krivard, line 2097: i mean I guess you could do this as accepting either `signal` or `signals` just to be nice, but it's up to you, there's arguably no existing user base to consider
```suggestion
    if(require_all($data, array('data_source', 'time_type', 'geo_type', 'time_values', 'signals'))
```

krivard (changes requested):
Some docs and naming to fix. optionally nuke the backwards-compatible parameter names. if you'd rather maintain parallel syntax with covidcast, then annotate both copies to that effect, so we don't forget.

chinandrew:
Thanks for catching. Removed the old `signal` arguments

krivard (approved):
nice work

PR #359, opened 2021-01-08
Covidcast upload early exit

krivard:
**WARNING** this PR includes a commit reducing the size of the covid_hosp_meta.revision_timestamp column from 1024 to 512. Nothing depends on that change except compliance with MariaDB maximum key length, but once this is merged we should make that change in prod. No value in that column is presently more than 23 characters, so shouldn't be an issue.

**Actual purpose:** Covidcast Acquisition currently counts the number of rows in the `covidcast` table regardless of whether the `receiving` directory is empty. This operation now takes upwards of 40 minutes, which is silly.

This PR rearranges the upload script somewhat so that it first checks `receiving`, then exits as a no-op if it finds no files.

In the process, I merged the default and with-specific-issue routines.

Unit tests updated and pass.

~Integration tests I seem to have forgot about. I'll update here when they're ready.~

Integration tests work automagically because they use no mocks.

Added a test to verify early-exit.

### DIFF
--- a/src/acquisition/covidcast/csv_to_database.py
+++ b/src/acquisition/covidcast/csv_to_database.py
@@ -145,7 +145,7 @@ def main(
 
   if args.is_wip_override and args.not_wip_override:
     print('conflicting overrides for forcing WIP option!  exiting...')
-    exit()
+    return
   wip_override = None
   if args.is_wip_override:
     wip_override = True
@@ -156,7 +156,7 @@ def main(
   path_details = collect_files_impl(args.data_dir, args.specific_issue_date)
   if not path_details:
     print('nothing to do; exiting...')
-    exit()
+    return
     
   database = database_impl()
   database.connect()
--- a/src/acquisition/covidcast/csv_to_database.py
+++ b/src/acquisition/covidcast/csv_to_database.py
@@ -32,107 +32,70 @@ def get_argument_parser():
     help='overrides all signals to mark them as *not* WIP.  NOTE: specify neither or only one of --is_wip_override and --not_wip_override.')
   return parser
 
-def scan_issue_specific(data_dir, database, is_wip_override=None, csv_importer_impl=CsvImporter, file_archiver_impl=FileArchiver):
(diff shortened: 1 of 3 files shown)


krivard:
Tagging George to check my merge of the with/without specific issue stuff

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -167,33 +136,38 @@ def archive_as_successful(path_src, filename, source):
...
-    scan_upload_archive_impl=scan_upload_archive,
-    scan_issue_specific_impl=scan_issue_specific):
+    collect_files_impl=collect_files,
+    upload_archive_impl=upload_archive):
   """Find, parse, and upload covidcast signals."""
 
   if args.is_wip_override and args.not_wip_override:
     print('conflicting overrides for forcing WIP option!  exiting...')
-    exit()
+    return
```
melange396, line 148: good call (pun intended)

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -32,107 +32,70 @@ def get_argument_parser():
...
-  data_dir: top-level directory where CSVs are stored
-  database: an open connection to the epidata database
-
-  The CSV storage layout is assumed to be as follows:
+  :path_details: output from CsvImporter.find*_csv_files 
+  
+  :database: an open connection to the epidata database
 
-  - Receiving: <data_dir>/receiving/<source name>/*.csv
-  - Archival: <data_dir>/archive/<status>/<source name>/*.csv[.gz]
```
melange396, line 105: do we care about preserving this note about the directory structure?

melange396 (approved):
LGTM

src/acquisition/covidcast/csv_to_database.py
krivard: I thought about that, but it's in this file as well: https://github.com/cmu-delphi/delphi-epidata/blob/main/src/acquisition/covidcast/data_dir_readme.md#layout

& that gets dropped into /common/receiving

PR #360, opened 2021-01-11
add nowcast to deploy json

chinandrew:
Forgot to do this and was causing deploy to fail.

@korlaxxalrok


korlaxxalrok (approved):
LGTM :shipit:

PR #364, opened 2021-01-11
Covid Hosp - Facility Lookup Placement

### LINKED ISSUE #352: Move COVID-19 Hospital Facility Lookup page to the correct section
https://cmu-delphi.github.io/delphi-epidata/api/covid_hosp_facility_lookup.html

This is listed under the toplevel, but it should be under Epidata API (Other Diseases).

It also lists the base URL as  https://delphi.cmu.edu/epidata/api.php, when it should probably be  https://delphi.cmu.edu/epidata/api.php?source=covid_hosp_facility_lookup. At the very least we should specify somewhere prominent what `source` (or `endpoint`, depending on when this gets fixed) should be set to.

Akvannortwick:
Hopefully fixes #352


krivard (approved):
double-checked on a fork that this does indeed fix the issue, good work

PR #365, opened 2021-01-12
Correct nowcast parameter spelling in requests

krivard:
This got missed when we removed the backwards-compatible `signal` parameter that had been copy-pasted from covidcast


chinandrew (approved):
thanks for catching

:shipit::shipit::shipit:

PR #371, opened 2021-01-12
Make Delphi-Epidata tests Pytest compatible.

chinandrew:
While discussing docker testing with @dshemetov and playing around with pytest a bit, realized that supporting it might not be too difficult and went ahead and did this.

This change allows for pytest to be run on the delphi-epidata unit tests and integration tests. They still must be run through the docker image since the docker replicates the file structure on the actual automation server which isn't the same as how things are locally. I'm not sure theres a simple way such that both docker and nondocker can be run at the same time.

These changes are backwards compatible with py3tester, so you can use whichever test suite you prefer. I didn't actually change any test syntax (e.g. assertEqual(a,b) -> assert a==b) since pytest can use unittest syntax.

Pytest can be run with
`docker run --rm delphi_python   pytest repos/delphi/delphi-epidata/tests/`
or
`docker run -it --rm delphi_python   pytest repos/delphi/delphi-epidata/tests/ --pdb`,
which will put you into an interactive pdb shell in the docker container if a test fails (which was my primary motivation for getting this to work).
(similar commands for integrations folder but with /integrations/ instead of /tests/)

I've verified that both pytest and py3tester conduct 64 passing unit  tests and 36 passing integration tests (same as main). pytest raises 2-3 warnings from src/acquisition/covid_hosp/common/test_utils, which contains a class called TestUtils that has test helper functions but is not a test itself. pytest just warns that it's not considering that class a test, which is correct.

Summary of changes:
- Import necessary functions/classes for all tests
- Add __init__ files to each dir with tests in it, with commands to append the post-dockerfile copy operation path for imports
- Update dev guides stating this is possible now

@krivard for review.


tests/acquisition/covid_hosp/common/test_network.py
```diff
@@ -5,10 +5,13 @@
 from unittest.mock import MagicMock
 from unittest.mock import sentinel
 
+from delphi.epidata.acquisition.covid_hosp.common.network import Network
+
 # py3tester coverage target
 __test_target__ = 'delphi.epidata.acquisition.covid_hosp.common.network'
 
 
+
```
krivard, line 14:
```suggestion
```

krivard (approved):
xlnt

Issue #289, opened 2020-11-17
Update COVIDcast metadata on insert


krivard:
Metadata updates are starting to be a real drag. If we refactor to update the metadata when new rows are inserted, we won't have to scan the entire database to figure out what changed. 

Bonus: we won't need to do any `wip` computations at all, since the flag value is already known from the name of the CSV file.

Issue #368, opened 2021-01-12
Consider incremental covidcast meta data update


sgratzl:
taking a look at:
```
MIN(t.`time_value`) AS `min_time`,
        MAX(t.`time_value`) AS `max_time`,
        COUNT(DISTINCT t.`geo_value`) AS `num_locations`,
        MIN(`value`) AS `min_value`,
        MAX(`value`) AS `max_value`,
        ROUND(AVG(`value`),7) AS `mean_value`,
        ROUND(STD(`value`),7) AS `stdev_value`,
        MAX(`value_updated_timestamp`) AS `last_update`,
        MAX(`issue`) as `max_issue`,
        MIN(`lag`) as `min_lag`,
        MAX(`lag`) as `max_lag`
```

a lot of these values could be easily incrementally computed (min, max) others with some effort (avg, std, count distinct). There are statistical versions of stable incremental avg/std that we could explore. If we cannot compute the changed rows per run, we could use a multi stage approach like:

general assumption: it is unlikely that values older than a month will get an new issue. 

1. we compute the metadata for all data which is older than e.g., December 2020 with additional information for the incremental metadata (count, list of distinct geo_values, ...)
2. each day we compute the metadata for all data younger than December 2020
3. merge the old cached with the new ones to produce the final meta data
4. every (two) weeks we shift the split date to reduce the amount of data we have to compute during each update

in the end it depends on finding a good method for incrementally (or in batches) find the mean/avg and variance/standard deviation.

krivard:
Yep. I believe @melange396 is working on a draft of this. Brief summary of offline discussion -- 

# assumption: values older than a month are unlikely to change

Inpatient, outpatient, and testing signals have 60-80 days of backfill. typical change between first and final issue is less than 20%, but individual regions can change by 300% or more.

some things we've thought about in response to this:
* prototype incremental updates in a separate table so we can easily compare with the exact values
* run a full exact update once a week (or otherwise periodically) to correct any estimates that have slipped

# different definitions of incremental

* just today's issue
* two-way split between 1-2 months ago (proposed above)
* monthly or quarterly bins (easier to support mass backissue uploads or extreme backfill)

# things we can ignore

* number of regions (do we even have users who need this)

# incremental math

mean is easy; stdev is slightly more complicated. [here's a closed form for adding a single sample](https://math.stackexchange.com/questions/102978/incremental-computation-of-standard-deviation); some work/research needed for adding a whole batch efficiently.

[...]

PR #375, opened 2021-01-14
incremental cache updating prototype

melange396:
re: issue https://github.com/cmu-delphi/delphi-epidata/issues/289

this is completely ignorant of re-issued data, so when a new issue of data happens, the cache may be somewhat stale and/or inaccurate.  the inaccuracies will probably be relatively small if the reissues are not wildly different than the previous issue.  also, those differences will get washed out as data series continue to grow in size.  a ~weekly full population of the cache should probably be done to clean up those inaccuracies.

this will break ingestion if new data is loaded before fully populating the cache with the existing data, because it expects a new field in the cache: 'num_points'.

the code is currently saving results to a 'test' cache table for comparing with the regular/main cache table.  TODOs have been included for lines that need to be reverted if/when testing is successful.

i have a feeling that the incremental changes could result in some loss of precision for the sd values, but i may just be paranoid.  theoretically, the results should be identical if floats have infinite precision.

changed 'Database.get_covidcast_meta()' to 'Database.compute_covidcast_meta()' for clarity


melange396:
oops, i somehow ran the wrong tests on this (copy pasta failed me)...  fixing

melange396:
should be good now, though i should introduce some new tests for this

krivard:
Is this more relevant to #368 or #289?

melange396:
yes

melange396:
but seriously, im not sure.  it sorta fits both.  ¯\\\_(ツ)\_/¯

src/acquisition/covidcast/database.py
```diff
@@ -213,10 +213,15 @@ def insert_or_update_batch(self, cc_rows, batch_size=2**20, commit_partial=False
 
 
         result = self._cursor.executemany(insert_into_tmp_sql, args)
+
+        meta_update = self.compute_covidcast_meta(table_name=tmp_table_name)
```
krivard, line 217: annotating since i definitely missed this the first time through

```suggestion
        # use the temp table to compute meta just for new data, 
        # then merge with the global meta
        meta_update = self.compute_covidcast_meta(table_name=tmp_table_name)
```

src/acquisition/covidcast/database.py
```diff
@@ -342,4 +348,48 @@ def retrieve_covidcast_meta_cache(self):
...
+      # combined stdev formula shamelessly stolen from:
+      # https://math.stackexchange.com/questions/2971315/how-do-i-combine-standard-deviations-of-two-groups
+      sx, xbar, n = [mainstats[i] for i in ('stdev_value', 'mean_value', 'num_points')]
+      sy, ybar, m = [updtstats[i] for i in ('stdev_value', 'mean_value', 'num_points')]
+      comb_mean = ( xbar*n + ybar*m ) / (n+m)
+      comb_sd = sqrt(
+        ( (n-1)*sx*sx + (m-1)*sy*sy ) / (n+m-1)
+        +
+        n*m*(xbar-ybar)**2 / ((n+m)*(n+m-1))
+      )
```
krivard, line 383: double checked parens, all correct 💪

integrations/acquisition/covidcast/test_covidcast_meta_caching.py
```diff
@@ -125,7 +126,7 @@ def test_caching(self):
     self.cur.execute('''
       update covidcast_meta_cache set
         timestamp = UNIX_TIMESTAMP(NOW()),
-        epidata = '[{"hello": "world"}]'
+        epidata = '[{"data_source": "hell", "signal": "temp", "time_type": "inf", "geo_type": "void", "is_wip": 0}]'
```
krivard, line 129: public repo, alas
```suggestion
        epidata = '[{"data_source": "world", "signal": "temp", "time_type": "inf", "geo_type": "void", "is_wip": 0}]'
```

krivard (approved):
Looks largely reasonable, pending clarification on what the default values of `table_name` parameters should be. Would also be good to write up the testing and/or deployment procedure, ie

1. run {SQL}
2. run {python}
3. check {data}
4. run {SQL}
5. etc

integrations/acquisition/covidcast/test_covidcast_meta_caching.py
melange396: :'(

src/acquisition/covidcast/database.py
```diff
@@ -327,12 +333,12 @@ def update_covidcast_meta_cache(self, metadata):
...
 
-  def retrieve_covidcast_meta_cache(self):
+  def retrieve_covidcast_meta_cache(self, table_name='covidcast_meta_cache_test'): # TODO: remove arg
     """Useful for viewing cache entries (was used in debugging)"""
 
-    sql = '''
+    sql = f'''
       SELECT `epidata`
-      FROM `covidcast_meta_cache`
+      FROM `{table_name}` -- TODO: remove '_test' after successfully testing incremental cache updates
```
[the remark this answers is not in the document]
melange396: well, both lines will change when we take this out of testing

src/acquisition/covidcast/database.py
```diff
@@ -327,12 +333,12 @@ def update_covidcast_meta_cache(self, metadata):
 
     self._cursor.execute(sql, (epidata_json,))
 
-  def retrieve_covidcast_meta_cache(self):
+  def retrieve_covidcast_meta_cache(self, table_name='covidcast_meta_cache_test'): # TODO: remove arg
```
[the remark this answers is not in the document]
melange396: negative, but they will match when we are out of testing

melange396:
deployment:
- halt any new csv ingestion ( `src/acquisition/covidcast/csv_to_database.py:main()` )
- execute this sql:
```
  CREATE TABLE `covidcast_meta_cache_test` (
    `timestamp` int(11) NOT NULL,
    `epidata` LONGTEXT NOT NULL,
    PRIMARY KEY (`timestamp`)
  ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
  INSERT INTO covidcast_meta_cache_test VALUES (0, '[]');
```
- run full metadata update process ( `src/acquisition/covidcast/covidcast_meta_cache_updater.py:main()` )
- check tables `covidcast_meta_cache[_test]` have matching entries and include field '`num_points`'
- resume new data file ingestion
- after some new data is loaded and before a full cache refresh is run again, check tables do not have matching entries (but they should still both have sane and not-too-different values)

src/ddl/covidcast.sql
```diff
@@ -138,4 +138,13 @@ CREATE TABLE `covidcast_meta_cache` (
   `epidata` LONGTEXT NOT NULL,
   PRIMARY KEY (`timestamp`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
-INSERT INTO covidcast_meta_cache VALUES (0, '');
+INSERT INTO covidcast_meta_cache VALUES (0, '[]');
```
benjaminysmith, line 141: Why do you need this line?

src/acquisition/covidcast/database.py
```diff
@@ -327,12 +335,12 @@ def update_covidcast_meta_cache(self, metadata):
...
```
benjaminysmith, line 338: Why is this defaulting to the test table rather then the normal one (compared to the method above which defaults to the existing one)

src/acquisition/covidcast/database.py
```diff
@@ -213,10 +213,17 @@ def insert_or_update_batch(self, cc_rows, batch_size=2**20, commit_partial=False
 
 
         result = self._cursor.executemany(insert_into_tmp_sql, args)
+
+        # use the temp table to compute meta just for new data, 
+        # then merge with the global meta
+        meta_update = self.compute_covidcast_meta(table_name=tmp_table_name)
```
benjaminysmith, line 219: I'm probably missing somethign really obvious, but where is tmp_table_name set?

benjaminysmith, line 219: If I understand correctly the (temp) metadata will be updated on every write to the database, is that right? How many times do we expect that to happen each day? (i.e., is it for each signal/source update, or do those happen all at once)

Would a possible alternative be to scan through the database once each day and update then?

src/acquisition/covidcast/database.py
```diff
@@ -342,4 +350,48 @@ def retrieve_covidcast_meta_cache(self):
     cache_hash = {}
     for entry in cache:
       cache_hash[(entry['data_source'], entry['signal'], entry['time_type'], entry['geo_type'])] = entry
+    if not len(cache) and table_name=='covidcast_meta_cache_test': # TODO: remove line
+      print("test cache empty -- falling back to regular cache ; this should happen ONLY ONCE (after regular cache populated)") # TODO: remove line
```
benjaminysmith, line 354: Don't need to do it now but please add a todo or issue for me to add log statements here.

src/ddl/covidcast.sql
melange396: empty string makes the JSON decoder puke.  an empty list is the "correct" cache for a new db w/o any data.

src/acquisition/covidcast/database.py
melange396: this is for keeping things consistent when dealing with the temp table...  im essentially side stepping the regular cache except for edge cases, and things will go back to a less complicated state when we decide we like this type of caching. 
 the method above existed already and the default value preserves the previous behavior.

src/acquisition/covidcast/database.py
melange396: line 107.  its a part of the normal ingestion process.

melange396: its pretty much one update to the cache for each input file (csv).  i *believe* the importer runs every so often as a cron job or whatever (multiple times daily?), and loads all new files in the data dump directory.  i also *believe* the files do not arrive on a strict schedule.  one file is unique in terms of (source, signal, time_type (day|week), geo_type (state|county|metro), time_value (yyyymmdd), issue_value (yyyymmdd)).

we could do it once a day but that includes some logic to identify which subset of data should be separated out.  doing it per file gives us the aggregation we need for free.

src/acquisition/covidcast/database.py
melange396: heres what it looks like when its been cleaned up to remove the temp table: https://github.com/cmu-delphi/delphi-epidata/pull/377

src/acquisition/covidcast/database.py
melange396: done

src/acquisition/covidcast/database.py
krivard: `csv_to_database.py` runs at :50 after every hour, with a no-op exit if there are no files in receiving. So, up to 24x a day, but the true frequency depends on how many of those periods include data drops.

The number of files handled during each run varies from 0 to 53k. About half of the 800 runs in the last log file handled 0 files, 3/4 under 1500 files, but there is that long tail where we'd be merging metadata 10k times in a single run. 

It _might_ be performant as-is, but we should probably think about whether there's a way to efficiently merge a whole pile of these.

benjaminysmith: @melange396 makes sense that this is simpler.

@krivard thanks, so this means we will need to monitor for increased runtime of csv_to_database once this is enabled. 

wdyt about just enabling and monitoring for a day? We can scrape that data as we are doing for the metadata update. Probably best to do it on monday when we are all around.

benjaminysmith (approved):

src/acquisition/covidcast/database.py
krivard: good idea. probably two ways to do it: on staging (easy, but won't see as many data files) or in production (either need to merge & back out a PR, or coordinate with @korlaxxalrok to switch prod to sync to a different branch and then switch back)

PR #378, opened 2021-01-14
fix state daily import

chinandrew:
Make state daily use state daily database and network (instead of state timeseries database and network)


krivard (approved):
👍

PR #384, opened 2021-01-19
Document upcoming V4 aggregates

capnrefsmmat:


docs/api/covidcast-signals/fb-survey.md
```diff
@@ -326,6 +326,11 @@ also available. These have names beginning `smoothed_w`, such as
 | --- | --- | --- |
 | `smoothed_accept_covid_vaccine` | Estimated percentage of respondents who would definitely or probably choose to get vaccinated, if a COVID-19 vaccine were offered to them today. **Note:** Until January 6, 2021, all respondents answered this question; beginning on that date, only respondents who said they have not received a COVID vaccine are asked this question. <br/> **First Available:** 2021-01-01 | V3 |
 | `smoothed_covid_vaccinated` | Estimated percentage of respondents who have already received a vaccine for COVID-19. **Note:** The Centers for Disease Control compiles data on vaccine administration across the United States. This signal may differ from CDC data because of survey biases and should not be treated as authoritative. However, the survey signal is not subject to the lags and reporting problems in official vaccination data. <br/> **First Available:** 2021-06-01 | V1 |
```
krivard, line 328: no idea how this got in there, but we're using yyyy-mm-dd for First Available

```suggestion
| `smoothed_covid_vaccinated` | Estimated percentage of respondents who have already received a vaccine for COVID-19. **Note:** The Centers for Disease Control compiles data on vaccine administration across the United States. This signal may differ from CDC data because of survey biases and should not be treated as authoritative. However, the survey signal is not subject to the lags and reporting problems in official vaccination data. <br/> **First Available:** 2021-01-06 | V1 |
```

krivard (approved):
Found an unrelated sneaky date format error, but the rest looks great.

PR #391, opened 2021-01-21
Have pandas upsert to db and clean up archive code

chinandrew:
Pandas to_sql only inserts and doesnt update if there's a duplicate unique key, so this change makes it upsert. Full disclosure, i just copy pasted most of this off a [stackoverflow answer ](https://stackoverflow.com/a/60891952) since im not very familiar with how sqlalchemy and pandas sql insertion methods work. Wrote a test that failed before, and this code made it pass :shrug: 

Also cleaned up some file moving code to reduce duplication and handle generating directories if they don't exist.


integrations/acquisition/covidcast_nowcast/test_csv_uploading.py
```diff
@@ -118,3 +116,39 @@ def test_uploading(self):
...
+      'epidata': [{
+        'time_value': 20200419,
+        'geo_value': 'ca',
+        'value': 2,
+        'issue': 20200421,
+        'lag': 2,
+        'signal': 'sig',
+      }],
+      'message': 'success',
+    })
```
krivard, line 154: add trailing newline

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -75,5 +81,23 @@ def load_and_prepare_file(filepath: str, attributes: tuple) -> pd.DataFrame:
...
+    print(f"{filepath} moved to {archive_dir}")
+
+
+def _create_upsert_method(meta):
+    def method(table, conn, keys, data_iter):
+        sql_table = sqlalchemy.Table(table.name, meta, autoload=True)
+        insert_stmt = sqlalchemy.dialects.mysql.insert(sql_table).values([dict(zip(keys, data)) for data in data_iter])
+        upsert_stmt = insert_stmt.on_duplicate_key_update({x.name: x for x in insert_stmt.inserted})
+        conn.execute(upsert_stmt)
+    return method
```
krivard, line 99: looks legit!

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -33,17 +36,20 @@ def main(csv_path: str = SENSOR_CSV_PATH) -> None:
     """
     user, pw = secrets.db.epi
     engine = sqlalchemy.create_engine(f"mysql+pymysql://{user}:{pw}@{secrets.db.host}/{DB_NAME}")
-    for filepath, attributes in CsvImporter.find_csv_files(csv_path):
-        if attributes is None:
-            move(filepath, filepath.replace("receiving", "archive/failed"))
+    for filepath, attr in CsvImporter.find_csv_files(csv_path):
```
krivard, line 39: does this rename have a semantic purpose, or just to save typing?

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -33,17 +36,20 @@ def main(csv_path: str = SENSOR_CSV_PATH) -> None:
...
-            data.to_sql(TABLE_NAME, engine, if_exists="append", index=False)
+            data = load_and_prepare_file(filepath, attr)
+            conn = engine.connect()
+            with conn.begin():
+                method = _create_upsert_method(sqlalchemy.MetaData(conn))
+                data.to_sql(TABLE_NAME, engine, if_exists="append", method=method, index=False)
         except Exception:
-            move(filepath, filepath.replace("receiving", "archive/failed"))
+            _move_after_processing(filepath, success=False)
             raise
```
krivard, line 51: Do you need a `finally` block to do any connection cleanup with the database? I know the mysql.connect library requires something like
https://github.com/cmu-delphi/delphi-epidata/blob/e6715c0602d889324f48153f6ad5e02409d678d8/src/acquisition/covidcast/csv_to_database.py#L179

but not sure if sqlalchemy needs something similar

krivard (changes requested):
nit + safety question

src/acquisition/covidcast_nowcast/load_sensors.py
chinandrew: i wondered that too when i first implemented it, since ive had to cloes sqlalchemy connections before, but the to_sql method can take an engine and not a connection. Based on [this answer](https://stackoverflow.com/a/51242577), looks like it isn't needed. The connection used for the connection metadata is in a context manager so i think thats ok as well

src/acquisition/covidcast_nowcast/load_sensors.py
chinandrew: i had shortened it to be lazy about a line that was too long, but fixed it properly. ill change it back.

krivard (approved):
👍

PR #393, opened 2021-01-22
Added repairs section to covid_hosp docs

krivard:
On January 22 we made repairs to two issues of covid_hosp_state_timeseries data to unscramble their columns. It's good to write these kinds of things down.


chinandrew (approved):
:shipit: :shipit: :shipit:

PR #395, opened 2021-01-26
Have nowcasting acquisition use issue date from file

chinandrew:
Since nowcasting uses sensor values computed as_of certain dates, the issue dates need to be the as_of dates and not datetime.today() like other acquisitions. These as_of dates are specified during sensorization (it calls the covidcast as_of indicator data to generate the sensors) and added to the files to be ingested as a new column.

cc @mariajahja

### DIFF
--- a/src/acquisition/covidcast_nowcast/load_sensors.py
+++ b/src/acquisition/covidcast_nowcast/load_sensors.py
@@ -13,7 +13,7 @@ SUCCESS_DIR = "archive/successful"
 FAIL_DIR = "archive/failed"
 TABLE_NAME = "covidcast_nowcast"
 DB_NAME = "epidata"
-CSV_DTYPES = {"sensor_name": str, "geo_value": str, "value": float, "issue": int}
+CSV_DTYPES = {"sensor_name": str, "geo_value": str, "value": float}
 
 
 def main(csv_path: str = SENSOR_CSV_PATH) -> None:
@@ -37,7 +37,8 @@ def main(csv_path: str = SENSOR_CSV_PATH) -> None:
     """
     user, pw = secrets.db.epi
     engine = sqlalchemy.create_engine(f"mysql+pymysql://{user}:{pw}@{secrets.db.host}/{DB_NAME}")
-    for filepath, attribute in CsvImporter.find_csv_files(csv_path):
+#    for filepath, attribute in CsvImporter.find_csv_files(csv_path):
+    for filepath, attribute in CsvImporter.find_issue_specific_csv_files(csv_path):
         if attribute is None:
             _move_after_processing(filepath, success=False)
             continue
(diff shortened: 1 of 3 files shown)


krivard:
Just to confirm, when running in production you expect each line of the CSV might have a different issue date?

chinandrew:
> Just to confirm, when running in production you expect each line of the CSV might have a different issue date?

That's supported here, though the way it's currently implemented only 1 issue gets produced at a time so the CSV files should be homogeneous in that field

krivard:
If you don't have a use case that needs a different issue per line, I'd prefer putting the issue into the filename instead of the file content -- saves a ton on disk space for the archived csvs.

chinandrew:
> If you don't have a use case that needs a different issue per line, I'd prefer putting the issue into the filename instead of the file content -- saves a ton on disk space for the archived csvs.

That makes sense to me. Since I'm sharing the covidcast CsvImporter that auto-detects source/geo/signal from the filename, would you recommend embedding the issue into signal name and parsing it out later (e.g. make the signal name <issuedate>_<actualsignalname>), adjusting CsvImporter, ,writing new code to parse, or something else

krivard:
Borrow the issue-specific importer instead 😄 

https://github.com/cmu-delphi/delphi-epidata/blob/2395b487feb760c02c056d27f022c10ef43cbbce/src/acquisition/covidcast/csv_importer.py#L82

Format is `issue_{issue}/{source}/{normal covidcast filename}`

chinandrew:
> Borrow the issue-specific importer instead 
> 
> https://github.com/cmu-delphi/delphi-epidata/blob/2395b487feb760c02c056d27f022c10ef43cbbce/src/acquisition/covidcast/csv_importer.py#L82
> 
> Format is `issue_{issue}/{source}/{normal covidcast filename}`

Oh cool, didn't know about that. Thanks for pointing it out, I'll work on integrating it now

chinandrew:
@krivard added issue specific csv importer, ready for review again

src/acquisition/covidcast_nowcast/load_sensors.py
```diff
@@ -36,7 +37,8 @@ def main(csv_path: str = SENSOR_CSV_PATH) -> None:
     """
     user, pw = secrets.db.epi
     engine = sqlalchemy.create_engine(f"mysql+pymysql://{user}:{pw}@{secrets.db.host}/{DB_NAME}")
-    for filepath, attribute in CsvImporter.find_csv_files(csv_path):
+#    for filepath, attribute in CsvImporter.find_csv_files(csv_path):
```
krivard, line 40:
```suggestion
```

chinandrew: oops, thanks

krivard (approved):
xlnt

PR #396, opened 2021-01-28
Add async method to epidata

chinandrew:
To be used by covidcast client (to speed up indicator combo) and also in nowcasting. 

One thing to note is that running this on some machines and/or in jupyter can lead to this an issue that [requires you run these two lines to solve](https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python/56434301#56434301). For that reason I'm not sure we'll want to make this the default method in covidcast, since we don't have much control over people's environments. Adding those lines at the top of the client file seems to resolves it for me, but not sure if that'll hold universally. If it does, then we might be fine. If not, we could set a protected flag in covidcast for async that only our internal users/superusers will change. 

Summary of changes:
- Added a method which takes in a list of raw params and sends an async request for each param. Does not do any formatting/cleaning/parsing of the params so it's compatible with any epidata source (at the expense of convenient) 2 helper methods are also added to make the requests.
- Tests

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -715,7 +715,7 @@ class Epidata:
       return await response.json(), params
 
   @staticmethod
-  async def fetch_epidata(param_combos):
+  async def async_fetch_epidata(param_combos):
     """Helper function to asynchronously make and aggregate Epidata GET requests."""
     tasks = []
     async with ClientSession() as session:
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -732,6 +732,6 @@ class Epidata:
     responses = []
     for batch in batches:
       loop = asyncio.get_event_loop()
-      future = asyncio.ensure_future(Epidata.fetch_epidata(batch))
+      future = asyncio.ensure_future(Epidata.async_fetch_epidata(batch))
       responses += loop.run_until_complete(future)
     return responses
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -10,7 +10,10 @@ Notes:
 
 # External modules
 import requests
+import asyncio
+import warnings
 
+from aiohttp import ClientSession
 from pkg_resources import get_distribution, DistributionNotFound
 
 # Obtain package version for the user-agent. Uses the installed version by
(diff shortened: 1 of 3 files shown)


src/client/delphi_epidata.py
```diff
@@ -704,3 +707,31 @@ def covidcast_nowcast(
 
     # Make the API call
     return Epidata._request(params)
+
+  @staticmethod
+  async def get(params, session):
```
krivard, line 712:
```suggestion
  async def async_get(params, session):
```

src/client/delphi_epidata.py
```diff
@@ -704,3 +707,31 @@ def covidcast_nowcast(
...
+    async with session.get(Epidata.BASE_URL, params=params) as response:
+      return await response.json(), params
+
+  @staticmethod
+  async def fetch_epidata(param_combos):
+    """Helper function to asynchronously make and aggregate Epidata GET requests."""
+    tasks = []
+    async with ClientSession() as session:
+      for param in param_combos:
+        task = asyncio.ensure_future(Epidata.get(param, session))
```
krivard, line 723:
```suggestion
        task = asyncio.ensure_future(Epidata.async_get(param, session))
```

src/client/delphi_epidata.py
```diff
@@ -704,3 +707,31 @@ def covidcast_nowcast(
...
     return Epidata._request(params)
+
+  @staticmethod
+  async def get(params, session):
+    """Helper function to make Epidata GET requests."""
+    async with session.get(Epidata.BASE_URL, params=params) as response:
+      return await response.json(), params
+
+  @staticmethod
+  async def fetch_epidata(param_combos):
```
krivard, line 718:
```suggestion
  async def async_fetch_epidata(param_combos):
```

src/client/delphi_epidata.py
```diff
@@ -704,3 +707,31 @@ def covidcast_nowcast(
...
+      return responses
+
+  @staticmethod
+  def async_call(param_list, batch_size=100):
+    """Make asynchronous Epidata calls for a list of parameters."""
+    batches = [param_list[i:i+batch_size] for i in range(0, len(param_list), batch_size)]
+    responses = []
+    for batch in batches:
+      loop = asyncio.get_event_loop()
+      future = asyncio.ensure_future(Epidata.fetch_epidata(batch))
```
krivard, line 735:
```suggestion
      future = asyncio.ensure_future(Epidata.async_fetch_epidata(batch))
```

krivard (changes requested):
Looks legit, but the documentation for delphi_epidata is so nonexistant that we definitely need people to be able to tell what get/fetch are for just by their names. No accidents from people doing `dir(Epidata)` and guessing

(do get/fetch even need to be static members or can they be local functions within async_call? will you be calling them directly from nowcast?)

chinandrew:
> Looks legit, but the documentation for delphi_epidata is so nonexistant that we definitely need people to be able to tell what get/fetch are for just by their names. No accidents from people doing `dir(Epidata)` and guessing
> 
> (do get/fetch even need to be static members or can they be local functions within async_call? will you be calling them directly from nowcast?)

Good point, only the async_call is the only one that will be used publicly (maybe i should switch the names of async_call and async_fetch_epidata), so the others don't need to be visible. Ill work on it

chinandrew:
Updated to make functions local and also renamed them.

krivard (approved):
👩‍🍳 💋

PR #401, opened 2021-02-02
HHS hosp docs: clarify handling of timeseries vs daily updates

krivard:
It's not currently clear in the docs how Delphi merges data from the timeseries and daily updates published by HHS. This PR aims to do that.

Before merging we should make sure this is all accurate, and decide whether to fix the code or the docs if not.


krivard:
Overridden by a more recent PR.

PR #404, opened 2021-02-03
Add structured logger to epidata, and log metadata generation

benjaminysmith:
- Add a structured logger to the epidata repo
- Add logging for time taken to generate and update metadata

Progress on #403

Note that this needs structlog installed before running since this adds a new dependency.

The logger is a fork of the logger in delphi_utils. We can unify these if we extract delphi utils into a standalone package.

### DIFF
--- a/src/acquisition/covidcast/covidcast_meta_cache_updater.py
+++ b/src/acquisition/covidcast/covidcast_meta_cache_updater.py
@@ -9,7 +9,6 @@ import time
 from delphi.epidata.acquisition.covidcast.database import Database
 from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
 from delphi.epidata.client.delphi_epidata import Epidata
-from delphi.operations import secrets
 
 def get_argument_parser():
   """Define command line arguments."""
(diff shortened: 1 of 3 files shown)


src/acquisition/covidcast/covidcast_meta_cache_updater.py
```diff
@@ -3,30 +3,38 @@
...
 
 
 def main(args, epidata_impl=Epidata, database_impl=Database):
   """Update the covidcast metadata cache.
 
   `args`: parsed command-line arguments
   """
+  logger = get_structured_logger(
+      "metadata_cache_updater",
+      filename=args.log_file)
```
melange396, line 28: i take it this means the log file argument is required...  is there maybe a way to let it default to stdout or somehing?

src/acquisition/covidcast/covidcast_meta_cache_updater.py
```diff
@@ -44,13 +52,22 @@ def main(args, epidata_impl=Epidata, database_impl=Database):
 
   # update the cache
   try:
+    metadata_update_start_time = time.time()
     database.update_covidcast_meta_cache(metadata)
+    metadata_update_interval_in_seconds = time.time() - metadata_update_start_time
```
melange396, line 57: this should always be very quick, but it doesnt hurt to measure it anyway

melange396 (approved):
LGTM

krivard:
Waiting to merge until we have the bandwidth to install structlog on prod

src/acquisition/covidcast/covidcast_meta_cache_updater.py
benjaminysmith: acknowledged

src/acquisition/covidcast/covidcast_meta_cache_updater.py
benjaminysmith: If the arg is empty/none then the logger will default to stdout.

benjaminysmith:
Structlog is in but lets wait to merge until we have the regular system functional again (don't want to introduce additional risk)

krivard:
Primary is relatively healthy; merged

PR #405, opened 2021-02-03
covid_hosp: fetch all files from daily batches

### LINKED ISSUE #402: More sophisticated ingestion of state covid_hosp daily snapshots
I’m working on some documentation revisions in #401, and I’ve come across a small snag — the state daily files are uploaded in batches, not every day. This would be fine, except the metadata which we use to detect new data available from HHS only surfaces the newest revision, and I can’t seem to find an easy way to list all known revisions. This means that our current method skips over the first (n-1) updates in each batch.

After some discussion with Evan, we settled on the following pipeline, which will get us roughly in line with their approach:

1. use metadata to check for a new revision_timestamp (as usual)
2. if there is one, then scrape the [revisions page](https://healthdata.gov/node/3281086/revisions) for all the files posted since the last revision
3. merge those files in chronological order
4. store that as the new issue tagged by revision_timestamp

-- so we don't need to figure out an issue for each update, we can collapse each batch into a single issue.

krivard:
Context: Daily update files are posted in batches of several days at a
time. Only the most recent file is surfaced through metadata, but all updates
are posted on the revisions page.

This PR adapts the state_daily pipeline to:

1. Fetch all files in the batch
2. Merge them (I checked with the Reich folks that this is the desired behavior)
3. Upload as a single issue based on the upload date (=date recorded in metadata)

This requires some substantial changes to Update and tests to account for and
mock out the pipeline differences to common, and minor changes to Network and
Database.

BeautifulSoup (`bs4`) was already listed in `operations.requirements.txt`, so it shouldn't need to be added to prod before merging.

Fixes #402

### DIFF
--- a/src/acquisition/covid_hosp/state_daily/network.py
+++ b/src/acquisition/covid_hosp/state_daily/network.py
@@ -22,6 +22,30 @@ class Network(BaseNetwork):
         *args, **kwags, dataset_id=Network.DATASET_ID)
 
   def fetch_revisions(newer_than):
+    """Scrape CSV urls from the HHS revisions pages.
+
+    The revisions page is a big table showing recent revisions to the state
+    daily dataset. Each row represents a single revision with a link to a
+    separate page where a download is available for that revision. 
+
+    The first row is different, and links directly back to the main dataset
+    page. Since this row represents the current published version, and the URL
+    for that CSV is included in the metadata, we do not retrieve it here.
+
+    The link for each revision is a date in the same format as
+    `revision_timestamp` from the metadata, so we bookend which links we retrieve
+    by checking against the issue code (YYYYMMDD) for the last issue we
+    imported. 
+
+    This approach will break under the following conditions:
+     - Automation starts the update routine after the first file of a batch is
+       uploaded by HHS, but before the final file of the batch is uploaded (HHS
+       typically uploads in the evenings, but there are some midmorning and
+       afternoon batches in the revision history, so who knows)
+     - HHS uploads batches on two days in a row, and the first batch isn't completed
+       until after midnight (causing files from two distinct batches to have the
+       same date in the link from the revisions page)
+    """
     urls = []
     soup = BeautifulSoup(requests.get(Network.ROOT + Network.REVISIONS).content, "html.parser")
     views = [
--- a/src/acquisition/covid_hosp/state_daily/network.py
+++ b/src/acquisition/covid_hosp/state_daily/network.py
@@ -1,10 +1,16 @@
+# third party
+from bs4 import BeautifulSoup
+import requests
+
 # first party
 from delphi.epidata.acquisition.covid_hosp.common.network import Network as BaseNetwork
-
+from delphi.epidata.acquisition.covid_hosp.common.utils import Utils
 
 class Network(BaseNetwork):
 
   DATASET_ID = '7823dd0e-c8c4-4206-953e-c6d2f451d6ed'
+  ROOT = "https://healthdata.gov"
+  REVISIONS = '/node/3281086/revisions'
 
   def fetch_metadata(*args, **kwags):
     """Download and return metadata.
@@ -14,3 +20,15 @@ class Network(BaseNetwork):
 
     return Network.fetch_metadata_for_dataset(
         *args, **kwags, dataset_id=Network.DATASET_ID)
+
+  def fetch_revisions(newer_than):
+    urls = []
+    soup = BeautifulSoup(requests.get(Network.ROOT + Network.REVISIONS).content, "html.parser")
+    views = [
+      a['href'] for a in soup.select("tr.diff-revision a[href^='/node']")
+      if Utils.get_issue_from_revision(a.text) > newer_than
+    ]
+    for view in views:
+      view_soup = BeautifulSoup(requests.get(Network.ROOT + view).content, "html.parser")
+      urls.append(view_soup.select_one(".download a")['href'])
+    return urls
--- a/src/acquisition/covid_hosp/state_daily/network.py
+++ b/src/acquisition/covid_hosp/state_daily/network.py
@@ -9,9 +9,10 @@ from delphi.epidata.acquisition.covid_hosp.common.utils import Utils
 class Network(BaseNetwork):
 
   DATASET_ID = '7823dd0e-c8c4-4206-953e-c6d2f451d6ed'
-  ROOT = "https://healthdata.gov"
+  ROOT = 'https://healthdata.gov'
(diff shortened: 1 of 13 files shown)


src/acquisition/covid_hosp/state_daily/network.py
```diff
@@ -14,3 +20,39 @@ def fetch_metadata(*args, **kwags):
...
+       typically uploads in the evenings, but there are some midmorning and
+       afternoon batches in the revision history, so who knows)
+     - HHS uploads batches on two days in a row, and the first batch isn't completed
+       until after midnight (causing files from two distinct batches to have the
+       same date in the link from the revisions page)
+    """
+    urls = []
+    soup = BeautifulSoup(requests.get(Network.ROOT + Network.REVISIONS).content, "html.parser")
+    views = [
+      a['href'] for a in soup.select("tr.diff-revision a[href^='/node']")
```
chinandrew, line 52: nit: single vs double quotes here and below.

src/acquisition/covid_hosp/state_daily/network.py
```diff
@@ -14,3 +20,39 @@ def fetch_metadata(*args, **kwags):
 
     return Network.fetch_metadata_for_dataset(
         *args, **kwags, dataset_id=Network.DATASET_ID)
+
+  def fetch_revisions(newer_than):
```
chinandrew, line 24: document `newer_than` argument in the docstring

tests/acquisition/covid_hosp/state_daily/test_network.py
```diff
@@ -24,3 +32,17 @@ def test_fetch_metadata(self):
...
+
+    with patch.object(requests, 'get',
+                      side_effect=list(self.test_utils.load_sample_revisions())) as requests_get:
+      result = Network.fetch_revisions(20201210)
+
+      self.assertEqual(requests_get.call_count, 4)
+      self.assertEqual(result, [
+        "https://healthdata.gov/sites/default/files/reported_hospital_utilization_20210130_2344.csv",
+        "https://healthdata.gov/sites/default/files/reported_hospital_utilization_20210129_1606.csv",
+        "https://healthdata.gov/sites/default/files/reported_hospital_utilization_20210128_2205.csv"
```
chinandrew, line 47: why are there only 3 results here?

tests/acquisition/covid_hosp/state_daily/test_update.py
```diff
@@ -2,27 +2,58 @@
...
+      mock_metadata.assert_called_once()
+      mock_details.assert_called_once()
+      mock_issue.assert_called_with(sentinel.revision)
+      mock_connect.assert_called_once()
+      mock_revisions.assert_called_with(sentinel.last_issue)
+      mock_fetch.assert_called_with(sentinel.url)
+      self.assertEqual(mock_db.insert_metadata.call_args.args[0], sentinel.issue)
+      self.assertEqual(mock_db.insert_metadata.call_args.args[1], sentinel.revision)
+      mock_db.insert_dataset.assert_called_with(sentinel.issue, sample_dataset)
+      self.assertEqual(result, True)
```
chinandrew, line 59: use `self.assertTrue()` here

src/acquisition/covid_hosp/state_daily/update.py
```diff
@@ -21,7 +23,51 @@ def run(network=Network):
       Whether a new dataset was acquired.
     """
 
-    return Utils.update_dataset(Database, network)
+    #return Utils.update_dataset(Database, network)
```
chinandrew, line 26: this just commented out for context with the below documentation?

src/acquisition/covid_hosp/state_daily/update.py
```diff
@@ -21,7 +23,51 @@ def run(network=Network):
...
+      
+      # add metadata to the database
+      metadata_json = json.dumps(metadata)
+      db.insert_metadata(issue, revision, metadata_json)
+
+      urls = network.fetch_revisions(max_issue) + [url]
+      print(f'acquiring {len(urls)} daily updates')
+      n = 0
+      dataset = None
+      for url in urls:
```
chinandrew, line 58: how confident are we that each `new` will have the same index/row order so that `update()` doesn't overwrite the wrong thing?

also as an alternative implementation, could use functools.reduce and do
```
dfs = [network.fetch_dataset(url) for url in urls]
dataset = reduce(lambda x, y: x.update(y),  dfs)
```

src/acquisition/covid_hosp/state_daily/network.py
```diff
@@ -14,3 +20,39 @@ def fetch_metadata(*args, **kwags):
...
+    The first row is different, and links directly back to the main dataset
+    page. Since this row represents the current published version, and the URL
+    for that CSV is included in the metadata, we do not retrieve it here.
+
+    The link for each revision is a date in the same format as
+    `revision_timestamp` from the metadata, so we bookend which links we retrieve
+    by checking against the issue code (YYYYMMDD) for the last issue we
+    imported. 
+
+    This approach will break under the following conditions:
```
chinandrew, line 40: I assume we're ok with these given they're documented here? If we intend on fixing in the future, should create a github issue for it.

tests/acquisition/covid_hosp/state_daily/test_network.py
krivard: The fourth URL is before the "last issue" and thus should be excluded from this batch.

I'll add a comment to clarify.

src/acquisition/covid_hosp/state_daily/update.py
krivard: Yep

src/acquisition/covid_hosp/state_daily/update.py
krivard: it's very likely I've misunderstood how indices work in pandas; let me double check some pessimal cases and get back to you

src/acquisition/covid_hosp/state_daily/network.py
krivard: I am grudgingly okay with it? a robust solution would switch the metadata "have we seen this update before" approach to track the CSV filename instead of revision_timestamp, but that seemed like way more work + greater opportunity to screw up the other covid_hosp pipelines.

We may want to review this with Roni and/or Nick before merging.

src/acquisition/covid_hosp/state_daily/update.py
chinandrew: :+1: i think adding a test case for this logic based on those checks would be a good idea

chinandrew (approved):
:shipit::shipit::shipit:

PR #408, opened 2021-02-10
Update nowcast db connection

chinandrew:
Turns out  the previous context manager using `.begin()` handled the transaction, but not the connection :(. Updated to manage the connection so it doesnt one for each loop.


krivard (approved):
lgtm

PR #411, opened 2021-02-12
Fix epidata requires and bump to 0.0.11

krivard:
ugh commas?


chinandrew:
my bad :(

krivard:
s'okay, I totally skipped testing the installer :(

PR #413, opened 2021-02-15
[POC] COVIDcast union

krivard:
With the limited dataset in place as the `covidcast` table, metadata updates take an hour instead of four, and CSV uploads take 20 minutes instead of 40-60. 

This PR is a POC for keeping the deep 2020 history in the `covidcast2` table, and leaving `covidcast` containing only the most recent issue for all dates plus the data history for dates from Jan 1 2021 onward. This solution would let us keep the fast meta and upload jobs without losing access to the full data history.

The tests include the following:
* Places where the row to be returned is only present in `covidcast`
* Places where the row to be returned is only present in `covidcast2`
* Places where the row to be returned is present in both tables with identical data
* Places where the row to be returned is present in both tables with different timestamps

The latter two in particular already exist in the production database, so it seemed prudent to check them. To handle the differing-timestamps case, rather than wrap everything in another terrible SQL query, I just did a pass through the results in PHP to toss out duplicates, preferring the most recently updated version. Definitely open to suggestions if folks think that won't scale.

UNION is only applied for lag, as-of, and issues queries, since the latest issue for each date is guaranteed to be in `covidcast`.

### DIFF
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -193,6 +193,7 @@ class DelphiEpidataPythonClientTests(unittest.TestCase):
 
     with self.subTest(name='request as-of a date capturing a covidcast-covidcast2 collision'):
       # fetch data, specifying as_of
+      # 20200499 grabs everything before may 1 (doesn't need to be a real date)
       response_1a = Epidata.covidcast(
           'src', 'sig', 'day', 'county', 20200414, '01234',
           as_of=20200499)
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -30,6 +30,7 @@ class DelphiEpidataPythonClientTests(unittest.TestCase):
         database='epidata')
     cur = cnx.cursor()
     cur.execute('truncate table covidcast')
+    cur.execute('truncate table covidcast2')
     cur.execute('truncate table covidcast_nowcast')
     cnx.commit()
     cur.close()
@@ -52,10 +53,20 @@ class DelphiEpidataPythonClientTests(unittest.TestCase):
 
   def test_covidcast(self):
     """Test that the covidcast endpoint returns expected data."""
+    self.maxDiff=None
 
     # insert dummy data
     self.cur.execute('''
       insert into covidcast values
+        (0, 'src', 'sig2', 'day', 'county', 20200414, '01234',
+          123, 1.5, 2.5, 3.5, 456, 4, 20200414, 0, 1, False),
+        (0, 'src', 'sig', 'day', 'county', 20200414, '01234',
+          678, 6.6, 2.2, 11.5, 678, 0, 20200416, 2, 1, False),
+        (0, 'src', 'sig', 'day', 'county', 20200412, '01234',
+          345, 9.5, 2.2, 11.5, 678, 0, 20200415, 3, 1, False)
+    ''')
+    self.cur.execute('''
+      insert into covidcast2 values
         (0, 'src', 'sig', 'day', 'county', 20200414, '01234',
           123, 1.5, 2.5, 3.5, 456, 4, 20200414, 0, 0, False),
         (0, 'src', 'sig2', 'day', 'county', 20200414, '01234',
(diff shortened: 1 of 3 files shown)


chinandrew:
I vaguely recall us discussing having a table for most recent issues and another holding all history back when metadata updates were dragging on, is that worth revisiting as we consider this change? I'm wondering if that would be a cleaner separation of data in the two tables.

krivard:
It would be cleaner, but would require the addition of a widget to move rows from `covidcast` to `covidcast2` whenever a new issue of an existing timepoint was added. by contrast, this PR only modifies the server. if you have bandwidth for taking a shot at such a widget though, go for it!

integrations/client/test_delphi_epidata.py
```diff
@@ -178,6 +191,29 @@ def test_covidcast(self):
         'message': 'success',
       })
 
+    with self.subTest(name='request as-of a date capturing a covidcast-covidcast2 collision'):
+      # fetch data, specifying as_of
+      response_1a = Epidata.covidcast(
+          'src', 'sig', 'day', 'county', 20200414, '01234',
+          as_of=20200499)
```
chinandrew, line 198: is this 04/99 intentional? (so it's in the mega future?)

krivard: Well, you know the rhyme,
```
30 days hath September, June, and November,
All the rest have 31, 
But April's 99
especially in a pandemic year
```

krivard: but really that's just "grab me the most recent issue in april"

chinandrew: lol that would be quite the year. Perhaps leave a comment explaining?

chinandrew:
I haven't examined the PHP closely but looks pretty reasonable to me. I can't imagine this adds much time to the total response time if the second table is relatively right?

src/server/api.php
```diff
@@ -991,12 +992,41 @@ function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $
...
+
+  $query = "{$query} ORDER BY {$order}";
   // get the data from the database
   $epidata = array();
   execute_query($query, $epidata, $fields_string, $fields_int, $fields_float);
+
+  // drop duplicates where some source+signal+geo+date is in both tables with
+  // different timestamps, preferring most recently updated copy
+  if ($union) {
+    $drop = array();
```
sgratzl, line 1018: can you add a comment that this logic requires that the rows are sorted such that the same keys are after each other and that the `value_updated_timestamp` is deceasing.

At the moment it is not clear the rder of the SQL statement and this logic is heavily dependent on each other

src/server/api.php
```diff
@@ -991,12 +992,41 @@ function get_covidcast($source, $signals, $time_type, $geo_type, $time_values, $
...
   }
   // the query
-  $query = "SELECT {$fields} FROM {$table} {$subquery} WHERE {$conditions} AND ({$condition_version}) ORDER BY {$order}";
+  $query_template = "SELECT {$fields} FROM {$table} {$subquery} WHERE {$conditions} AND ({$condition_version})";
+
+  $from_2021 = str_replace("#TABLE#", "covidcast", $query_template);
+  $query = "";
+  if ($union) {
+    $from_2020 = str_replace("#TABLE#", "covidcast2", $query_template);
+    $query = "({$from_2020}) UNION ({$from_2021})";
```
sgratzl, line 1004: have you considered using an `UNION ALL` instead? Are there any duplicates that need to be filtered out?

src/server/api.php
```diff
@@ -2077,7 +2107,7 @@ function get_covidcast_nowcast($source, $signals, $sensor_names, $time_type, $ge
     // get the metadata
     $epidata = get_covidcast_meta();
     store_result($data, $epidata);
-  } else if($endpoint === 'covid_hosp' || $source === 'covid_hosp_state_timeseries') {
+  } else if($endpoint === 'covid_hosp' || $endpoint === 'covid_hosp_state_timeseries') {
```
sgratzl, line 2110: good catch

src/server/api.php
chinandrew: If there are duplicates that need to be filtered out, wouldn't we want to use `UNION`? If there aren't duplicates I guess `UNION ALL` should be a bit faster?

Or maybe theres a case for not filtering out duplicates through SQL that I'm unaware of

krivard: There are two kinds of duplicates: 
* rows with is_latest_issue=1 are identical in both tables, except
* some rows with `issue` between 20210201 and 20210205 have different ids and timestamps but identical values 

If we decide to go with this solution and UNION ALL is faster than UNION, we should definitely consider dropping the covidcast2 version of each duplicate.

krivard:
We discarded this approach as having too high latency for forecast use

PR #414, opened 2021-02-16
COVIDcast: update database definition to match BIGINT fix

### LINKED ISSUE #593: Update covidcast.sql definition with correct `id` column type
`id` is still listed as `int`, which is asking for trouble

krivard:
I believe this is correct but @korlaxxalrok can correct me

Fixes #593


korlaxxalrok (approved):
Yep, that looks correct 👍

PR #416, opened 2021-02-16
Explcitly check for None args

benjaminysmith:
Explcitly check for None args when extracting log file location (since main is called directly with None args for integration tests)


benjaminysmith:
This passes unit and integration tests locally.

krivard (approved):
👍

krivard:
although oddly, CI checks didn't run. this may be another "PR from fork" problem?

PR #417, opened 2021-02-17
NAN coding database and acquisition changes

dshemetov:
This PR begins to address issue #195. Indicator and epidata client PRs to follow. See the [design doc](https://docs.google.com/document/d/1paAYnvHD-swtI4uqjgnE2z7WzU_n-Z0-gN8a2uUU9uA/edit) for more info. The following changes are implemented:
* add 3 columns to the covidcast.ddl for missingness fields
* add acquisition validation for handling the new missing columns, with backwards compatibility for CSVs that do not have these (minimal missing code inference, either NOT_MISSING or UNKNOWN, else error)
* add missing fields to CovidcastRow and RowValues
* allow NULLS in the "val" column in the database
* add integration tests for the new columns
* missingness codes depend on a constants file in the cmu-delphi/utils repo (which is currently manually synced with the file in the indicators repo)

TODO:
- [x] update the Epidata client to request and receive missingness columns
- [x] update the API server to handle requests for and send missingness columns
- [x] add a complete suite of integration tests for missing codes
- [ ] update nancodes with data ranges
- [x] create a survey of the geomaps in the indicators


src/ddl/covidcast.sql
```diff
@@ -30,7 +30,7 @@ Data is public.
 | is_latest_issue              | binary(1)   | NO   |     | NULL    |                |
 | is_wip                       | binary(1)   | YES  |     | NULL    |                |
 | missing_value                | int(11)     | YES  |     | NULL    |                |
-| missing_std                  | int(11)     | YES  |     | NULL    |                |
+| missing_stderr               | int(11)     | YES  |     | NULL    |                |
 | missing_sample_size          | int(11)     | YES  |     | NULL    |                |
```
chinandrew, line 34: there should be `int(1)` based on the new ddl

src/acquisition/covidcast/csv_importer.py
```diff
@@ -31,6 +31,7 @@ class CsvImporter:
 
   # set of required CSV columns
   REQUIRED_COLUMNS = {'geo_id', 'val', 'se', 'sample_size'}
+  # REQUIRED_COLUMNS = {'geo_id', 'val', 'se', 'sample_size', 'missing_val', 'missing_se', 'missing_sample_size'}
```
chinandrew, line 34:
```suggestion
```
if not needed

src/acquisition/covidcast/csv_importer.py
```diff
@@ -271,13 +279,62 @@ def extract_and_check_row(row, geo_type):
 
     # optional not-too-small float
     try:
+      # Returns either a float, None, or causes an exception
       sample_size = CsvImporter.maybe_apply(float, row.sample_size)
     except ValueError:
       # expected a number, but got a string
       return (None, 'sample_size')
 
+    if hasattr(row, "missing_val"):
```
chinandrew, line 288: is hasattr needed if the row is initialized with these values?

src/acquisition/covidcast/csv_importer.py
```diff
@@ -271,13 +279,62 @@ def extract_and_check_row(row, geo_type):
...
+        # if not an int or a nan, then we have problems
+        else:
+          return (None, 'missing_val')
+      # missing code contradicts the value being present
+      elif value is not None and missing_value > 0:
+        return (None, 'missing_val')
+    else:
+      missing_value = 0
+
+    if hasattr(row, "missing_se"):
```
chinandrew, line 304: we should standardize on se vs stderr

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -239,3 +252,52 @@ def test_load_csv_with_valid_header(self):
...
+    geo_type = 'state'
+
+    rows = list(CsvImporter.load_csv(filepath, geo_type, pandas=mock_pandas))
+
+    self.assertTrue(mock_pandas.read_csv.called)
+    self.assertTrue(mock_pandas.read_csv.call_args[0][0], filepath)
+    self.assertEqual(len(rows), 4)
+
+    self.assertEqual(rows[0].geo_value, 'ca')
+    self.assertTrue(rows[0].value is None)
```
chinandrew, line 280: use assertIsNone for these cases

src/acquisition/covidcast/csv_importer.py
dshemetov: This is backwards compatibility thing for CSVs that don't have the `missing` columns. `row` should just be an unprocessed line from a recent `read_csv` call, so it will have whatever the CSV has.

src/acquisition/covidcast/csv_importer.py
dshemetov: yea, so in the CSV files, the standard is `val`, `se`, and `sample_size`. in the database, these get expanded out to `value`, `stderr`, and `sample_size`. the attributes of `RowValues` reflects that. we could change the database to make it more consistent. there would probably be many downstream SQL queries that would need to change then.

src/acquisition/covidcast/csv_importer.py
dshemetov: I suppose there's a question of whether to strictly enforce the presence of missing columns or to impute them if they're missing. I think I'm leaning towards making them required now... this way the NAN coding is contained in one place (mostly), instead of two different codebases

chinandrew: got it, makes sense.

src/acquisition/covidcast/csv_importer.py
chinandrew: I see, didn't realize the db had the stderr for se already. agree that missing_stderr should be consistent

chinandrew:
could you add some integration tests for the acquisition step?

dshemetov:
what do you mean by integration tests? I'm just vaguely familiar with the term. like test communication with a database?

chinandrew:
> what do you mean by integration tests? I'm just vaguely familiar with the term. like test communication with a database?

Yea, see the integrations/ folder for some examples. They rely on a local web server/db and can run the acquisition end to end

src/acquisition/covidcast/csv_importer.py
```diff
@@ -271,13 +278,62 @@ def extract_and_check_row(row, geo_type):
...
+        # if not an int or a nan, then we have problems
+        else:
+          return (None, 'missing_se')
+      # missing code contradicts the value being present
+      if stderr is not None and missing_stderr > 0:
+        return (None, 'missing_se')
+    else:
+      missing_stderr = 0
+
+    if hasattr(row, "missing_sample_size"):
```
benjaminysmith, line 319: It looks like this code is mostly replicated 3 times -- is there a way to extract the common stuff into a function?

integrations/acquisition/covidcast/test_db_connection.py
```diff
@@ -0,0 +1,70 @@
...
+        `is_latest_issue` BINARY(1) NOT NULL,
+        `is_wip` BINARY(1) NOT NULL,
+        `missing_value` int(1) DEFAULT 0,
+        `missing_stderr` int(1) DEFAULT 0,
+        `missing_sample_size` int(1) DEFAULT 0
+      ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+    '''
+    truncate_tmp_table_sql = f'TRUNCATE TABLE {tmp_table_name};'
+    drop_tmp_table_sql = f'DROP TABLE {tmp_table_name}'
+    breakpoint()
```
benjaminysmith, line 48: what is this for?

benjaminysmith (approved):

integrations/acquisition/covidcast/test_db_connection.py
dshemetov: Sorry, I'm going to put this PR into a draft state, as I'm making some new changes to it. This file is my own db integration test before I found that the db setup/teardown code exists, my plan is to absorb it into existing tests.

dshemetov:
Recording some decision logs with @krivard. So there’s a bit of variation in county coverage across the indicators. A majority of the indicators have 3k+ counties, three have ~1500, ~2000, ~2500, and two outliers have ~100 and ~900.
```
Number of counties in signal across all time for latest issue:
{'covid-act-now pcr_specimen_positivity_rate': 3032,
 'chng smoothed_adj_outpatient_covid': 3041,
 'doctor-visits smoothed_adj_cli': 2521,
 'google-symptoms sum_anosmia_ageusia_smoothed_search': 109,
 'hospital-admissions smoothed_covid19_from_claims': 925,
 'quidel covid_ag_smoothed_pct_positive': 2074,
 'safegraph median_home_dwell_time': 3230,
 'fb-survey smoothed_wcli': 1526,
 'jhu-csse confirmed_cumulative_num': 3282,
 'usa-facts confirmed_cumulative_num': 3193}
```

To construct an expected geo map for an indicator, we need to consider the time window to use for it. If we use the full time window, we’ll get everything, but we might get geos that have stopped being reported. While this would add some nans to the database, this doesn’t seem to be a big issue though. As the table below shows, most counties that have reported at least once in the full time window reported at least once in the last month, which suggests that we won't have many empty counties reporting nothing but nans.
```
Fraction of counties in signal present in past month compared to full window:
{'covid-act-now pcr_specimen_positivity_rate': 1.0,
 'chng smoothed_adj_outpatient_covid': 0.9993423216047352,
 'doctor-visits smoothed_adj_cli': 1.0,
 'google-symptoms sum_anosmia_ageusia_smoothed_search': 1.0,
 'hospital-admissions smoothed_covid19_from_claims': 1.0,
 'quidel covid_ag_smoothed_pct_positive': 1.0,
 'safegraph median_home_dwell_time': 0.9996904024767802,
 'fb-survey smoothed_wcli': 1.0,
 'jhu-csse confirmed_cumulative_num': 1.0,
 'usa-facts confirmed_cumulative_num': 1.0}
```

Our current plan is to make a utility that will build custom expected geo maps for some of the indicators. Given the slow rate of change in the geos, we will schedule an update to this list once a month.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -251,33 +299,40 @@ def extract_and_check_row(row, geo_type):
...
-    # optional nonnegative float
-    try:
-      stderr = CsvImporter.maybe_apply(float, row.se)
-    except ValueError:
-      # expected a number, but got a string
+    stderr = CsvImporter.validate_value(row, "se")
+    # stderr was a string or another dtype
+    if stderr == "Error":
       return (None, 'se')
     if stderr is not None and stderr < 0:
```
melange396, line 311: could replace `if` with an `or` but im really picking nits now...

src/acquisition/covidcast/csv_importer.py
```diff
@@ -251,33 +299,40 @@ def extract_and_check_row(row, geo_type):
...
-    # optional not-too-small float
-    try:
-      sample_size = CsvImporter.maybe_apply(float, row.sample_size)
-    except ValueError:
-      # expected a number, but got a string
+    sample_size = CsvImporter.validate_value(row, "sample_size")
+    # sample_size was a string or another dtype
+    if sample_size == "Error":
+      return (None, 'sample_size')
+    if sample_size is not None and sample_size < 0:
```
melange396, line 317: ditto above comment ^

src/acquisition/covidcast/csv_importer.py
```diff
@@ -188,10 +192,54 @@ def floaty_int(value):
...
+      # since that will be filtered in the export_to_csv util in 
+      # covidcast-indicators; nonetheless this code is here for safety
+      if attr_value is not None and missing_entry > 0:
+        return None
+      return missing_entry
+    else:
+      if attr_value is not None:
+        return Nans.NOT_MISSING.value
+      else:
+        return Nans.UNKNOWN.value
```
melange396, line 241:
```suggestion
      if attr_value not None:
        return Nans.UNKNOWN.value
      return Nans.NOT_MISSING.value
```

src/acquisition/covidcast/csv_importer.py
```diff
@@ -188,10 +192,54 @@ def floaty_int(value):
...
+    if hasattr(row, "missing_" + attr_name):
+      missing_entry = getattr(row, "missing_" + attr_name)
+      try:
+        missing_entry = int(missing_entry)
+      except ValueError:
+        return None
+      # A missing code should never contradict the value being present, 
+      # since that will be filtered in the export_to_csv util in 
+      # covidcast-indicators; nonetheless this code is here for safety
+      if attr_value is not None and missing_entry > 0:
```
melange396, line 234:
```suggestion
      if attr_value is not None and missing_entry != Nans.NOT_MISSING.value:
```

src/acquisition/covidcast/csv_importer.py
melange396: (slightly simpler)

src/acquisition/covidcast/csv_importer.py
melange396: (just to tie it back to the enum and not use a dangling int literal)

melange396 (approved):
i found a couple nits, but otherwise LGTM

src/acquisition/covidcast/csv_importer.py
```diff
@@ -188,10 +192,54 @@ def floaty_int(value):
...
-    if str(value).lower() not in ('', 'na', 'nan', 'inf', '-inf', 'none'):
+    if str(value).lower() in ('inf', '-inf'):
+      raise ValueError
+    elif str(value).lower() in ('', 'na', 'nan', 'none'):
+      return None
+    else:
       return func(value)
 
+  @staticmethod
+  def validate_value(row, attr_value):
```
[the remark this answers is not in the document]
dshemetov: Good point, I'm renaming value to quantity in a few more functions.

src/acquisition/covidcast/csv_importer.py
dshemetov: Combined these and made comments to make clear what cases are being captured in each.

dshemetov:
Ooof, let's wait to merge this until next week, I found a couple things I'd like to fix.

melange396 (approved):
good stuff

dshemetov:
Notes on rollout steps:
1. alter database: add [three missing columns](https://github.com/cmu-delphi/delphi-epidata/blob/a1538f6a1aadaf728d7aab010dfbee2e969ffbb7/src/ddl/covidcast.sql#L104) and [allow null in the value column](https://github.com/cmu-delphi/delphi-epidata/blob/a1538f6a1aadaf728d7aab010dfbee2e969ffbb7/src/ddl/covidcast.sql#L95)
2. merge in these db changes
3. merge in indicator changes

krivard (changes requested):
It looks like this PR does not include any changes to the API documentation, however it does add three columns to the query results output. Please [document these appropriately](https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/api/covidcast.md) so users know what to expect.

@chinandrew will these additional columns crash the covidcast library? If so we should release new versions which will accept the change and announce the new release to the mailing list before we break everybody's research.

Even if this is not a breaking change for the covidcast library, we still need to write an announcement of the new nan features before merging. @dshemetov I'll link you to the mail drafts doc offline to get started on that.

chinandrew:
> @chinandrew will these additional columns crash the covidcast library? If so we should release new versions which will accept the change and announce the new release to the mailing list before we break everybody's research.

Python one should be fine since it just takes the JSON and converts to dataframe, and as long as we don't remove/rename existing columns we shouldn't run into any issues with the column-specific operations.  We will want to update the documentation, and perhaps make a change to adjust the column order so the new columns are the the end. Not sure about R package, will have to look into that more

chinandrew:
R package seems to behave like the Python one, it just takes a CSV and converts to dataframe without caring what the new colums are.

dshemetov:
@krivard I added some API documentation.

docs/api/covidcast.md
```diff
@@ -167,6 +167,9 @@ require knowing when we last confirmed an unchanged value, please get in touch.
 | `epidata[].stderr` | approximate standard error of the statistic with respect to its sampling distribution, `null` when not applicable | float |
 | `epidata[].sample_size` | number of "data points" used in computing the statistic, `null` when not applicable | float |
 | `epidata[].issue` | time unit (e.g. date) when this statistic was published | integer |
+| `epidata[].missing_value` | an integer code that is zero when the `value` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
+| `epidata[].missing_stderr` | an integer code that is zero when the `stderr` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
+| `epidata[].missing_sample_size` | an integer code that is zero when the `sample_size` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
 | `epidata[].lag` | time delta (e.g. days) between when the underlying events happened and when this statistic was published | integer |
```
krivard, line 173: Keep `issue` and `lag` together

```suggestion
| `epidata[].issue` | time unit (e.g. date) when this statistic was published | integer |
| `epidata[].lag` | time delta (e.g. days) between when the underlying events happened and when this statistic was published | integer |
| `epidata[].missing_value` | an integer code that is zero when the `value` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
| `epidata[].missing_stderr` | an integer code that is zero when the `stderr` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
| `epidata[].missing_sample_size` | an integer code that is zero when the `sample_size` field is present and non-zero when the data is missing (see [missing codes](missing_codes.md)) | integer |
```

docs/api/missing_codes.md
```diff
@@ -0,0 +1,23 @@
+---
+title: NaN Missing Codes
+parent: COVIDcast Epidata API
+nav_order: 5
+---
+
+# Missing Value Coding
+
+Occasionally, data will be missing from our database and will be explicitly coded as a NaN.
```
krivard, line 9:
```suggestion
Occasionally, data will be missing from our database and will be explicitly coded as NaN.
```

docs/api/missing_codes.md
```diff
@@ -0,0 +1,23 @@
...
+In these cases, we strive to supply our best-known reason for the value to be missing by
+providing an integer code in the corresponding `missing_` column (i.e. `missing_value`
+corresponds to the `value` column). The integer codes are as follows
+
+| Code | Name | Description |
+| --- | --- | --- |
+| 0 | DEFAULT | This is the default value for when the entry is not missing. |
+| 1 | NOT APPLICABLE | This value is used when the field is not expected to have a value (e.g. stderr for a signal without a sample size). |
```
krivard, line 17: (sample size is its own column)
```suggestion
| 1 | NOT APPLICABLE | This value is used when the field is not expected to have a value (e.g. stderr for a signal that is not estimated from a sample). |
```

docs/api/missing_codes.md
```diff
@@ -0,0 +1,23 @@
...
+| 2 | REGION EXCEPTION | This value is used when the field is not reported because the particular indicator does not serve the geographical region requested. |
+| 3 | PRIVACY | This value is used when the field has been censored for data privacy reasons. This could be due to reasons such as low sample sizes or simply a requirement from our data partners. |
+| 4 | DELETED | This value is used when the field was present in previous issues, but is no longer reported. Deletions can arise due to reasons such as bug fixes or changing censorship requirements. |
```
krivard, line 20:
```suggestion
| 4 | DELETED | This value is used when the field was present in previous issues, but is no longer reported. Deletions can arise due to bug fixes, changing censorship requirements, or data corrections from the source. |
```

docs/api/missing_codes.md
```diff
@@ -0,0 +1,23 @@
...
+# Missing Value Coding
+
+Occasionally, data will be missing from our database and will be explicitly coded as a NaN.
+In these cases, we strive to supply our best-known reason for the value to be missing by
+providing an integer code in the corresponding `missing_` column (i.e. `missing_value`
+corresponds to the `value` column). The integer codes are as follows
+
+| Code | Name | Description |
+| --- | --- | --- |
+| 0 | DEFAULT | This is the default value for when the entry is not missing. |
```
krivard, line 16: consistent terminology
```suggestion
| 0 | DEFAULT | This is the default value for when the field is not missing. |
```

src/acquisition/covidcast/nancodes.py
```diff
@@ -0,0 +1,19 @@
+"""Provides unified not-a-number codes for the indicators.
+
+Currently requires a manual sync between the covidcast-indicators
+and the delphi-epidata repo.
+* in covidcast-indicators: _delphi_utils_python/delphi_utils
+* in delphi-epidata: src/acquisition/covidcast
```
krivard, line 6: is there a reason we're not importing these from delphi_utils directly?

krivard (changes requested):
Great work! Minor nits.

also forgive me if i missed context on this, but wasn't the point of publishing delphi_utils on pypi so that we could avoid needing to duplicate the nan codes in acquisition.covidcast?

src/acquisition/covidcast/nancodes.py
dshemetov: there is a reason, but it is not a good one (i forgot!) 

fixing now (thanks for catching!).

dshemetov:
If our infra uses Docker, then a [requirements file](https://github.com/cmu-delphi/operations/pull/62) needs to be updated.

krivard (approved):
xlnt

dshemetov:
A few things coming up while merging Flask server and working with Docker:
* python requirements need to be updated: python-dotenv, flask, orjson, sqlalchemy==1.3.24
* the last one because the Flask server currently depends on a <1.4 version of SqlAlchemy because of [RowProxy](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#rowproxy-is-no-longer-a-proxy-is-now-called-row-and-behaves-like-an-enhanced-named-tuple)

I'm getting a few failing tests due to databases not existing at the moment.

krivard:
Can you paste the exact errors in thread? CI doesn't run on forks so we can't see what you're seeing.

dshemetov:
Once I installed the [right dependencies](https://github.com/cmu-delphi/operations/pull/64) and added the flag 
`--env SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata`
to the Docker run string, the tests that were failing now pass with `pytest`, but they fail when running with `py3tester`. A lot of integrations tests are also failing.

dshemetov:
Once I updated the `delphi_web_epidata` container correctly, all the tests pass now.

dshemetov:
For future ref, [this link to the CI recipe](https://github.com/cmu-delphi/delphi-epidata/blob/76cc4c513fe1fa64eede08a6a9202aaa25e0dc1b/.github/workflows/ci.yaml#L59) from @sgratzl was really helpful. Going to add that to the ops docs.

krivard:
🙏  thank you for both going on that goose chase and ensuring future-us won't have to chase that particular goose again!

If you change this to merging into `dev` that will make it easier to deploy with the new release system Monday.

dshemetov:
We should incorporate [these](https://github.com/cmu-delphi/covidcast-indicators/pull/1044) naming changes to `delphi_utils` before we release on Monday.

krivard:
https://pypi.org/project/delphi-utils/0.1.1/ is up; merging

PR #415, opened 2021-02-16
Add logging to CSV ingestion

benjaminysmith:
Adds logging to CSV ingestion.

Progress on #403

PR #418, opened 2021-02-18
Retrieve the modified DB row count using cursor.rowcount

benjaminysmith:
Previously this always returned None

Improves #415

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -212,17 +212,18 @@ class Database:
         ) for row in cc_rows[start:end]]
 
 
-        result = self._cursor.executemany(insert_into_tmp_sql, args)
+        self._cursor.executemany(insert_into_tmp_sql, args)
         self._cursor.execute(insert_or_update_sql)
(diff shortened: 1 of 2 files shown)


benjaminysmith:
Verified this passes unit and integration tests locally.

krivard:
This line is super expensive; does this PR mean we can drop it?
https://github.com/cmu-delphi/delphi-epidata/blob/b512b6b55885cdd5c2485414595184ba259b833e/src/acquisition/covidcast/csv_to_database.py#L196

benjaminysmith:
> This line is super expensive; does this PR mean we can drop it?
> 
> https://github.com/cmu-delphi/delphi-epidata/blob/b512b6b55885cdd5c2485414595184ba259b833e/src/acquisition/covidcast/csv_to_database.py#L196

I believe so, but suggest we merge this version and compare the output first (i.e., the logging of that statement vs. the sum of the new row count) -- given that the python SQL row count behavior is kind of waffly. If it looks good in prod we can switch this line for a sum of the individual counts. wdyt?

krivard (approved):
looks good!

PR #420, opened 2021-02-18
Remove unsed stdev function

chinandrew:
Removing an old function. See [this slack thread](https://delphi-org.slack.com/archives/C013AH5N01E/p1613671692177000).

### DIFF
--- a/tests/acquisition/covidcast/test_database.py
+++ b/tests/acquisition/covidcast/test_database.py
@@ -73,32 +73,6 @@ class UnitTests(unittest.TestCase):
     self.assertIn('select count(1)', sql)
     self.assertIn('from `covidcast`', sql)
 
-  def test_get_data_stdev_across_locations_query(self):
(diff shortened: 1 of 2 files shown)


tests/acquisition/covidcast/test_database.py
```diff
@@ -73,32 +73,6 @@ def test_count_all_rows_query(self):
     self.assertIn('select count(1)', sql)
     self.assertIn('from `covidcast`', sql)
 
-  def test_get_data_stdev_across_locations_query(self):
-    """Query to get signal-level standard deviation looks sensible.
-
-    NOTE: Actual behavior is tested by integration test.
```
krivard, line 79: Was this orphaned or is there a matching integration test which should also be removed/modified?

chinandrew: I wasn't able to find any other references in the repo

krivard (approved):

PR #421, opened 2021-02-18
Fix import for state daily integration test

chinandrew:
Some of the state_daily update PRs were opened before #371  and merged after, so this pytest-required import got missed


krivard (approved):
oooo thanks!

PR #422, opened 2021-02-19
API documentation for CAN indicator

eujing:


docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,106 @@
...
+
+$$
+\begin{aligned}
+    N_{Gt} &= \sum_{i \in G} w_i N_{it} \\
+    \hat{p}_{Gt} &= \frac{ \sum_{i \in G} w_i \hat{Y}_{it} }{ N_{Gt} }
+\end{aligned}
+$$
+
+Note that in order to more closely mirror the CAN data source, no pseudo-observations and priors 
+are used in the estimation of these proportions.
```
krivard, line 64:
```suggestion
Note that in order to more closely mirror the CAN data source, no pseudo-observations
(such as we would typically apply for a Jeffreys prior)
are used in the estimation of these proportions. Instead of using the Jeffreys prior to
more accurately estimate the standard error when the value is zero, we suppress all
entries with a value of zero.
```

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,106 @@
...
+Estimates for geographical levels beyond counties may be inaccurate due to how aggregations 
+are done on smoothed values instead of the raw values. Ideally we would aggregate raw values 
+then smooth, but the raw values are not accessible in this case.
+
+The positivity rate here should not be interpreted as the population positivity rate as 
+the testing performed are typically not randomly sampled, especially for early data 
+with lower testing volumes.
+
+A few counties, most notably in California, are also not covered by this data source.
+
```
krivard, line 93: A line here repeating that we suppress zeros

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,106 @@
...
+## Lag and Backfill
+
+The lag for these signals vary depending on the reporting patterns of individual counties.
```
krivard, line 96:
```suggestion
The lag for these signals varies depending on the reporting patterns of individual counties.
```

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,106 @@
...
+Most counties have their latest data report with a lag of 2 days, while others can take 9 days 
+or more in the case of California counties.
+
+These signals are also backfilled as backlogged test results could get assigned to older 7-day timeframes.
```
krivard, line 100: Do we know how much backfill there is?

* How many days are submitted in each issue?
* What's the median change from the first day a value is issued to its final update?

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,106 @@
...
+\begin{aligned}
+    \hat{se}(p_{it}) &= \sqrt{\frac{p_{it} (1 - p_{it})}{N_{it}}} \\
+    \hat{se}(\hat{p}_{Gt}) &= \sqrt{\frac{\hat{p}_{Gt} (1 - \hat{p}_{Gt})}{N_{Gt}}}
+\end{aligned}
+$$
+
+### Smoothing
+
+No additional smoothing is done to avoid double-smoothing, since the data pulled from CAN / CDC 
+is already smoothed as a 7-day-average.
```
krivard, line 80:
```suggestion
is already smoothed with a 7-day-average.
```

krivard (changes requested):
Excellent work. I added some more detail about zero suppression. Check it for accuracy, then add a short mention of zero suppression to the limitations section. I'd also like some more detailed stats on backfill.

docs/api/covidcast-signals/covid-act-now.md
eujing: Unfortunately I have only pulled a few non-contiguous days of CAN data on my side to have accurate statistics about issue dates.
Will go through them to try get a rough idea, but will also spin up a cron job in the meantime to collect it for a few days.

eujing: Ooh ok CAN commits their [parquet files into github](https://github.com/covid-projections/covid-data-public/commits/main/data/can-scrape), so I can go through the commit history to get the statistics.

eujing: Regarding how many days are submitted in each issue, we get data from the very start (Mar 2020) every time we pull from CAN, but rely on the archive differ to figure out the differences.
Should that be mentioned in this section too?

docs/api/covidcast-signals/covid-act-now.md
eujing: Thanks for the detail! I added on a short bit to clarify that we suppress entries with zero sample size.

docs/api/covidcast-signals/covid-act-now.md
krivard: No, the description of diff-based issues is covered on the versioning page. This is just to get an idea of the scale of the average revision. For example, both doctor-visits and covid antigen tests update 60 days in the past, every day, while ght only ever updates 1 day, and cases/deaths typically updates 1 day but occasionally updates much larger spans, sometimes 300 days or more. Your paragraph below is sufficient -- nice work!

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -91,13 +94,20 @@ with lower testing volumes.
...
-The lag for these signals vary depending on the reporting patterns of individual counties.
+The lag for these signals varies depending on the reporting patterns of individual counties.
 Most counties have their latest data report with a lag of 2 days, while others can take 9 days 
 or more in the case of California counties.
 
 These signals are also backfilled as backlogged test results could get assigned to older 7-day timeframes.
+Most recent test positivity rates do not change substantially with backfill (having a median delta of close to 0).
+However, most recent total tests performed is expected to increase in later data revisions (having a median increase of 7%).
+Values more than 5 days in the past are expected to remain fairly static, with total tests performed 
+having a median increase of 1% of less), as most major revisions have already occurred.
```
krivard, line 110:
```suggestion
Values more than 5 days in the past are expected to remain fairly static (with total tests performed 
having a median increase of 1% of less), as most major revisions have already occurred.
```

krivard (approved):
👍

docs/api/covidcast-signals/covid-act-now.md
```diff
@@ -0,0 +1,116 @@
+---
+title: COVID Act Now
+parent: Data Sources and Signals
+grand_parent: COVIDcast Epidata API
+---
+
+# COVID Act Now (CAN)
+{: .no_toc}
+
+* **Source name:** `covid-act-now`
```
chinandrew, line 10: FYI the data folder on staging has underscores instead of dashes

krivard: Fixed in https://github.com/cmu-delphi/covidcast-indicators/pull/916

PR #428, opened 2021-02-25
Document new V5abc signals (vaccine hesitancy reasons)

nmdefries:


docs/api/covidcast-signals/fb-survey.md
```diff
@@ -334,6 +334,21 @@ also available. These have names beginning `smoothed_w`, such as
...
 | `smoothed_vaccine_likely_politicians` | Estimated percentage of respondents who would be more likely to get a COVID-19 vaccine if it were recommended to them by politicians, among respondents who have not yet been vaccinated. <br/> **Earliest date available:** 2021-01-20 | V4 |
 | `smoothed_received_2_vaccine_doses` | Estimated percentage of respondents who have received two doses of a COVID-19 vaccine, among respondents who have received either one or two doses of a COVID-19 vaccine. This item was shown to respondents starting in Wave 7. <br/> **Earliest date available:** 2021-02-06 | V2 |
+| `smoothed_hesitancy_reason_sideeffects` | Estimated percentage of respondents who say they are hesitant to get vaccinated because they are worried about side effects, among respondents who selected any response on item V5a, V5b, or V5c (displayed if responded "Yes, probably", "No, probably not", or "No, definitely not" to item V3). This series of items was shown to respondents starting in Wave 8. <br/> **Earliest date available:** 2021-02-26 | V5abc |
```
capnrefsmmat, line 337:
```suggestion
| `smoothed_hesitancy_reason_sideeffects` | Estimated percentage of respondents who say they are hesitant to get vaccinated because they are worried about side effects, among respondents who answered "Yes, probably", "No, probably not", or "No, definitely not" when asked if they would get vaccinated if offered (item V3). This series of items was shown to respondents starting in Wave 8. <br/> **Earliest date available:** 2021-02-26 | V5abc |
```
This simplifies the language a bit and makes it so you don't have to look up the survey instrument to understand it; can we apply this to the others as well?

capnrefsmmat (changes requested):
One suggested change for clarity (to be applied to them all).

Also, this table is going to get big. Can we break it up into three tables with sub-headings? I'm thinking:

* Vaccine Uptake and Acceptance
* Reasons for Hesitancy (would include these and `smoothed_worried_vaccine_side_effects`)
* Outreach and Image (including all the `smoothed_vaccine_likely*` signals)

capnrefsmmat (approved):
shiiiny

PR #432, opened 2021-03-03
Have async raise 4xx/5xx status codes and add connection limit

chinandrew:
- Added raise_for_status so 4xx and 5xx codes will raise properly
- Properly implemented simultaneous connection limit (interestingly, the default is 100)
- Lowered the limit to 50

### DIFF
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -5,7 +5,9 @@ import unittest
 from unittest.mock import patch, MagicMock
 
 # third party
+from aiohttp.client_exceptions import ClientResponseError
 import mysql.connector
+import pytest
 
 # first party
 from delphi.epidata.client.delphi_epidata import Epidata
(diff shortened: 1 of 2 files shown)


krivard (commented):
Is this change testable?

chinandrew:
> Is this change testable?

I'll add tests for the status codes being raised. The connections....might be difficult.

krivard:
The three things I was hoping for are
* status codes (excellent)
* no more than 50 connections (difficult; that's fair, can skip for now)
* properly reconstructs results when more than 50 params items are requested (doable? i'd be satisfied with a one-off manual test if doing it up as a unit test is difficult)

chinandrew:
> * status codes (excellent)

Added. Not sure what the best way to change the endpoint to something invalid is, so added a decorator to change the URL within the function. The function uses the class and not the instance attribute for the URL so i can't just make an Epidata instance with a bad URL, and mocking aiohttp responses also gets a bit wonky.

> * no more than 50 connections (difficult; that's fair, can skip for now)

Skipped for now

> * properly reconstructs results when more than 50 params items are requested (doable? i'd be satisfied with a one-off manual test if doing it up as a unit test is difficult)

The [current integration test](https://github.com/cmu-delphi/delphi-epidata/blob/0e59e6b74c627705e3243ef3aaff7911d6ff6ba9/integrations/client/test_delphi_epidata.py#L514) makes 24 calls with a limit of 10.

krivard:
uhhh unfortunately it looks like what that actually does is:
1. use async_epidata to fetch 2 params, in batches of 10 (so batching is never actually engaged)
2. duplicate the response 12 times
3. check against the expected result for those two params, duplicated 12 times

I think what you meant to do there is
1. use async_epidata to fetch 2 params, duplicated 12 times
2. extract the response
3. check against the expected result for those two params, duplicated 12 times

chinandrew:
> uhhh unfortunately it looks like what that actually does is:
> 
> 1. use async_epidata to fetch 2 params, in batches of 10 (so batching is never actually engaged)
> 2. duplicate the response 12 times
> 3. check against the expected result for those two params, duplicated 12 times
> 
> https://github.com/cmu-delphi/delphi-epidata/blob/0e59e6b74c627705e3243ef3aaff7911d6ff6ba9/integrations/client/test_delphi_epidata.py#L493-L518
> 
> I think what you meant to do there is
> 
> 1. use async_epidata to fetch 2 params, duplicated 12 times
> 2. extract the response
> 3. check against the expected result for those two params, duplicated 12 times

:facepalm: my bad, should be fixed now.

krivard (approved):
xlnt

PR #437, opened 2021-03-05
bump delphi-epidata version to 0.0.12

krivard:
This will let us release correct handling for 5XX and 4XX response codes and more polite batching.


chinandrew (approved):
:shipit::shipit::shipit:

PR #440, opened 2021-03-09
Update survey docs with more Wave 10 indicators

nmdefries:
This describes the 7-day versions of mental health indicators and state travel indicator, discontinuing the previous 5-day versions, and the new "indoors" versions of C13 (social activities and activities outside the home), discontinuing the previous not-explicitly-indoors indicators.


capnrefsmmat (approved):
Looks good; I may make a PR later to reorganize these a bit, just since we're getting so many signals

PR #441, opened 2021-03-10
Move GHT docs to Inactive Signals

krivard:
Requests to the GHT API now generate the following error:

```
googleapiclient.errors.UnknownApiNameOrVersion: name: trends  version: v1beta
```

which means it's time to deactivate the `ght` COVIDcast source.

### DIFF
--- a/docs/api/covidcast-signals/ght.md
+++ b/docs/api/covidcast-signals/ght.md
@@ -22,7 +22,7 @@ searches in a given location, on a given day. This signal is measured in
 arbitrary units (its scale is meaningless); larger numbers represent higher
 numbers of COVID-related searches.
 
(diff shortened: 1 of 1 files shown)


docs/api/covidcast-signals/ght.md
```diff
@@ -22,6 +22,11 @@ searches in a given location, on a given day. This signal is measured in
 arbitrary units (its scale is meaningless); larger numbers represent higher
 numbers of COVID-related searches.
 
+These signals were updated daily until May 8, 2021. After that date, Google
```
chinandrew, line 25:
```suggestion
These signals were updated daily until March 8, 2021. After that date, Google
```

PR #449, opened 2021-03-16
Update Covid hosp to use new healthdata.gov API

### LINKED ISSUE #446: Update all COVID Hospitalization code to use new healthdata.gov API
Thanks to @ftrotter for #445, which will help us limp along until we switch to the new API.

This also requires us to update our documentation, since a lot of the links have been moved without redirects (grn).

New healthdata.gov locations:
* [COVID-19 Reported Patient Impact and Hospital Capacity by State Timeseries](https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/g62h-syeh)
* [COVID-19 Reported Patient Impact and Hospital Capacity by Facility](https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/anag-cw7u)

chinandrew:
Fixes #446.

Changes covid_hosp to use the new API. Main changes revolve around updating the metadata/revision acquisition, which is now it's own dataset with a CSV download URL. This means that we can fetch the dataset with a simple pd.read_csv and throw away the web scraping code.

This also updates the dataset IDs to use their new ones.

@krivard, I saw NCHS uses the same Socrata API and takes a token, so we may want to do that here as well (either with the same or a new token). It works without one, but they recommend it. 

In terms of the switchover once this is merged, I could see one anticipated error case where the following line doesn't evaluate to True when it should, since the new revisions look differently. This would lead to two copies of the revision being ingested. We may be able to workaround by adding a date check on top of the revision check so if the latest issue = the max issue in the db and the date < the day we merge, we don't ingest (this fails if it was actually a new revision on that exact date, though that seems less likely).
```
      # bail if the dataset has already been acquired
      if db.contains_revision(revision):
        print('already have this revision, nothing to do')
        return False
```

Leaving in draft mode while we work this out.


src/acquisition/covid_hosp/common/network.py
```diff
@@ -7,9 +7,9 @@ class Network:
 # This code is a temporary (two weeks or so) fix until the legacy API is fully retired. Code using this style of URL will be phased out entirely then
 # This needs to migrate to the new healthdata.gov API, which is different
```
krivard, line 8:
```suggestion
```

src/acquisition/covid_hosp/common/utils.py
```diff
@@ -178,8 +85,8 @@ def update_dataset(database, network):
 
     # get dataset details from metadata
     metadata = network.fetch_metadata()
-    url, revision = Utils.extract_resource_details(metadata)
-    issue = Utils.get_issue_from_revision(revision)
+    issue = max(metadata.index)
```
krivard, line 88: What will happen here if HHS has posted more than one update since the last time we ran this code?

testdata/acquisition/covid_hosp/state_daily/metadata.csv
```diff
@@ -0,0 +1,9 @@
+Update Date,Days Since Update,User,Rows,Row Change,Columns,Column Change,Metadata Published,Metadata Updates,Column Level Metadata,Column Level Metadata Updates,Archive Link
```
krivard, line 1: This appears to be the facility metadata, not the state metadata.

testdata/acquisition/covid_hosp/state_timeseries/metadata.csv
```diff
@@ -0,0 +1,9 @@
+Update Date,Days Since Update,User,Rows,Row Change,Columns,Column Change,Metadata Published,Metadata Updates,Column Level Metadata,Column Level Metadata Updates,Archive Link
```
krivard, line 1: This also appears to be for facility and not state

tests/acquisition/covid_hosp/common/test_utils.py
```diff
@@ -67,75 +67,6 @@ def test_parse_bool(self):
...
-      url, revision = Utils.extract_resource_details(metadata)
-
-      expected_url = (
-        'https://healthdata.gov/sites/default/files/'
-        'estimated_inpatient_all_20201213_1757.csv'
-      )
-      self.assertEqual(url, expected_url)
-      self.assertEqual(revision, 'Sun, 12/13/2020 - 22:36')
-
   def test_run_skip_old_dataset(self):
```
krivard, line 70: Add a test here that checks for collisions between keys generated by the old API and the new API

krivard (changes requested):
I think common.utils will wind up skipping some issues.

we should also be able to test for correct handling of comparisons between the old and new keys

testdata/acquisition/covid_hosp/state_daily/metadata.csv
chinandrew: replaced with state metadata

testdata/acquisition/covid_hosp/state_timeseries/metadata.csv
chinandrew: replaced with state timeseries metadata

src/acquisition/covid_hosp/common/utils.py
chinandrew: For state timeseries and facility, it gets only the most recent one (the previous code with extract_resource_details seems to do the same, but perhaps I misunderstood the metadata values).

For the state_daily, it should grab all issues after the highest issue in the db and submit them as one issue dated as the latest issue.

krivard: the previous API didn't give us a choice  -- it could only ever yield the metadata for the most recent issue.

Since the new API _does_ expose all issues, we should make sure we don't skip any.

chinandrew:
Alright made a couple bigger changes in this PR. Now all of facility, state daily, and state timeseries share the same logic for acquiring data which is to 
1. download metadata csv
2. grab the highest issue already acquired (based on the covid_hosp_metadata table)
3. select all issues for days after that issue and before the current day. issues after the highest issue but on the same day as that issue are not acquired. This choice I think is open to debate, though handling new issues on past ingested dates may require some effort.
4. For each day, roll all its issues up into a single dataset insertion and upload that. Upload the last issue that date into the metadata.

src/acquisition/covid_hosp/common/database.py
```diff
@@ -168,3 +169,22 @@ def insert_dataset(self, publication_date, dataframe):
                        id_and_publication_date +
                        tuple(values) +
                        tuple(i[0] for i in self.additional_fields))
+
+  def get_max_issue(self):
+    """Fetch the most recent state-daily issue.
```
krivard, line 174:
```suggestion
    """Fetch the most recent issue.
```

src/acquisition/covid_hosp/common/database.py
```diff
@@ -168,3 +169,22 @@ def insert_dataset(self, publication_date, dataframe):
...
+
+    This is used to bookend what links we scrape from the HHS revisions page.
```
krivard, line 176:
```suggestion
    This is used to bookend what updates we pull in from the HHS metadata.
```

src/acquisition/covid_hosp/common/utils.py
```diff
@@ -67,6 +68,61 @@ def parse_bool(value):
...
     raise CovidHospException(f'cannot convert "{value}" to bool')
 
+  def issues_to_fetch(metadata, newer_than, older_than):
+    """
+    Construct all issue dates and URLs to be ingested based on metadata.
+
+    Parameters
+    ----------
+    newer_than Date
+      Lower bound (not inclusive) of days to get issues for.
```
krivard, line 78:
```suggestion
    metadata pd.DataFrame
      HHS metadata indexed by issue date and with column "Archive Link"
    newer_than Date
      Lower bound (exclusive) of days to get issues for.
    older_than Date
      Upper bound (exclusive) of days to get issues for
```

src/acquisition/covid_hosp/common/utils.py
```diff
@@ -67,6 +68,61 @@ def parse_bool(value):
...
+    Returns
+    -------
+      Dictionary of {issue day: list of download urls} for issues after newer_than
```
krivard, line 81:
```suggestion
      Dictionary of {issue day: list of download urls} for issues after newer_than and before older_than
```

src/acquisition/covid_hosp/common/utils.py
```diff
@@ -82,32 +138,27 @@ def update_dataset(database, network):
...
-      db.insert_dataset(issue_int, dataset)
-
-      print(f'successfully acquired {len(dataset)} rows')
+      for issue, revisions in daily_issues.items():
+        issue_int = int(issue.strftime("%Y%m%d"))
+        # download the dataset and add it to the database
+        dataset = Utils.merge_by_key_cols([network.fetch_dataset(url) for url, _ in revisions],
+                                          db.KEY_COLS)
+        db.insert_dataset(issue_int, dataset)
+        # add metadata to the database using the last revision seen.
```
krivard, line 156: 👍

krivard (approved):
nice work! some docs to fix but otherwise yer good

krivard:
What's remaining for this to come out of draft? 

iirc the rollout plan is a two-day sequence:

* reschedule Automation tasks to skip a day
* merge
* monitor the first active day of the new code to make sure we didn't duplicate anything

if we're ready, I can set us up to skip Monday's run and monitor Tuesday's.

chinandrew:
> What's remaining for this to come out of draft?
> 
> iirc the rollout plan is a two-day sequence:
> 
> * reschedule Automation tasks to skip a day
> * merge
> * monitor the first active day of the new code to make sure we didn't duplicate anything
> 
> if we're ready, I can set us up to skip Monday's run and monitor Tuesday's.

API token? otherwise should be good to go

krivard:
Oh right. Let's go without for now, and I'll make an issue for adding it once the infra inferno of bigint+flask+docker has calmed down. delphi-epidata doesn't have as nice a mechanism for storing secrets as covidcast-indicators does.

PR #457, opened 2021-03-19
update to latest main and add signal-dashboard endpoints

sgratzl:


sgratzl:
@krivard the only relevant new changes are the two files in  src/server/endpoints. The rest is just a sync

krivard:
cc @benjaminysmith 

but also @sgratzl see switch to counts in https://github.com/cmu-delphi/delphi-epidata/pull/458

you two should combine your efforts

sgratzl:
> but also @sgratzl see switch to counts in #458

done thx for checking this

benjaminysmith:
Thanks Sam! I think it's fine if we get out of sync with API changes for now given that we are rapidly iterating. I think a good first step is getting the Flask API for the dash endpoint up in some form, and then we can sync stuff over.

krivard (approved):
👍

krivard:
@sgratzl I'm good to merge this now unless you want to keep accumulating?

Issue #438, opened 2021-03-09
Create API for signal dashboard


benjaminysmith:

sgratzl:
can we close this?

krivard:
Sure -- further work can be individually tracked.

PR #458, opened 2021-03-19
Switch coverage to use counts

benjaminysmith:
- Update coverage generation to count all geo_types
- Change DB schema to persist counts instead of individual geos
- Update API to return counts
- Update tests
- Also fix column format for persisted dates

Iterates on #438

### DIFF
--- a/src/acquisition/covidcast/signal_dash_data_generator.py
+++ b/src/acquisition/covidcast/signal_dash_data_generator.py
@@ -15,6 +15,9 @@ from typing import List
 import delphi.operations.secrets as secrets
 from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
 from delphi.epidata.client.delphi_epidata import Epidata
+# import delphi.operations.secrets as secrets
+# from logger import get_structured_logger
+# import covidcast
 
 
 @dataclass
@@ -35,7 +38,7 @@ class DashboardSignalCoverage:
     signal_id: int
     date: datetime.date
     geo_type: str
-    geo_value: str
+    count: int
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


src/acquisition/covidcast/signal_dash_data_generator.py
```diff
@@ -192,16 +188,20 @@ def get_coverage(dashboard_signal: DashboardSignal,
         signal,
         end_day=latest_time_value,
         start_day=latest_time_value)
+    count_by_geo_type_df = latest_data.groupby(
+        ['geo_type', 'data_source', 'time_value', 'signal']).size().to_frame('count').reset_index()
+
+    if len(count_by_geo_type_df) > 1:
+        raise ValueError(f"Expected one row for coverage, got {len(count_by_geo_type_df)}.")
```
krivard, line 195: This sounds like it will generate a single timeseries where each point is from a different issue. Is that what we want, or do we want multiple timeseries, one as_of each issue? ie we compute a timeseries today for all the coverages from 2020-01-01 to today, as of today, then tomorrow we compute a new timeseries, etc. or maybe not since there will be a ton of duplication for signals that don't backfill?

benjaminysmith: I can see arguments for a couple of different options. I think it's probably more interesting to see the coverage for the most recent as-of (because it gets at the question of "if I want to use this data right now, how well does it cover the map") which would mean we need to change this to cover more days, and as you point out, process a list. We can do something similar to the existing coverage dashboards where we pull the last 28 days of data.

Right now though I am making assumptions in the code of having only a single data point, so putting this check in to assert that we are really in that state. I propose we start with this version, discuss in chat what the right version is, and then we can update the code and I will alter this check accordingly. wdyt?

krivard: sounds good!

krivard (approved):
lgtm

PR #459, opened 2021-03-19
Add screening test positivity to survey docs

nmdefries:


capnrefsmmat (approved):
ready to go when the signal deploys, I think

nmdefries:
Screening test positivity is now available in the API.

PR #460, opened 2021-03-22
Switch to covidcast for signal dash

benjaminysmith:
- Also fixes a datetime->date bug.
- Also updates the API to return all the coverage data, nested.

Progress on #438


krivard (approved):
👍

PR #461, opened 2021-03-23
Fix incorrect join by adding ID

benjaminysmith:
We were accidentally doing a cross-product of all signals with dates, so add the missing ID condition.

Improves #438.

@sgratzl as FYI, but don't worry about pulling this into Flask yet.


benjaminysmith:
Ignore this one -- it is superseded by https://github.com/cmu-delphi/delphi-epidata/pull/463

PR #462, opened 2021-03-23
Add metadata IDs to hhs datasets

chinandrew:
The new healthdata.gov API has different dataset IDs for the metadata and the raw data, so adding that as an attribute.

e.g.
data: https://healthdata.gov/dataset/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/6xf2-c3ie
metadata: https://healthdata.gov/dataset/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/4cnb-m4rz


krivard (approved):
lgtm

PR #463, opened 2021-03-23
Include coverage time series in signal dash status

benjaminysmith:
Changes:
- Add covidcast_signal to the dashboard signal defs
- Embed the coverage time-series in the status response
- Add missing ID condition to status
- Update tests

Improves #438


krivard (approved):
👍

PR #464, opened 2021-03-23
update date parsing for covid_hosp

chinandrew:
- Add support for slashes (and keep support for dashes)
- Truncate dates in case the value contains time at the end "e.g. yyyy/mm/dd hh:mm:ss"


krivard (approved):
lgtm

PR #465, opened 2021-03-24
update to latest main + sync latest signal changes

sgratzl:
yet another update. 

However, I don't know why the tests fail


krivard:
Looks like it's confused about imports. The signals dash code should be importing the `covidcast` library from PyPI, which is currently listed in the operations requirements.txt that builds the test. Instead, it's trying to import `covidcast` from `tests/acquisition/covidcast`, which (understandably) doesn't have the same members.

@benjaminysmith any ideas?

benjaminysmith:
> Looks like it's confused about imports. The signals dash code should be importing the `covidcast` library from PyPI, which is currently listed in the operations requirements.txt that builds the test. Instead, it's trying to import `covidcast` from `tests/acquisition/covidcast`, which (understandably) doesn't have the same members.
> 
> @benjaminysmith any ideas?

Looking...

benjaminysmith:
> > Looks like it's confused about imports. The signals dash code should be importing the `covidcast` library from PyPI, which is currently listed in the operations requirements.txt that builds the test. Instead, it's trying to import `covidcast` from `tests/acquisition/covidcast`, which (understandably) doesn't have the same members.
> > @benjaminysmith any ideas?
> 
> Looking...

@sgratzl do you see the failure when running via pytester?:
 run --rm --network delphi-net --env "SQLALCHEMY_DATABASE_URI=***delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" python3 -m undefx.py3tester.py3tester --color repos/delphi/delphi-epidata/tests

benjaminysmith:
> > > Looks like it's confused about imports. The signals dash code should be importing the `covidcast` library from PyPI, which is currently listed in the operations requirements.txt that builds the test. Instead, it's trying to import `covidcast` from `tests/acquisition/covidcast`, which (understandably) doesn't have the same members.
> > > @benjaminysmith any ideas?
> > 
> > 
> > Looking...
> 
> @sgratzl do you see the failure when running via pytester?:
> run --rm --network delphi-net --env "SQLALCHEMY_DATABASE_URI=***delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" python3 -m undefx.py3tester.py3tester --color repos/delphi/delphi-epidata/tests

Alright this definitely looks like something related to the python test loading behavior.

- Using py3tester works (which I believe is what is currently run in CI for delphi-epidata)
- Using vanilla pytest does not work and gives the error above
- Using "python -m  pytest --import-mode importlib" does work

@sgratzl -- it looks like you switched away from using py3tester to pytest -- is there a reason for that?

sgratzl:
> @sgratzl -- it looks like you switched away from using py3tester to pytest -- is there a reason for that?

afaik it is the goal to switch away from undexf py3tester to the standard pytest. But I can change the CI to use the last version you mentioned

PR #468, opened 2021-03-25
Update coverage to cover multiple days

benjaminysmith:
Change dashboard coverage to look back over previous 28 days instead of just the most recent day. Update tests.

Improves #438

### DIFF
--- a/src/acquisition/covidcast/signal_dash_data_generator.py
+++ b/src/acquisition/covidcast/signal_dash_data_generator.py
@@ -17,6 +17,8 @@ import delphi.operations.secrets as secrets
 from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
 
 
+LOOKBACK_DAYS_FOR_COVERAGE = 28
+
 @dataclass
 class DashboardSignal:
     """Container class for information about dashboard signals."""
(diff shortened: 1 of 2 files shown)


src/acquisition/covidcast/signal_dash_data_generator.py
```diff
@@ -184,25 +186,24 @@ def get_coverage(dashboard_signal: DashboardSignal,
...
-        date=latest_time_value,
-        geo_type=count_by_geo_type_df['geo_type'].iloc[0],
-        count=count_by_geo_type_df['count'].iloc[0].item())
-    signal_coverage_list.append(signal_coverage)
+    for _, row in count_by_geo_type_df.iterrows():
+        signal_coverage = DashboardSignalCoverage(
+            signal_id=dashboard_signal.db_id,
+            date=row['time_value'].date(),
+            geo_type=row['geo_type'],
+            count=row['count'])
```
krivard, line 205: What happens when we run this two days in a row:
* DashboardSignalCoverage timeseries for this signal (when retrieved using the signal dash API) gets longer by 1 day, though least-recent day was generated using (effectively) a different as_of than today's
* Previous DashboardSignalCoverage timeseries values for this timeseries are dropped first; timeseries remains 28 days long

benjaminysmith: I think it's interesting to keep the older data -- would be neat to see the coverage going back many months. As you point out the caveat is that it may be out of date, but we could either disclaim that ("any changes that happen after 28 days may not be included") or add longer backfill occasionally (I'm not a huge fan of that since it adds complexity). I suspect it probably doesn't change too much.

If you prefer to keep it more conceptually simple for now we can just drop the older points, though.

krivard: Cool, let's drop the older points then, it'll make the UI simpler/easier to interpret.

benjaminysmith: Sounds good -- mind if I do this in a follow-up so we can get more data into the system for testing sooner?

krivard (approved):
👍

PR #471, opened 2021-03-26
[fb-survey] Document new expansion pack indicators

nmdefries:
This documents a bunch of new indicators being added to the contingency tables, and notes prefix format of the grouping variables.

Most indicators that appear in both the API survey signals and the contingency tables have the same names, but [some names differ](https://docs.google.com/spreadsheets/d/1NzYCx46wjlNmQw_JDkG5bAYR2lYhX57i9kpruTKT_Mw/edit?usp=sharing) for conciseness and clarity. I've attempted to unify naming where possible and otherwise match the meaning in the API name (e.g. `hesitancy_reason_unnecessary` vs `barrier_dontneed`). The largest naming disparity (`hesitancy_reason` vs `barrier`) is noted in the documentation.

Some indicators previously released in the contingency tables were renamed in the standardization process; these have the former name noted.

This should be deployed after the expansion tables are uploaded and shareable.


nmdefries:
Including both the plain and "hesitant" versions of indicators makes the tables really long. We could consider listing only the plain versions, and describing the meaning and construction of "hesitant" indicators in a single place.

capnrefsmmat:
> Including both the plain and "hesitant" versions of indicators makes the tables really long. We could consider listing only the plain versions, and describing the meaning and construction of "hesitant" indicators in a single place.

That sounds good to me.

docs/symptom-survey/contingency-tables.md
```diff
@@ -90,34 +90,131 @@ alphabetically. Each aggregate reports four columns (unrounded):
...
+items V5a, V5b, V5c, or V5d) are asked a follow-up item asking why they don't
+believe they need the vaccine. These signals summarize the reasons selected.
+Respondents who do not select any reason (including "Other") are treated as
+missing.
+
+This item was shown to respondents starting in Wave 8.
+
+| Indicator | Description | Survey Item |
+| --- | --- | --- |
+| `pct_hesitant_dontneed_reason_had_covid` | Estimated percentage of respondents who say they don't need to get a COVID-19 vaccine because they already had the illness, among respondents who answered No, probably not" or "No, definitely not" when asked if they would get vaccinated if offered (item V3) and provided at least one reason for why they believe a COVID-19 vaccine is unnecessary. <br/> **Earliest date available:** 2021-02-08 | V6 |
```
capnrefsmmat, line 208: Two things:

1. Missing quotation mark on "No, probably not" for all of these
2. Are these really only among "No" respondents, and not "Yes, probably"? Because V6 is shown to people who answer "Yes, probably" but say they don't need the vaccine.

docs/symptom-survey/contingency-tables.md
```diff
@@ -90,34 +90,131 @@ alphabetically. Each aggregate reports four columns (unrounded):
...
+| `pct_hesitant_dontneed_reason_not_beneficial` | Estimated percentage of respondents who say they don't need to get a COVID-19 vaccine because they don't think vaccines are beneficial, among respondents who answered No, probably not" or "No, definitely not" when asked if they would get vaccinated if offered (item V3) and provided at least one reason for why they believe a COVID-19 vaccine is unnecessary. <br/> **Earliest date available:** 2021-02-08 | V6 |
+| `pct_hesitant_dontneed_reason_other` | Estimated percentage of respondents who say they don't need to get a COVID-19 vaccine for another reason, among respondents who answered No, probably not" or "No, definitely not" when asked if they would get vaccinated if offered (item V3) and provided at least one reason for why they believe a COVID-19 vaccine is unnecessary. <br/> **Earliest date available:** 2021-02-08 | V6 |
 
 Note: CSVs for the month of January 2021 only use data from January 6-31 due to
```
capnrefsmmat, line 216: I think this note should get moved up to the top now, since it's buried under a huge list of indicators.

capnrefsmmat (changes requested):
Some things to double-check and a couple typos to fix

docs/symptom-survey/contingency-tables.md
nmdefries: Yes, this is only among "No"/hesitant respondents. To calculate this indicator, I nulled out responses from people who'd gotten to V6 from V5a or V5d

capnrefsmmat (approved):
I think this is good to go, pending upload to AWPS and new links

nmdefries:
> TODO: Update links at top of docs, ideally to AWPS rather than Box.

AWPS needs more space for all the files, so this will continue to link to Box for now.

This is ready to be merged/deployed.

PR #478, opened 2021-04-02
[flask]: compatibility mode

### LINKED ISSUE #477: Align Flask behavior of legacy URL to match PHP API
The with Flask server in the load-balancing pool, the slight differences between it and the PHP API are starting to cause problems in existing usage. This isn't really a bug since the differences are all positive improvements, but it's a bit rude to yank existing behaviors out from under folks if we can avoid it.

Ideally, we should make queries to `/epidata/api.php?` behave identically with the PHP API, and provide all improvements (streaming, more descriptive error reporting, etc) only under the new URLs (`/epidata/covidcast/`). 

If it turns out that the easiest way to do this is to literally run the old PHP file, we can do that with the understanding that `api.php` is feature-frozen and will receive no development effort going forward.

If it turns out both approaches are prohibitively difficult, we should discuss which of the changes are the most disruptive vs easiest to fix and go from there.

The current list of failures we've run into is:
- [x] Including the `epidata` key in "No results" responses causes a KeyError in the `covidcast` library.

sgratzl:
closes #477 

e.g. 
image

whereas
image


krivard (approved):
👍  nice work

PR #479, opened 2021-04-05
Extend SafeGraph limitations section

capnrefsmmat:
This shouldn't be a subsection of Weekly Patterns (it applies to both), and we should mention research on its biases.

### DIFF
--- a/docs/api/covidcast-signals/safegraph.md
+++ b/docs/api/covidcast-signals/safegraph.md
@@ -121,9 +121,12 @@ meaning estimates for a specific day are only available 3-9 days later. It may
 take up to an additional day for SafeGraph's data to be ingested into the
 COVIDcast API.
 
-### Limitations
+## Limitations
 
-This data source is based on mobile devices that are members of SafeGraph panels, which is not necessarily the same thing as measuring the general public. This means that counts are subject to bias if some regions have a greater density of SafeGraph panel members as a percentage of the population. These counts do not represent absolute counts, and only count visits by members of the panel in that region.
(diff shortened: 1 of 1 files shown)


docs/api/covidcast-signals/safegraph.md
```diff
@@ -121,9 +121,12 @@ meaning estimates for a specific day are only available 3-9 days later. It may
...
+
+* **Geographic bias.** If some regions have a greater density of SafeGraph panel members as a percentage of the population than other regions, comparisons of metrics between regions may be biased. Regions with more SafeGraph panel members will appear to have more visits counted, even if the rate of visits in the general population is the same.
+* **Demographic bias.** SafeGraph panels may not be representative of the local population as a whole. For example, [some research suggests](https://arxiv.org/abs/2011.07194) that "older and non-white voters are less likely to be captured by mobility data", so this data will not accurately reflect behavior in those populations. Since population demographics vary across the United States, this can also contribute to geographic biases.
 
 The number of POIs coded as bars is much smaller than the number of POIs coded as restaurants. 
 SafeGraph's Weekly Patterns data consistently lacks data on bar visits for Alaska, Delaware, Maine, North Dakota, New Hampshire, South Dakota, Vermont, West Virginia, and Wyoming.
```
krivard, line 132: I like moving the general-safegraph limitations into a ## section, but you'll want to keep the Patterns-specific limitations in ###.

krivard (changes requested):

docs/api/covidcast-signals/safegraph.md
capnrefsmmat: Done, although I didn't add a `### Limitations` subsection, just put it under the other Patterns info because that section is already short and subsectioning doesn't seem too useful.

krivard (approved):
👍

PR #483, opened 2021-04-08
[flask]: /covidcast/trend endpoint

sgratzl:
based on #481 

adds a new custom endpoint `/covidcast/trend` which computes the trend for a given date and window.

e.g.

http://localhost:5000/covidcast/trend?signal=fb-survey:smoothed_cli&geo=state:ny&date=20210401&window=20200101-20210402

will result in: 
image

note: I just have limited local test data, so results may vary in your system

### DIFF
--- a/tests/server/endpoints/covidcast_utils/test_trend.py
+++ b/tests/server/endpoints/covidcast_utils/test_trend.py
@@ -27,6 +27,10 @@ class UnitTests(unittest.TestCase):
             self.assertAlmostEqual(compute_trend_value(9.5 + 5, 10 + 5, 5), -0.05)
             self.assertAlmostEqual(compute_trend_value(9 + 5, 10 + 5, 5), -0.1)
             self.assertAlmostEqual(compute_trend_value(8 + 5, 10 + 5, 5), -0.2)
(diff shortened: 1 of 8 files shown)


tests/server/endpoints/covidcast_utils/test_trend.py
```diff
@@ -0,0 +1,61 @@
+# standard library
+import unittest
+
+from delphi.epidata.server.endpoints.covidcast_utils.trend import compute_trend_value, compute_trend_class, TrendEnum, compute_trend, Trend
+
+
+class UnitTests(unittest.TestCase):
+    def test_compute_trend_value(self):
```
krivard, line 8: Add a check for when basis == min, basis != current.

& we never expect a case when basis == min, basis > current, right?

krivard (approved):
Minor test coverage nit otherwise 👍

tests/server/endpoints/covidcast_utils/test_trend.py
sgratzl: how can `basis > current and basis == min` be true, it would imply that `min > current`, so that the minimal value is larger than some other value

added the other tests

PR #485, opened 2021-04-08
[flask] add /covidcast/csv endpoint

sgratzl:
similar to covidcast-csv-server https://github.com/cmu-delphi/covidcast-csv-server.

as mentioned in the docs with the new streaming and custom endpoints we could add the csv server functionality into the flask api directly.

### DIFF
--- a/src/server/endpoints/covidcast.py
+++ b/src/server/endpoints/covidcast.py
@@ -1,9 +1,10 @@
-from typing import List, Iterable, Dict, Any
+from typing import List, Iterable, Dict, Any, Optional, Union, Tuple
 from itertools import groupby
-from flask import Blueprint, request
+from datetime import date, datetime, timedelta
+from flask import Blueprint, request, make_response
 
 from .._common import is_compatibility_mode
-from .._exceptions import ValidationFailedException
+from .._exceptions import ValidationFailedException, DatabaseErrorException
 from .._params import (
     GeoPair,
     SourceSignalPair,
@@ -15,7 +16,7 @@ from .._params import (
     parse_day_range_arg,
 )
 from .._query import QueryBuilder, execute_query, filter_integers, filter_strings, run_query, parse_row
-from .._printer import create_printer
+from .._printer import create_printer, CSVPrinter
 from .._validate import (
     extract_date,
     extract_dates,
@@ -24,7 +25,7 @@ from .._validate import (
     require_all,
     require_any,
 )
-from .covidcast_utils import compute_trend, shift_time_value
+from .covidcast_utils import compute_trend, shift_time_value, date_to_time_value, time_value_to_iso
 
 # first argument is the endpoint name
 bp = Blueprint("covidcast", __name__)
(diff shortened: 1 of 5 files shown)


src/server/endpoints/covidcast.py
```diff
@@ -170,3 +173,78 @@ def gen(rows):
...
     return p(gen(r))
+
+
+@bp.route("/csv", methods=("GET", "POST"))
+def handle_export():
+    source, signal = request.args.get("signal", "jhu-csse:confirmed_incidence_num").split(":")
+    start_day = request.args.get("start_day", "2020-04-01")
+    end_day = request.args.get("end_day", "2020-09-01")
+    geo_type = request.args.get("geo_type", "county")
+    geo_values = request.args.get("geo_values", "*")
```
krivard, line 184: Is there reasoning behind these defaults?

src/server/endpoints/covidcast.py
```diff
@@ -170,3 +173,78 @@ def gen(rows):
...
+    # execute query
+    try:
+        r = run_query(p, (str(q), q.params))
+    except Exception as e:
+        raise DatabaseErrorException(str(e))
+
+    # special case for no data to be compatible with the CSV server
+    first_row = next(r, None)
+    if not first_row:
+        return "No matching data found for signal {source}:{signal} " "at {geo} level from {start_day} to {end_day}, as of {as_of}.".format(
```
krivard, line 245: What's with the `" "`?

src/server/endpoints/covidcast.py
```diff
@@ -170,3 +173,78 @@ def gen(rows):
...
+    q.params["start_day"] = date_to_time_value(start_day)
+    q.params["end_day"] = date_to_time_value(end_day)
+    q.where_geo_pairs("geo_type", "geo_value", [GeoPair(geo_type, True if geo_values == "*" else geo_values)])
+
+    _handle_lag_issues_as_of(q, None, None, date_to_time_value(as_of) if as_of is not None else None)
+
+    # tag as_of in filename, if it was specified
+    as_of_str = "-asof-{as_of}".format(as_of=as_of.isoformat()) if as_of is not None else ""
+    filename = "covidcast-{source}-{signal}-{start_day}-to-{end_day}{as_of}".format(source=source, signal=signal, start_day=start_day.isoformat(), end_day=end_day.isoformat(), as_of=as_of_str)
+    p = CSVPrinter(filename)
```
krivard, line 213: This is a good idea but we should make sure the covidcast library isn't expecting a universal/static filename before merging

krivard (commented):
Small questions

sgratzl:
all of those questions are regarding code that is directly taken from 

https://github.com/cmu-delphi/covidcast-csv-server/blob/main/src/app.py#L44-L89

I just copied them to mimic its behavior

src/server/endpoints/covidcast.py
sgratzl: I don't understand, this new endpoint is solely for  https://delphi.cmu.edu/covidcast/export/ to replace the covidcast-csv-server

krivard: ah, I misunderstood. This is fine then

krivard (approved):
👍  as a separate covidcast/export endpoint

PR #488, opened 2021-04-12
[flask]: add /covidcast/backfill endpoint

sgratzl:
based on #484 

adds a new `/covidcast/backfill` endpoint that can be used to query all issues for a given signal/geo combo over a given period of time:

image

Based on this data, one can create a backfill profile and visualization


chinandrew:
I don't have a whole lot of context, what's the purpose of this and the correlations endpoint?

sgratzl:
> I don't have a whole lot of context, what's the purpose of this and the correlations endpoint?

both (and others before) are about custom COVIDcast 2.0 endpoints to move logic and algorithms from the frontend to the backend. In this specific case it is about fetching all issues for a signal/region at once instead of doing numerous as_of/lag calls

sgratzl:
note: they are not that urgent since the flask api is only partially in production yet. Thus, the frontend cannot use those custom API calls yet

krivard (approved):
lgtm

PR #491, opened 2021-04-19
Enable New Relic APM (application performance monitoring) reporting for the flask container

korlaxxalrok:
This PR does the following:

- Installs the New Relic Python agent.

- Adds a simple wrapper to help determine if we should start via `start.sh` or
  with `new-relic run-program start.sh`.

  - The wrapper makes a decision based on the existence of the `NEW_RELIC_LICENSE_KEY`
    environment variable. This is configued in `delphi-ansible-web`. Additional
    New Relic-specific configuration can be supplied in that repository as well.

  - When enabled and started with the NR commands the container will log useful
    APM data to Delphi's New Relic account.
    
    See [https://docs.newrelic.com/docs/agents/python-agent/installation/install-python-agent-docker/]


sgratzl (approved):
sweet

PR #492, opened 2021-04-21
[flask]: send something for empty csv files

sgratzl:
maybe otherwise the proxy is waiting till it "end" but there is never anything


chinandrew (approved):
:shipit: :shipit: :shipit: 

wait to deploy on staging since @korlaxxalrok  did some deploy stuff to staging yesterday

PR #493, opened 2021-04-22
Only mention raw signals where they exist

capnrefsmmat:
Since this confused at least one user into thinking there were raw versions for all survey signals.

### DIFF
--- a/docs/api/covidcast-signals/fb-survey.md
+++ b/docs/api/covidcast-signals/fb-survey.md
@@ -96,7 +96,7 @@ survey data, we estimate the percentage of people (age 18 or older) who have a
 COVID-like illness, or influenza-like illness, in a given location, on a given
 day.
 
-Signals beginning `smoothed_w` are [adjusted using survey weights
+Signals beginning `raw_w` or `smoothed_w` are [adjusted using survey weights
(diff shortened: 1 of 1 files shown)


docs/api/covidcast-signals/fb-survey.md
```diff
@@ -96,7 +96,7 @@ survey data, we estimate the percentage of people (age 18 or older) who have a
 COVID-like illness, or influenza-like illness, in a given location, on a given
 day.
 
-Signals beginning `raw_w` or `smoothed_w` are [adjusted using survey weights
+Signals beginning `smoothed_w` are [adjusted using survey weights
```
nmdefries, line 99:
```suggestion
Signals beginning `raw_w` or `smoothed_w` are [adjusted using survey weights
```

We should leave `raw_w` here to be clear that they are also weighted estimates.

nmdefries (changes requested):

docs/api/covidcast-signals/fb-survey.md
capnrefsmmat: ah yes, overzealous find-and-replace

nmdefries (approved):
Looks good!

PR #502, opened 2021-04-29
Remove google survey and ght from changelog

### LINKED ISSUE #501: Drop the GHT section from the changelog in the API docs
GHT is now deprecated, and never had any entries in the changelog anyhow, so we might as well just drop it.

https://cmu-delphi.github.io/delphi-epidata/api/covidcast_changelog.html#ght

chinandrew:
Both of these are inactive and have no changes. 

Fixes #501


krivard (approved):
🚀  you're a star, ty

PR #506, opened 2021-05-03
[flask]: add covidcast/meta endpoint

sgratzl:
to have an optimized data structure to transfer less data and guess some meta data about the signals


krivard (approved):
👍

PR #507, opened 2021-05-03
Fix links in HHS API docs and add state_daily info

### LINKED ISSUE #505: Fix HHS Hosp API doc links
Links in the epidata api documentation [need to be updated](https://app.slack.com/client/TF95CN37E/CUQFVN7T9/thread/CUQFVN7T9-1619888725.023100) after  https://github.com/cmu-delphi/delphi-epidata/issues/446

chinandrew:
Forgot to change these links after #445. Fixes #505. Also adds info about how the state data is a combination of the state timeseries and state daily datasets. Wasn't sure how much info we wanted to put on that, so currently just left it as close to the original text as possible.

### DIFF
--- a/docs/api/covid_hosp.md
+++ b/docs/api/covid_hosp.md
@@ -6,13 +6,16 @@ parent: Epidata API (Other Diseases)
 # COVID-19 Hospitalization by State
 
 This data source is a mirror of the "COVID-19 Reported Patient Impact and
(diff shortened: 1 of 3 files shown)


krivard (approved):
xlnt

capnrefsmmat:
@chinandrew  Missed one link:

Screen Shot 2021-05-10 at 11 44 44 AM

PR #513, opened 2021-05-05
[flask] separate image step and only build on certain branches

### LINKED ISSUE #489: [flask]: limit image generation to certain sub branches
once the flask implementation is in dev/main we should limit the docker image generation to only certain branches to avoid that the container registry gets filled with useless docker images.

sgratzl:
closes #489


krivard (approved):
👍

PR #515, opened 2021-05-05
build: match is required for globbing

sgratzl:
see https://github.com/cmu-delphi/github-deploy-repo/blob/main/src/actions/copymove.py#L138


korlaxxalrok (approved):
LGTM 👍

PR #516, opened 2021-05-05
[flask]: add index hint

### LINKED ISSUE #509: Use an index hint for queries of <N geo values
When someone queries covidcast for a small number of geo values over a large timespan, mariadb tends to pick the wrong index (`source`) and has to throw out huge numbers of rows with extraneous geo values, particularly for dense signals like cases/deaths or safegraph. Instead, queries like these should hint with the correct index:

```
SELECT ... FROM ... USE INDEX (by_issue) WHERE ...
```

(`by_issue` and `by_lag` work equally well for default "as of today" queries, but obvs if someone is querying specifically by issue or by lag, use the right one)

Possibly related to https://github.com/cmu-delphi/www-covidcast/issues/918, since cases/deaths queries of only one or two counties take ages

sgratzl:
closes #509 

based on #513 to avoid image


krivard (approved):
Nice work! 5 is a reasonable guess at a crossover count for now since it'll cover frontend usage. We can set a summer intern on tuning that value with actual profiling information to cover research usage.

PR #525, opened 2021-05-06
feat: support filtering fields in correlation, trend, and backfill

sgratzl:
to only return a subset of the fields if needed


krivard (approved):
xlnt

PR #527, opened 2021-05-06
feat: add /trendseries for a time series of trends

sgratzl:
computes trends for each time point in a series

### DIFF
--- a/src/server/endpoints/covidcast_utils/trend.py
+++ b/src/server/endpoints/covidcast_utils/trend.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass, asdict
-from typing import Optional, Iterable, Tuple
+from typing import Optional, Iterable, Tuple, Dict, List, Callable
 from enum import Enum
+from collections import OrderedDict
+from ...utils import shift_time_value
 
 
 class TrendEnum(str, Enum):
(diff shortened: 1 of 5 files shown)


src/server/endpoints/covidcast_utils/trend.py
```diff
@@ -62,6 +65,54 @@ def compute_trend(geo_type: str, geo_value: str, signal_source: str, signal_sign
...
+        )
+
+        trends.append(t)
+
+        if t.value is None or t.min_value is None:
+            continue
+
+        t.basis_trend = compute_trend_class(compute_trend_value(t.value, t.basis_value, t.min_value)) if t.basis_value else TrendEnum.unknown
+        t.min_trend = compute_trend_class(compute_trend_value(t.value, t.min_value, t.min_value))
+        t.max_trend = compute_trend_class(compute_trend_value(t.value, t.max_value, t.min_value)) if t.max_value else TrendEnum.unknown
```
krivard, line 111: is the downstream code that makes use of these attributes aware that they are not always set?

sgratzl: yes, see also interface on the www-covidcast side: https://github.com/cmu-delphi/www-covidcast/blob/9ba9f934aa37c24c936cb06374f203f061cebd1e/src/data/api.ts#L115-L136

krivard (approved):
👍

PR #528, opened 2021-05-06
feat: exclude fields

### LINKED ISSUE #524: Add option to filter out fields by removing unwanted ones
so far it is possible to limit the returned fields in a positive way by using the `fields=` parameter, e.g., `fields=geo_value,value` will only return these two columns. An alternative approach would be support negative fields, e.g. `fields_exclude=stderr` won't transfer `stderr` but everything else. Alternative syntax: prepending a minus, e.g., `fields=-stderr`

sgratzl:
closes #524

by using `fields=-fieldname` syntax

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -274,6 +274,31 @@ class CovidcastTests(unittest.TestCase):
       'message': 'success',
     })
 
+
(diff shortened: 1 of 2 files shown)


krivard (approved):
👍

PR #532, opened 2021-05-10
Remove newline from link

### LINKED ISSUE #531: Broken link under combination indicator page
Hello,
The hyperlink under "Statistical Combination Signals (Inactive)" on this page is broken.
https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/indicator-combination.html
We have been using the NMF combined indicator (nmf_day_doc_fbc_fbs_ght) since last October for a project. I understand you stopped updating this as of March this year but I would like to read more about this indicator and see if we can recreate it ourselves. 
Thanks!

krivard:
In theory a newline inside link text shouldn't cause a problem, but for some reason Jekyll isn't rewriting the URL on this one so maybe it's using a more conservative filter?

Fixes #531

### DIFF
--- a/docs/api/covidcast-signals/indicator-combination.md
+++ b/docs/api/covidcast-signals/indicator-combination.md
@@ -21,8 +21,8 @@ calculated or composed by Delphi. It is not a primary data source.
 ## Statistical Combination Signals (Inactive)
 
 The NMF combination signals were deactivated on March 17, 2021. Documentation for
-these signals is still available on the page for [inactive indicator-combination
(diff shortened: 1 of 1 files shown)


krivard:
Preview: https://krivard.github.io/delphi-epidata/api/covidcast-signals/indicator-combination.html

krivard:
ah @sgratzl looks like Alex merged changes directly into main instead of going through dev first, should I drop those commits from this pr or leave them in?

PR #538, opened 2021-05-12
add related signals to /covidcast/meta

sgratzl:
e.g.:

image

adds more information about the signals
 * is_cumulative
 * is_smoothed
 * is_weighted
 * format = per100k | raw | raw_count | percentage | fraction
 * related_signals ... list of signals in the same data source that are related (different set of flags)


krivard (approved):
excellent

PR #540, opened 2021-05-13
add version endpoint

### LINKED ISSUE #529: add endpoint to return api version
currently there is no way to know which version is currently running on the system

+ do we have a versioning of delphi-epidata?

sgratzl:
closes #529 

add a simple version endpoint `/version` which returns the current epidata version. Moreover, start with Github Release Action workflow similar to www-main or www-covidcast. For now, it is just about bumping the version and merging dev to main


krivard (approved):
xlnt

PR #541, opened 2021-05-14
[Prototype] Client-side caching option for Epidata.covidcast Python

dshemetov:
### Description
`Epidata.covidcast` is idempotent on the hour-scale (I think). This PR adds client-side caching to `Epidata.covidcast` in Python, to save user effort in writing their own file saving mechanisms.

Example:
```
# works as normal
> df = Epidata.covidcast('source', 'signal', 'day', 'county', 20200414, '01234')
# a regular call to the API that caches locally
> df = Epidata.covidcast('source', 'signal', 'day', 'county', 20200414, '01234', cache_timeout=20)
# performs a local lookup (if run within 20s)
> df = Epidata.covidcast('source', 'signal', 'day', 'county', 20200414, '01234', cache_timeout=20)
```

The goal is to expose `cache_timeout` to the `covidcast` Python library in another PR. We can keep the default behavior cacheless to remove hidden state issues. 

Overlapping, but not identical time windows will look to the cache and the web gracefully (because the `covidcast` Python library splits date range requests into daily `Epidata.covidcast` calls). This is not true for geos (but could be if we can provide metadata of available geos per indicator).

### Changes
Switch to [requests_cache](https://requests-cache.readthedocs.io/en/v0.6.3/) when the `cache_timeout` argument is sent.


src/client/delphi_epidata.py
```diff
@@ -63,9 +65,18 @@ def _request(params):
...
+      if cache_timeout is not None:
+        session = requests_cache.CachedSession(
+          'demo_cache', expire_after=cache_timeout
+        )
+      else:
+        session = requests.Session()
+      req = session.request('get', Epidata.BASE_URL, params, headers=_HEADERS)
+      # Fallback to requests if we have to use POST
       if req.status_code == 414:
         req = requests.post(Epidata.BASE_URL, params, headers=_HEADERS)
```
chinandrew, line 78: this line probably also should be updated to use `session` for consistency

integrations/client/test_delphi_epidata.py
```diff
@@ -284,21 +284,28 @@ def test_covidcast(self):
...
+  def test_request_method(self, _Session, _CachedSession):
     """Test that a GET request is default and POST is used if a 414 is returned."""
-    with self.subTest(name='get request'):
+    with self.subTest(name='get request, no cache'):
+      Session = MagicMock()
+      _Session.return_value = Session
       Epidata.covidcast('src', 'sig', 'day', 'county', 20200414, '01234')
-      get.assert_called_once()
-      post.assert_not_called()
+      Session.request.assert_called_once()
```
chinandrew, line 295: I'd use the assert_called_once_with() functions here to check the post vs get calls

src/client/delphi_epidata.py
```diff
@@ -601,8 +610,13 @@ def covidcast(
     if 'format' in kwargs:
       params['format'] = kwargs['format']
 
+    if 'cache_timeout' in kwargs:
+      cache_timeout = kwargs['cache_timeout']
+    else:
+      cache_timeout = None
```
chinandrew, line 616: since there's a default value here, would just make the cache timeout variable a function argument with a default None value

tests/client/test_delphi_epidata.py
```diff
@@ -2,10 +2,15 @@
 
 # standard library
 import unittest
+import time
```
chinandrew, line 5: what are all these changes for?

integrations/client/test_delphi_epidata.py
dshemetov: Do you know if there's a way to check only one of the arguments win `assert_called_once_with`? I didn't want to write out the full params dict, but maybe I'm just being lazy.

tests/client/test_delphi_epidata.py
dshemetov: 1. I realized that cache_timeout can take `int`, `timedelta`, and `datetime`, so I wanted to be clear about that.
2. Removed unused imports.

dshemetov: Oh this is a different file, oops. Um, I wrote my own benchmarking tests in this file and then reverted it, but it looks like some imports stayed. Thanks for catching.

integrations/client/test_delphi_epidata.py
chinandrew: You could try accessing the mocks call_args_list attribute and checking for inclusion of an element

integrations/client/test_delphi_epidata.py
```diff
@@ -284,21 +284,31 @@ def test_covidcast(self):
...
+  def test_request_method(self, _Session, _CachedSession):
     """Test that a GET request is default and POST is used if a 414 is returned."""
-    with self.subTest(name='get request'):
+    with self.subTest(name='get request, no cache'):
+      Session = MagicMock()
+      _Session.return_value = Session
       Epidata.covidcast('src', 'sig', 'day', 'county', 20200414, '01234')
-      get.assert_called_once()
-      post.assert_not_called()
+      assert call('get') in Session.request.call_args_list()
```
chinandrew, line 295: nit: check that the length of the call args (number of calls) is correct. otherwise lgtm

krivard:
Are any of these changes:
* breaking
* announce-able
* needing public documentation

PR #548, opened 2021-05-17
upgrade JS client

sgratzl:
changes
 * add jest testing
 * upgrade to JavaScript only and removed CoffeeScript
 * add new class version `EpidataAsync` which returns Promises instead of expecting a callback param
 * wrap with an UMD header such that the library can be used both in the browser, AMD, and CommonJS (node) setups
 * add TypeScript typings


krivard (approved):
nice work

PR #556, opened 2021-05-21
[fb-survey] Describe and link contingency rollup files

nmdefries:
Document contingency rollup files.


capnrefsmmat:
Is there any reason to provide access to the non-rollup files? The rollup files seem like they are strictly better.

nmdefries:
RE usefulness: If someone already has all the historical files, they could want to download only new data, although the rollup files, especially in compressed form, aren't huge.

> I'd advertise the rollup files first, and stick the non-rollup link somewhere obscure on the page.

I'll do this for now, but I can remove the individual files entirely if you think they definitely won't be useful.

docs/symptom-survey/contingency-tables.md
```diff
@@ -91,6 +99,11 @@ is a concatenated list of other grouping variables used, ordered
 alphabetically. Values for variables used in file naming align with those
 within files as specified in the column section below.
 
+"Rollup" files containing all time periods for a given period type-aggregation
+type combination have names of the form:
+
+    {period_type}_{geo_type}_{aggregation_type}.csv.gz
+
```
capnrefsmmat, line 106: I think if you move this description up a bit (before the naming of the non-rollup files), this is ready to merge

capnrefsmmat (changes requested):
Good to go with one change

capnrefsmmat (approved):

PR #557, opened 2021-05-24
Add retry to Epidata request

chinandrew:
Add a single retry if an exception is encountered when making a request to the Epidata endpoint. If the retry fails, the exception encountered is reraised.

Also learned about reset_mock() so slipped that into an old test.

Requires https://github.com/cmu-delphi/operations/pull/66 otherwise the CI will fail on the import


krivard (approved):
👍

PR #560, opened 2021-05-27
add all endpoints to JS client

sgratzl:
except the signal_dashboard ... ones since they are more internal

added new endpoints
 * twitter
 * wiki
 * kcdc_ili

reordered the client to match the alphabetical order of the endpoints

plus: make the wiki language parameter optional similar to the R client


krivard (approved):
👍

PR #562, opened 2021-05-27
unify client versions and bump them at once

sgratzl:
* add the `client_version` to all clients
 * add the version function to the JS client using the new server version endpoint
 * unify bumping all versions when triggering the GitHub Action

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -12,3 +12,5 @@ tag = False
 [bumpversion:file:src/client/packaging/npm/package.json]
 
 [bumpversion:file:src/client/packaging/pypi/setup.py]
+
+[bumpversion:file:src/client/packaging/pypi/delphi_epidata/__init__.py]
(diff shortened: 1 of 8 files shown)


.bumpversion.cfg
```diff
@@ -4,3 +4,11 @@ commit = False
...
 
 [bumpversion:file:src/server/_config.py]
+
+[bumpversion:file:src/client/delphi_epidata.js]
+
+[bumpversion:file:src/client/delphi_epidata.R]
+
+[bumpversion:file:src/client/packaging/npm/package.json]
+
+[bumpversion:file:src/client/packaging/pypi/setup.py]
```
krivard, line 14: do you also need

```
[bumpversion:file:src/client/packaging/pypi/delphi_epidata/__init__.py]
```

?

sgratzl: good catch, I was looking at the delphi_epidata.py only which is extracting the version

krivard (approved):
👍

PR #564, opened 2021-05-27
create meta data covidcast file + use in /meta

sgratzl:
/covidcast/meta will be now:

image

the excel sheet with the meta data: https://docs.google.com/spreadsheets/d/1zb7ItJzY5oq1n-2xtvnPBiJu2L3AqmCKubrLkKJZVHs/edit#gid=0

and there is a GitHub action to update the CSV files from the Google Docs

TODO
 * [x] resolve aliases in covidcast and otherwise
 * ~[ ] update the dash signal generator to also use this metadata instead of the table~


sgratzl:
@dshemetov will also be interesting for you, since the meta data will give you information regarding which signal to derive from which one and in which way

sgratzl:
>  update the dash signal generator to also use this metadata instead of the table
invalid with the new /covidcast/coverage endpoint this job won't be needed anymore since it will be computed on the fly

krivard:
Roni has cleared the doc for publication. Anything else needed before this can come out of draft mode?

krivard (approved):
🚀

PR #568, opened 2021-06-08
add minimum issue date to the metadata with some tests

zhuoran-Cheng16:


zhuoran-Cheng16:
Duplicate of #

src/acquisition/covidcast/database.py
```diff
@@ -275,6 +275,7 @@ def compute_covidcast_meta(self, table_name='covidcast', use_index=True):
         ROUND(STD(`value`),7) AS `stdev_value`,
         MAX(`value_updated_timestamp`) AS `last_update`,
         MAX(`issue`) as `max_issue`,
+        MIN(`issue`) as `min_issue`,
```
krivard, line 278: We want min over all issues for this source+signal, not just the ones that are marked latest. This will probably need to be done as a separate query, since all the other stats should only apply to the latest issues.

krivard (changes requested):
Is this in regards to #236 ? If so, then the covid_hosp and signals dash modifications are unnecessary, though they were probably good practice.

src/acquisition/covidcast/database.py
zhuoran-Cheng16: Ok! Me and Tara will go through to find out what can we do to this problem.

melange396 (changes requested):
good stuff, but i found a few points.  also, the stuff in the `covid_hosp` subdirectory shouldnt (?) affect the metacache for the `covidcast` table and is outside the scope of this.

krivard:
Could you rebase this onto `dev` please?

zhuoran-Cheng16:
> Could you rebase this onto `dev` please?

PR #571, opened 2021-06-10
dynamically check if the missing_* columns exist

sgratzl:
use SQLAlchemy to check whether the missing columns (`missing_value`, ...) exists in the covidcast table. If they to, query them, if they don't provide dummy values instead. 

This could be a workaround till we have the database updated

### DIFF
--- a/src/server/endpoints/covidcast.py
+++ b/src/server/endpoints/covidcast.py
@@ -31,6 +31,7 @@ from .._validate import (
     require_all,
     require_any,
 )
+from .._db import sql_table_has_columns
 from .._pandas import as_pandas
 from .covidcast_utils import compute_trend, compute_trends, compute_correlations, compute_trend_value, CovidcastMetaEntry, AllSignalsMap
 from ..utils import shift_time_value, date_to_time_value, time_value_to_iso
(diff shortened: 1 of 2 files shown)


src/server/endpoints/covidcast.py
```diff
@@ -144,6 +145,15 @@ def handle():
...
     q.set_fields(fields_string, fields_int, fields_float)
 
+    missing_fields = ["missing_value", "missing_stderr", "missing_sample_size"]
+    fields_int.extend(missing_fields)
+    if sql_table_has_columns("covidcast", missing_fields):
+        # real fields
+        q.fields.extend([f"{q.alias}.{field}" for field in missing_fields])
+    else:
+        # fake fields
+        q.fields.extend([f"0 as {field}" for field in missing_fields])
```
chinandrew, line 155: Are we ok with setting 0 as the default value (as opposed to "not implemented" or something), since it means entry not missing, which isn't necessarily true for existing data.

sgratzl: what would be an alternative value?

chinandrew: That's....a good question. Does it make sense to not return those fields at all?

krivard: Prefer not returning those fields at all, unless that will continue to block a project we were hoping to unblock with this PR. How many of the things that are currently blocked on nans, actually need the missingness columns and not just a compatible database+server?

- [ ] dashboard builder
- [ ] speed improvements for frontend queries (there are piles of these; are any of them trying to explicitly exclude the missingness columns? & liable to get confused when there are no missingness columns to exclude?)
- [ ] advanced metadata
- [x] quidel revisions (need missingness annotations to handle deletions)

sgratzl: for now I changed it that the columns are not returned. I guess for the quidel revisions returning dummy values won't help this sub project either

krivard (approved):
👍  merge at will

PR #572, opened 2021-06-11
add a simple coverage endpoint

sgratzl:
e.g. `/covidcast/coverage?signal=fb-survey:smoothed_cli&geo_type=state&days=20` would return the state count between today and the last 20 days.

image

params
 * `days=<int>` optional number of days back, default = 30
 * `latest=YYYYMMDD` optional latest day, default = today
 * `geo_type=only-county` special county geo_type which filters out mega counties
 * `window=YYYYMMDD-YYYYMMDD` specify the time window to return the coverage for


krivard (approved):
👍

PR #580, opened 2021-06-14
add and modified the SQL instances

zhuoran-Cheng16:


krivard:
Already merged as part of #574

PR #581, opened 2021-06-14
Fix the incorrect coding issue and document it

capnrefsmmat:
I tried to expand the description because it's a little confusing to understand what happened. Is this clear enough?


docs/symptom-survey/problems.md
```diff
@@ -17,6 +17,31 @@ results. All problems will be logged here.
...
+
+For example, item C13 on waves 4-8 asked respondents to select which activities
+they did in the previous 24 hours. Item C13a carried forward any choices they
+selected and asked if they wore a mask while doing those activities. The
+numerical coding of responses in C13 was not consecutive, but the documentation
+of C13a suggested that responses to it *were* coded consecutively.
+
+However, in the microdata made available to users with data access, the coding
+of C13 and C13a is identical. The Qualtrics-exported document giving the coding
+is in error.
```
nmdefries, line 34: I'm confused by this bit. This is clearer to me/more in line with my understanding:

```suggestion
For example, item C13 on waves 4-8 asked respondents to select which activities
they did in the previous 24 hours. Since item C13a asked respondents if they wore a 
mask while doing those activities, it was set up to carry forward response options 
from C13 so that the wording and numerical coding of response options matched 
between the two items.

The numerical coding of responses in C13 was not consecutive, but the 
documentation of C13a suggested that responses to it *were* coded consecutively. 
However, in the microdata made available to users with data access, the coding of
C13 and C13a is identical. It is the Qualtrics-exported document giving the coding 
that is in error.
```

capnrefsmmat: Thanks, I like this.

PR #582, opened 2021-06-14
Modified SQL queries and keywords

zhuoran-Cheng16:
Change the SQL queries by adding the corresponding column name such as (id, source, signal, time_type, etc...) between 'covidcast' and 'values' to make it explicit, and capitalize all the SQL keywords

### DIFF
--- a/integrations/acquisition/covidcast/test_covidcast_meta_caching.py
+++ b/integrations/acquisition/covidcast/test_covidcast_meta_caching.py
@@ -67,11 +67,11 @@ class CovidcastMetaCacheTests(unittest.TestCase):
 
     # insert dummy data
     self.cur.execute(f'''
-      INSERT INTO
-        `covidcast` (`id`, `source`, `signal`, `time_type`, 
-        `geo_type`, `time_value`, `geo_value`, `value_updated_timestamp`, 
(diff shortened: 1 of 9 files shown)


integrations/acquisition/covidcast/test_covidcast_meta_caching.py
```diff
@@ -67,11 +67,11 @@ def test_caching(self):
...
-      INSERT INTO
-        `covidcast` (`id`, `source`, `signal`, `time_type`, 
-        `geo_type`, `time_value`, `geo_value`, `value_updated_timestamp`, 
-        `value`, `stderr`, `sample_size`, `direction_updated_timestamp`, 
-        `direction`, `issue`, `lag`, `is_latest_issue`, `is_wip`) 
+      INSERT INTO covidcast
+          (id, data_source, signal, time_type, geo_type,
+          time_values, geo_value, value_updated_timestamp,
+          value, stderr, sample_size, direction_updated_timestamp, direction, issue,
+          lag, result, is_wip)
```
krivard, line 74: Keep the backticks! Some of the column names (notably `signal`) are reserved words, and the query won't run if you don't escape the column names properly.

krivard (changes requested):

PR #587, opened 2021-06-15
ci: enforce origin/dev branch in creating release

sgratzl:
`dev` didn't seem to be unique which is weird since it is working in other repos


krivard (approved):
👍

PR #594, opened 2021-06-16
Add issue range arguments to update dataset function

chinandrew:
**Prerequisites**:

- [x] Build is successful

### Summary
Add support to the `update_dataset` function so that it can take in date ranges to pass to `issues_to_fetch`. It's not as precise as specifying individual URLs to acquire, but is simple because it uses existing functionality.

To use it to hopefully patch our missing issues, something like the following would get run on the prod server.
```
from delphi.epidata.acquisition.covid_hosp.state_daily.database import Database
from delphi.epidata.acquisition.covid_hosp.state_daily.network import Network
from delphi.epidata.acquisition.covid_hosp.common.utils import Utils

Utils.update_dataset(Database, Network, date(2021, 6, 10), date(2021, 6, 13))
```
from the prod server. We should test on staging first.

Also fixed a random old docstring to be a bit more clear.

### DIFF
--- a/integrations/acquisition/covid_hosp/state_daily/test_scenarios.py
+++ b/integrations/acquisition/covid_hosp/state_daily/test_scenarios.py
@@ -1,19 +1,21 @@
 """Integration tests for acquisition of COVID hospitalization."""
 
 # standard library
+from datetime import date
 import unittest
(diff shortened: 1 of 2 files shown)


krivard (approved):
💯

PR #598, opened 2021-06-16
finish add SQL Query and pass the integrations tests

zhuoran-Cheng16:
### Summary

  Finish the SQL query part and pass all the unit test and integrations test.


krivard (approved):
Nice work!

PR #601, opened 2021-06-17
Update development guide to correct images

TaraLakdawala:
Update development guide to use correct python images based on CI recipe

Edit: made the delphi_web_epidata image command more concise

### DIFF
--- a/docs/epidata_development.md
+++ b/docs/epidata_development.md
@@ -82,8 +82,12 @@ From the root of your workspace, all of the images can be built as follows:
 docker build -t delphi_web \
   -f repos/delphi/operations/dev/docker/web/Dockerfile .
 
-docker build -t delphi_web_epidata \
-  -f repos/delphi/delphi-epidata/dev/docker/web/epidata/Dockerfile .
+docker build -t delphi_web_python \
+  -f repos/delphi/delphi-epidata/dev/docker/python/Dockerfile .
+
+cd ./repos/delphi/delphi-epidata
+docker build -t delphi_web_epidata -f ./devops/Dockerfile .
+cd ../../../
 
 docker build -t delphi_database \
   -f repos/delphi/operations/dev/docker/database/Dockerfile .
--- a/docs/epidata_development.md
+++ b/docs/epidata_development.md
@@ -85,9 +85,8 @@ docker build -t delphi_web \
 docker build -t delphi_web_python \
   -f repos/delphi/delphi-epidata/dev/docker/python/Dockerfile .
 
-cd ./repos/delphi/delphi-epidata
-docker build -t delphi_web_epidata -f ./devops/Dockerfile .
-cd ../../../
+docker build -t delphi_web_epidata \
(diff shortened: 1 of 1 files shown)


docs/epidata_development.md
```diff
@@ -82,8 +82,12 @@ From the root of your workspace, all of the images can be built as follows:
...
   -f repos/delphi/operations/dev/docker/web/Dockerfile .
 
-docker build -t delphi_web_epidata \
-  -f repos/delphi/delphi-epidata/dev/docker/web/epidata/Dockerfile .
+docker build -t delphi_web_python \
+  -f repos/delphi/delphi-epidata/dev/docker/python/Dockerfile .
+
+cd ./repos/delphi/delphi-epidata
+docker build -t delphi_web_epidata -f ./devops/Dockerfile .
+cd ../../../
```
krivard, line 90: huh. does
```
docker build -t delphi_web_epidata \
   -f repos/delphi/delphi-epidata/devops/Dockerfile .
```

not work?

TaraLakdawala: It does work, but I decided to use the version in the CI Recipe [here](https://github.com/cmu-delphi/delphi-epidata/blob/76cc4c513fe1fa64eede08a6a9202aaa25e0dc1b/.github/workflows/ci.yaml#L58) but I'll update it to the more concise version

krivard (commented):
Looks like the `delphi_python` image name occurs a few more times in this document, largely in the instructions for running unit and integration tests; is that intentional or should those be changed to `delphi_web_python`?

dshemetov:
@krivard I think they can probably be updated to `delphi_web_python`. 

Also, not sure if we are deprecating py3tester, but I've been running the Flask tests with
```
docker run -i --rm --network delphi-net --env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" delphi_web_python python -m pytest --import-mode importlib repos/delphi/delphi-epidata/tests/
```

If so, maybe the unit test commands are worth updating?

krivard:
> not sure if we are deprecating py3tester, but I've been running the Flask tests with `--import-mode importlib`

We're not deprecating py3tester in the sense of scrubbing it from the docs and removing the compatibility code for it, more like we're just encouraging people to use pytest whenever possible. [There are a couple of pytest commands in the unit tests section](https://github.com/cmu-delphi/delphi-epidata/blob/63f048bcc7cb1ffe7f598c05d09180186c195e22/docs/epidata_development.md#unit-tests) though which should be modified to include your edits if they don't work as written. I'm equally happy for that to be part of this PR or handled separately, whatever you prefer @TaraLakdawala

TaraLakdawala:
It definitely makes sense to update the unit tests now for consistency's sake and I'm happy to add those updates in my next commit!

krivard:
wontfix; invalidated by Makefile and local dev tools

PR #606, opened 2021-06-21
add Min_issue query to metadata and add 3 tests

zhuoran-Cheng16:
### Summary

- Add min_issue query by using JOIN to combine with original inner_sql, modified 3 test file by adding some new test values without the is_latest_issue to test if it can get the correct min_issue. Already pass all the unit test and integration test.

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -346,20 +346,18 @@ class Database:
         print("no jobs left, thread terminating: " + threading.current_thread().name)
       finally:
         worker_dbc.disconnect(False) # cleanup
-      
 
     threads = []
     for n in range(n_threads):
       t = threading.Thread(target=worker, name='MetacacheThread-'+str(n))
       t.start()
-      threads.append(t)      
+      threads.append(t)
 
     srcsigs.join()
     print("jobs complete")
     for t in threads:
       t.join()
     print("threads terminated")
-
     # sort the metadata because threaded workers dgaf
     sorting_fields = "data_source signal time_type geo_type".split()
     sortable_fields_fn = lambda x: [(field, x[field]) for field in sorting_fields]
@@ -367,11 +365,7 @@ class Database:
     tuple_representation = list(map(prepended_sortables_fn, meta))
     tuple_representation.sort()
     meta = list(map(dict, tuple_representation)) # back to dict form
-
     return meta
-
-
-    
 
 
   def update_covidcast_meta_cache(self, metadata):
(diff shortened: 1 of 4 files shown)


src/acquisition/covidcast/database.py
```diff
@@ -270,36 +270,60 @@ def compute_covidcast_meta(self, table_name='covidcast', use_index=True):
...
-        MIN(`value`) AS `min_value`,
-        MAX(`value`) AS `max_value`,
-        ROUND(AVG(`value`),7) AS `mean_value`,
-        ROUND(STD(`value`),7) AS `stdev_value`,
-        MAX(`value_updated_timestamp`) AS `last_update`,
-        MAX(`issue`) as `max_issue`,
-        MIN(`lag`) as `min_lag`,
-        MAX(`lag`) as `max_lag`
+      SELECT 
+        inner_main_sql.*,inner_min_sql.*
```
krivard, line 274: do you need all the columns from inner_min_sql to be returned? if not, you can do
```suggestion
        inner_main_sql.*,inner_min_sql.min_issue     
```
-- a column doesn't have to be in the outer SELECT to be used in JOIN .. ON

src/acquisition/covidcast/database.py
```diff
@@ -312,7 +336,7 @@ def worker():
       try:
         while True:
           (source, signal) = srcsigs.get_nowait() # this will throw the Empty caught below
-          w_cursor.execute(inner_sql, (source, signal))
+          w_cursor.execute(inner_sql, (source, signal,source, signal))
```
krivard, line 339:
```suggestion
          w_cursor.execute(inner_sql, (source, signal, source, signal))
```

krivard (commented):
Final approval pending scalability results, but in the meantime we can probably cut the number of columns returned in half

src/acquisition/covidcast/database.py
zhuoran-Cheng16: I will fix that problem soon! Thanks for pointing out.

krivard:
Summary from offline discussion -- 

this is not as performant as we'd like on dense datasets like doctor-visits and jhu-csse, and splitting the JOIN into two queries doesn't seem to help. instead, we'd like to pursue an approach based on the incremental metadata PR (#375), in which we would figure out the min issue once for everything, then update it at _acquisition time_ to account for new earlier issues (as batch uploads) or new source+signal pairs (as part of initial metadata estimates).

we'll return to this PR to either modify or close it once we've done some more work on the incremental meta code.

krivard:
wontfix; invalidated by v4

PR #622, opened 2021-06-30
fix logger bug for batch issue upload

TaraLakdawala:
change logger.warning to logger.info to avoid error

still need to add/update unit and integration tests to catch this issue in the future

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -95,7 +95,7 @@ class CsvImporter:
           logger.info('processing csv files from issue date: "' + str(issue_date) + '", directory', path)
           yield from CsvImporter.find_csv_files(path, issue=(issue_date, epi.Week.fromdate(issue_date)), glob=glob)
         else:
-          logger.info('invalid issue directory day', issue_date_value)
+          logger.warning('invalid issue directory day', issue_date_value)
 
   @staticmethod
   def find_csv_files(scan_dir, issue=(date.today(), epi.Week.fromdate(date.today())), glob=glob):
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -88,7 +88,7 @@ class CsvImporter:
     logger = get_structured_logger('find_issue_specific_csv_files')
     for path in sorted(glob.glob(os.path.join(scan_dir, '*'))):
       issuedir_match = CsvImporter.PATTERN_ISSUE_DIR.match(path.lower())
-      if issuedir_match and glob.glob(os.path.isdir(path)):
+      if issuedir_match:
         issue_date_value = int(issuedir_match.group(2))
         issue_date = CsvImporter.is_sane_day(issue_date_value)
         if issue_date:
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -92,7 +92,7 @@ class CsvImporter:
         issue_date_value = int(issuedir_match.group(2))
         issue_date = CsvImporter.is_sane_day(issue_date_value)
         if issue_date:
-          logger.info(event='processing csv files from issue', detail=issue_date, file=path)
+          logger.info(event='processing csv files from issue', detail=issue_date, path)
           yield from CsvImporter.find_csv_files(path, issue=(issue_date, epi.Week.fromdate(issue_date)), glob=glob)
         else:
           logger.warning(event='invalid issue directory day', detail=issue_date_value, file=path)
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -88,11 +88,11 @@ class CsvImporter:
     logger = get_structured_logger('find_issue_specific_csv_files')
     for path in sorted(glob.glob(os.path.join(scan_dir, '*'))):
       issuedir_match = CsvImporter.PATTERN_ISSUE_DIR.match(path.lower())
-      if issuedir_match and os.path.isdir(path):
+      if issuedir_match and glob.glob(os.path.isdir(path)):
         issue_date_value = int(issuedir_match.group(2))
         issue_date = CsvImporter.is_sane_day(issue_date_value)
         if issue_date:
-          logger.info(event='processing csv files from issue', detail=issue_date, path)
+          logger.info(event='processing csv files from issue', detail=issue_date, file=path)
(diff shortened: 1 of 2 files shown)


krivard:
The line referred to in the error message is actually this one:

https://github.com/cmu-delphi/delphi-epidata/blob/00e6b805acab6200798541e639a5db9b384d44a7/src/acquisition/covidcast/csv_importer.py#L95

but it looks like the line you're currently repairing has the same problem. So probably just extend the fix to both locations? and hopefully the test you write will catch either of them.

TaraLakdawala:
Not sure how I missed that! I've extended the fix to line 95 now

TaraLakdawala:
Not sure how I missed that, but I've extended the fix to line 95. Hopefully, that should cover everything!

Testing the logger is turning out to be a bit of a challenge, so we will open a separate issue to track that.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -92,10 +92,10 @@ def find_issue_specific_csv_files(scan_dir, glob=glob):
         issue_date_value = int(issuedir_match.group(2))
         issue_date = CsvImporter.is_sane_day(issue_date_value)
         if issue_date:
-          logger.info('processing csv files from issue date: "' + str(issue_date) + '", directory', path)
+          logger.info(event='processing csv files from issue', detail=issue_date, file=path)
           yield from CsvImporter.find_csv_files(path, issue=(issue_date, epi.Week.fromdate(issue_date)), glob=glob)
         else:
-          logger.warning(event='invalid issue directory day', detail=issue_date_value, file=path)
+          logger.info(event='invalid issue directory day', detail=issue_date_value, file=path)
```
krivard, line 98: why is this changed to info?

zhuoran-Cheng16: when we use warning with the structure logging previously like this:
`logger.warning(event='invalid issue directory day', detail=issue_date_value, file=path)`
 the unit test will not pass.

zhuoran-Cheng16: My fault, I change it back.

src/acquisition/covidcast/csv_importer.py
```diff
@@ -88,11 +88,11 @@ def find_issue_specific_csv_files(scan_dir, glob=glob):
     logger = get_structured_logger('find_issue_specific_csv_files')
     for path in sorted(glob.glob(os.path.join(scan_dir, '*'))):
       issuedir_match = CsvImporter.PATTERN_ISSUE_DIR.match(path.lower())
-      if issuedir_match and os.path.isdir(path):
+      if issuedir_match:
```
krivard, line 91: interesting -- what will happen if `path` is a text file called `issue_12345678`? will the `find_csv_files` call on line 96 handle that gracefully, or crash?

zhuoran-Cheng16: I take a try on this approach by picking up the issue_12345 by a text filed called issue_12345, here is some code of my approach: 
```python
path='prefix/to/the/data/issue_20200408.csv' 
issuedir_match = CsvImporter.PATTERN_ISSUE_DIR.match(('.').join(path.split('.')[:-1]))
```
it will pass on the tests but will fail on integration tests. Because the change of form of issue_path will effect file load_sensors.py,  so it will need some extra changes if I want to use the text file form as issue_123456.csv.

krivard: sorry, i meant that we want the following:
1. if someone mistakenly creates a file which matches `CsvImporter.PATTERN_ISSUE_DIR` but is _not_ a directory
2. and we try to load the directory containing that bad file as a batch import
3. then the code should ignore the bad file

This was originally accomplished using the `and os.path.isdir(path)` clause. Without that clause, will the code crash in the above scenario?

zhuoran-Cheng16: Gotcha! I previously hadn't make dir path to a valid dir so the test won't pass. I already add back the check code and add one more test of the invalid dir path.  One more question, can I know why in the function` find_csv_files`  check the csv path(the path format) instead to check the dir path? Should I add this`os.path.isdir(path)` also in the function` find_csv_files`?

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -42,6 +42,35 @@ def test_is_sane_week(self):
     self.assertFalse(CsvImporter.is_sane_week(202054))
     self.assertFalse(CsvImporter.is_sane_week(20200418))
 
+  def test_find_issue_specific_csv_files(self):
+      """Recursively explore and find issue specific CSV files."""
+      # valid path
+      path_prefix='prefix/to/the/data/issue_20200408'
+      os.makedirs(path_prefix, exist_ok=True)
+      self.assertTrue(os.path.isdir(path_prefix))
```
krivard, line 50: see if you can do this with [`unittest.mock.patch`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) instead -- maybe something like

```
@patch("os.path.isdir")
def test_find_issue_specific_csv_files(self, os_isdir_mock):
    ...
    os_isdir_mock.return_value = True
    ...
```

src/acquisition/covidcast/csv_importer.py
krivard: both `find_issue_specific_csv_files` and `find_csv_files` check that the paths they produce match a regular expression because the regex is used to extract information from the path (issue, source, time_value, geo_type, and signal). 

`find_issue_specific_csv_files` checks that the paths it produces are directories, because we've fed it bad input before by mistake 😉 . It doesn't get run very often, and each run has to be configured by hand, which means it's easy to accidentally do it wrong.

We run `find_csv_files` 24 times a day, so if we ever _did_ accidentally feed it a filesystem that had directories with names that should have been normal files, or normal files with names that should have been directories, we would have fixed it in early summer 2020. All the things that feed `find_csv_files` are automated and less susceptible to human error.

krivard (dismissed):
🎉

dev/docker/python/Dockerfile
```diff
@@ -1,4 +1,8 @@
 # start with the `delphi_python` image
 FROM delphi_python
 
+# copy over all source files
+COPY repos repos
+RUN chmod -R o+r repos/
```
krivard, line 6: Hey @sgratzl I added this to the Dockerfile to avoid having to repeatedly rebuild both `delphi_python` and `delphi_web_python` during development. Will this break CI or can I leave it in?

sgratzl: why don't you just add this as a volume into the docker container? then you just have to restart the container. At least that is what I'm doing.

krivard (approved):
still good, just dropped the unnecessary docker edit 👍

PR #632, opened 2021-07-12
Mention Cyberduck problem and solution

capnrefsmmat:
Documents the solution to a common Cyberduck issue experienced by data users when trying to download large files.


krivard:
we're doing a delphi-epidata release tomorrow morning. if you need the docs published earlier you may need to cherry-pick them into main

capnrefsmmat:
Tomorrow morning should be fine.

PR #643, opened 2021-07-15
Increase gunicorn timeout

korlaxxalrok:
**Prerequisites**:

- [X] Build is successful

### Summary

Increases gunicorn's worker timeout to 300s (5min).

It looks like the default of 30s is too short and workers were probably being
killed prematurely. My admittedly somewhat limited testing on a single instance
running a manually edited version of the config file suggests increasing the
timeout will help.


krivard (approved):
yolo

PR #599, opened 2021-06-16
Feature: Add streaming smoothing and cumulative/incidence derivations to server

dshemetov:
Fixes #607 
Fixes #608 

**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [ ] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [ ] Build is successful
- [ ] Code is cleaned up and formatted

### Summary

Adds support for streaming 7-day averaging and differencing calculations to the Flask server. Relies on #564 for signal name aliasing.

The following signals are currently supported (based on the recent [meta-table](https://docs.google.com/spreadsheets/d/1zb7ItJzY5oq1n-2xtvnPBiJu2L3AqmCKubrLkKJZVHs/edit#gid=329338228)):

```
google-symptoms	ageusia_raw_search	TRUE	ageusia_smoothed_search
google-symptoms	anosmia_raw_search	TRUE	anosmia_smoothed_search
google-symptoms	sum_anosmia_ageusia_raw_search sum_anosmia_ageusia_smoothed_search

jhu-csse	confirmed_incidence_num	TRUE	confirmed_7dav_cumulative_num
jhu-csse	confirmed_incidence_num	TRUE	confirmed_7dav_incidence_num
jhu-csse	confirmed_incidence_num	TRUE	confirmed_cumulative_num
jhu-csse	deaths_incidence_num	TRUE	deaths_7dav_cumulative_num
jhu-csse	deaths_incidence_num	TRUE	deaths_7dav_incidence_num
jhu-csse	deaths_incidence_num	TRUE	deaths_cumulative_num

usa-facts	confirmed_incidence_num	TRUE	confirmed_7dav_cumulative_num
usa-facts	confirmed_incidence_num	FALSE	confirmed_7dav_incidence_num
usa-facts	confirmed_incidence_num	FALSE	confirmed_cumulative_num
usa-facts	deaths_incidence_num	TRUE	deaths_7dav_cumulative_num
usa-facts	deaths_incidence_num	FALSE	deaths_7dav_incidence_num
usa-facts	deaths_incidence_num	FALSE	deaths_cumulative_num
```

There may be others, but these were immediately apparent.

TODO:
- [x] figure out how to use the name aliasing feature and handle "source:*" requests
- [x] choose the row ordering as they are streamed from the database
- [ ] make sure "issue" and "as_of" params are handled
[...]

PR #646, opened 2021-07-15
[JIT]: just-in-time series computations

### LINKED ISSUE #607: Compute 7-day averages at query time in covidcast
For every signal X with a `7dav` variant _that does not include low-counts aggregands not present in X_, we should only store X in the database, and compute the `7dav` signal at query time.

### LINKED ISSUE #608: Compute incidence at query time
For every cumulative signal X with an `incidence` variant, we should only store X in the database, and compute the `incidence` signal at query time.

dshemetov:
[PRD for this work](https://docs.google.com/document/d/1CX7TcmDxpAfdKR0SbT31BIeH-8xdZA-rXut2nWI96Lk/edit)

- Remake of #599.
- Fixes #607 
- Fixes #608 

Built on #1045.

**Prerequisites**:

- [x] Build is successful

### Summary

Adds support for JIT streaming 7-day averaging and differencing calculations to the API server.

TODO:
- [x] rebase to `dev` after #954
- [x] figure out how to use the name aliasing feature and handle "source:*" requests
- [x] handle multiple signal requests
- [x] reimplement derivable signal logic: add a "derivable" column to the meta-table, use signal base for the "derivable" signal, and use "is_smoothed", "is_cumulative" to determine which transformations to apply
- [x] change the base signals in the db / meta table from incidence to cumulative
- [x] write integration tests
  - [x] add an integration test with multiple geos
  - [x] add integration test with geo:* option
- [x] add old API compatibility mode 
- [x] update buffer to be cleaner and more memory efficient
- [x] ~~add smoother params handling~~ more features in a future update
- [x] add time-padding fetch logic
- [x] add time-gap handling (likely nan filling at first)
- [x] handle "issue", "as_of", and "lag" params - "issue" and "lag" are not supported, "as_of" is supported naturally
- [x] handle other endpoints
  - [x] /trend
  - [x] /trendseries
  - [x] ~~/correlation~~ (unavailable bc streaming is defeated by full-length time-series computations)
  - [x] /csv
  - [x] ~~/backfill~~ (multiple problems)
  - [x] /meta work in #995  
  - [x] ~~/coverage~~ (out of scope)
  - [x] ~~/anomalies~~ (only does a sheet lookup)


integrations/server/test_covidcast_endpoints.py
```diff
@@ -164,6 +269,64 @@ def test_basic(self):
...
+
+        rows = [CovidcastRow(source="jhu-csse", signal="confirmed_unknown", time_value=20200401 + i, value=i) for i in range(10)]
+        first = rows[0]
+        self._insert_rows(rows)
+
+        with self.subTest("unknown signal"):
+            out = self._fetch("/", signal="jhu-csse:confirmed_unknown", geo=first.geo_pair, time="day:*")
+            self.assertEqual(len(out["epidata"]), len(rows))
+            out_values = [row["value"] for row in out["epidata"]]
+            expected_values = [float(i) for i in range(10)]
```
sgratzl, line 285: why not using the rows object and their values directly?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 Flask==1.1.2
 SQLAlchemy==1.3.22
+more_itertools
```
sgratzl, line 3: prefer specifying the version for reproducibility

src/server/endpoints/covidcast.py
```diff
@@ -122,16 +122,61 @@ def guess_index_to_use(time: List[TimePair], geo: List[GeoPair], issues: Optiona
     return None
 
 
+# TODO: Write an actual smoother arg parser.
```
sgratzl, line 125: TODO

src/server/endpoints/covidcast.py
```diff
@@ -122,16 +122,61 @@ def guess_index_to_use(time: List[TimePair], geo: List[GeoPair], issues: Optiona
...
+
+def _get_transformed_row_generator(
+    parsed_rows: Iterable[Dict],
+    iterator_buffer: Generator[Iterable[Tuple[Dict, Union[Tuple[str, str], str]]], None, None],
+    is_compatibility: bool,
+    original_source: Optional[str] = None,
+    smoother_args: Dict = {},
+    rowsort_keyfunc: Optional[Callable] = None,
+    group_keyfunc: Optional[Callable] = None,
+    data_signals_by_key: Dict[Tuple[str, str], DataFrame] = data_signals_by_key,
```
sgratzl, line 143: DataFrame?

src/server/endpoints/covidcast.py
```diff
@@ -122,16 +122,61 @@ def guess_index_to_use(time: List[TimePair], geo: List[GeoPair], issues: Optiona
...
+    for x, _ in iter:
+        yield x
+
+
+def _get_transformed_row_generator(
+    parsed_rows: Iterable[Dict],
+    iterator_buffer: Generator[Iterable[Tuple[Dict, Union[Tuple[str, str], str]]], None, None],
+    is_compatibility: bool,
+    original_source: Optional[str] = None,
+    smoother_args: Dict = {},
```
sgratzl, line 140: optional args with dict/lists shouldn't be used in my knowledge since it won't create new instances

src/server/endpoints/covidcast.py
```diff
@@ -122,16 +122,61 @@ def guess_index_to_use(time: List[TimePair], geo: List[GeoPair], issues: Optiona
...
+    rowsort_keyfunc: Optional[Callable] = None,
```
sgratzl, line 141: can you specify the Callable args?

src/server/endpoints/covidcast.py
```diff
@@ -140,18 +185,20 @@ def handle():
     fields_int = ["time_value", "direction", "issue", "lag", "missing_value", "missing_stderr", "missing_sample_size"]
     fields_float = ["value", "stderr", "sample_size"]
     is_compatibility = is_compatibility_mode()
+
     if is_compatibility:
-        q.set_order("signal", "time_value", "geo_value", "issue")
+        original_source = source_signal_pairs[0].source
```
sgratzl, line 190: an alternative approach is to also return query the source field in the fields but the filter out the fields you don't want in the compatibility mode before yielding it

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -272,3 +283,128 @@ def map_row(source: str, signal: str) -> str:
         return signal_source.source
 
     return transformed_pairs, map_row
+
+
+def _resolve_all_signals(
+    source_signals: Union[SourceSignalPair, List[SourceSignalPair]], data_sources_by_id: DataFrame
```
sgratzl, line 289: DataFrame?

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -272,3 +283,128 @@ def map_row(source: str, signal: str) -> str:
...
+    # First iterator pass.
+    for x in iter:
+        key = keyfunc(x)
+        if name_dict.get(key):
+            buffer.setdefault(key, []).append(x)
+            continue
+        yield x, key
+
+    # Buffer pass as needed.
+    for key in buffer:
```
sgratzl, line 334: consider .items()

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -272,3 +283,128 @@ def map_row(source: str, signal: str) -> str:
...
+            if source:
+                return SourceSignalPair(source.source, [s.signal for s in source.signals])
+        return source_signals
+    if isinstance(source_signals, list):
+        return [_resolve_all_signals(pair, data_sources_by_id) for pair in source_signals]
+    raise TypeError("source_signals is not Union[SourceSignalPair, List[SourceSignalPair]].")
+
+
+def _buffer_and_tag_iterator(
+    iter: Iterable[Dict], name_dict: Dict[Any, List[Any]], keyfunc: Callable[[Dict], Any]
```
sgratzl, line 303: consider using generics instead of the Any to better indicate the connection between the different arguments

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -272,3 +283,128 @@ def map_row(source: str, signal: str) -> str:
...
+    # Buffer pass as needed.
+    for key in buffer:
+        for value in name_dict[key]:
+            for x in buffer[key]:
+                yield x, value
+
+
+def _get_basename_signals(
+    source_signals: List[SourceSignalPair],
+    data_sources_by_id: Dict[str, DataFrame] = data_sources_by_id,
```
sgratzl, line 342: DataFrame?

tests/server/test_params.py
```diff
@@ -357,3 +357,29 @@ def test_parse_day_arg(self):
                 self.assertRaises(ValidationFailedException, parse_day_arg, "time")
             with app.test_request_context("/?time=week:20121010"):
                 self.assertRaises(ValidationFailedException, parse_day_arg, "time")
+
+    def test__combine_source_signal_pairs(self):
+        source_signal_pairs = [
+            SourceSignalPair("src1", ["sig1", "sig2"]),
+            SourceSignalPair("src2", "sig1"),
```
sgratzl, line 364: does the typing https://github.com/cmu-delphi/delphi-epidata/blob/5a2d6b49214d3584eeec7e2e1cbf0ba6b75e0b9a/src/server/_params.py#L82 actually allow a single string?

src/server/endpoints/covidcast_utils/smooth_diff.py
```diff
@@ -0,0 +1,168 @@
...
+            raise ValueError
+        values = array(values, copy=False)
+        smoothed_value = dot(values, kernel)
+    except (ValueError, TypeError):
+        smoothed_value = array(values).mean()
+
+    return round(smoothed_value, 8)
+
+
+def _iso_int_to_date(di: int) -> date:
```
sgratzl, line 31: isn't there a utility function for that?

src/server/endpoints/covidcast_utils/smooth_diff.py
```diff
@@ -0,0 +1,168 @@
...
+    s = str(di)
+    year, month, day = int(s[:4]), int(s[4:6]), int(s[6:8])
+    return date(year, month, day)
+
+
+def _date_to_iso_int(dt: date) -> int:
+    return int(str(dt.year) + str(dt.month).zfill(2) + str(dt.day).zfill(2))
+
+
+def _pad_group(group: Iterable[Dict], pad_length: int, pad_fill_value: Union[Number, str] = nan) -> Iterable[Dict]:
```
sgratzl, line 41: have you considered using an enum instead of a string which could be anything?

src/server/endpoints/covidcast_utils/smooth_diff.py
```diff
@@ -0,0 +1,168 @@
...
+    elif isinstance(pad_fill_value, Number):
+        padded_values = (dict(ChainMap({"time_value": _date_to_iso_int(date), "value": pad_fill_value}, start_entry)) for date in date_range(left_bound, right_bound))
+    else:
+        raise ValueError("pad_fill_value should be a float or 'first' or 'drop'.")
+    return chain(padded_values, group)
+
+
+def generate_smooth_rows(
+    rows: Iterable[Dict],
+    smoother_kernel: Union[str, List[Number]] = "average",
```
sgratzl, line 67: have you considered using an enum instead of a string which could be anything?

src/server/endpoints/covidcast_utils/smooth_diff.py
sgratzl: another option would be to enforce the structure of the dict (time_value, value) using TypedDicts

sgratzl (commented):
what is your plan for the other covidcast endpoints? like /meta /trend /backfill, ....

dshemetov:
@sgratzl I haven't thought it too much, but I wanted to redirect a call to /trend to / and then grab the output from there. Not sure if that's possible with Flask.

integrations/server/test_covidcast_endpoints.py
dshemetov: Do you mean to compute the expected_values via differences?

tests/server/test_params.py
dshemetov: ah, good point. could've sworn it did!

src/server/endpoints/covidcast_utils/smooth_diff.py
dshemetov: All I could find was Python's [datetime.fromisoformat](https://docs.python.org/3/library/datetime.html?highlight=datetime#datetime.date.fromisoformat) but it expects dashes in the string. I saw you wrote some time parsers, but they only converted between str and int.

src/server/endpoints/covidcast_utils/smooth_diff.py
dshemetov: made some updates to use Enums there

src/server/endpoints/covidcast_utils/smooth_diff.py
sgratzl: what about https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/utils/dates.py

integrations/server/test_covidcast_endpoints.py
sgratzl: I mean when you were creating the database entries you were using a list of CovidcastEntry objects. You could use these object also to verify the output. Thus, you don't have to use internal knowledge (like the range(10) thing) to verify the output

sgratzl:
another question regarding smoothing: when I request a smoothed signal starting at 2021-01-10 how will the first entry looks like? which unsmoothed values are used for computing the first smoothed value?

sgratzl:
> @sgratzl I haven't thought it too much, but I wanted to redirect a call to /trend to / and then perform the trend calculations on the output there. Not sure if that's possible with Flask.

that won't really work since some of the endpoints use specialized SQL queries and don't operate on the normal data returned by the regular API endpoint, e.g. the backfill one as an extreme case.

integrations/server/test_covidcast_endpoints.py
dshemetov: got it

src/server/endpoints/covidcast_utils/smooth_diff.py
dshemetov: ah i didnt see these, thanks

dshemetov:
> another question regarding smoothing: when I request a smoothed signal starting at 2021-01-10 how will the first entry looks like? which unsmoothed values are used for computing the first smoothed value?

I wrote a couple options for this. By default, smoothing will use a dynamic window, e.g. 2021-01-10 will use just the same value from that day, 2021-01-11 will average that day's value and the previous, and so on up until the window length. If pad_fill_value is given (floats and nans allowed), then (window_length - 1) many pad_fill_values will be appended prior to 2021-01-10 and used for averaging. If a smoother kernel is provided (i.e. a list of floats), then padding has similar logic, but weighted averaging is used (this is assuming we can find a way to send a list of floats in an HTTP query string.)

sgratzl:
> In both diffing and smoothing, I am currently setting stderr and sample_size to None. Obtaining those from a group of values seems difficult.

but we have the previous 7 days in the database. The values shouldn't change for 2021-01-10 regardless whether I request the timeframes 2021-01-03--2021-01-12, 2021-01-07--2021-01-12, or 2021-01-10--2021-01-12. as far as I understand they would in the current implementation, wouldn't they?

dshemetov:
> > @sgratzl I haven't thought it too much, but I wanted to redirect a call to /trend to / and then perform the trend calculations on the output there. Not sure if that's possible with Flask.
> 
> that won't really work since some of the endpoints use specialized SQL queries and don't operate on the normal data returned by the regular API endpoint, e.g. the backfill one as an extreme case.

Hmm, I will need to learn more about the other endpoints and get back with some ideas.

dshemetov:
> but we have the previous 7 days in the database. The values shouldn't change for 2021-01-10 regardless whether I request the timeframes 2021-01-03--2021-01-12, 2021-01-07--2021-01-12, or 2021-01-10--2021-01-12. as far as I understand they would in the current implementation, wouldn't they?

You're right, they would change. I didn't want to complicate the smoothing logic on this first pass implementation. And I think it can get quite complicated. If we want the smoothed values to not change, then we could modify the time range of the request to make the calculation. 

> but we have the previous 7 days in the database

One complication is that this won't always be true: signals could have outages or could be requesting data on the "left" boundary. The decision to pad would then need to be made when you first see the row value with the earliest timestamp - if the smoothing window is not covered with data, then no padding is necessary.

dshemetov:
@sgratzl here are my notes for the other endpoints, pls let me know if I misread the code.

- /trend and /trendseries can be either rerouted to / or have the same `_get_transformed_row_generator` code as / inserted before trend calculations
- /correlation can do the same
- /csv I don't understand this endpoint; what is the goal here? why is the param parsing different? I assume it's doing something more than just returning a CSV version of the / endpoint response?
- /backfill it appears that this endpoint returns all the issues in a given specified range; I'm not sure that we can support this case for all time-series computations, more on that below*.
- /meta this requires reporting max/min statistics on all the signals; if we shift over to dropping smoothed/incidence signal from the table, this would require performing a full time-series calculation for each derived signal, which would be really expensive; one option would be to perform this computation once daily and cache it, otherwise we just might not support it
- /coverage this requires reporting geo-coverage statistics on all the signals; my thinking is similar to the /meta endpoint

*Say we have a query that returns multiple rows with the same (geo, source, signal, time_value). It is not clear to me how to order this in a way that would allow for sensible computations that rely on time-ordering. For example, suppose we have a query that returns the rows below.

id|geo|source|signal|value|time_value|issue
-|-|-|-|-|-|-
1|01234|"src"|"sig"|5|20210701|20210702
2|01234|"src"|"sig"|7|20210702|20210702
3|01234|"src"|"sig"|6|20210701|20210703

I don't know how to handle smoothing or diffing in this case, because the value "previous to" 20210702 is multiply defined. To be fair, one option is to report every possible combination (i.e. average rows 2 and 1 together and 2 and 3 together, using the max issue of the involved rows). However, this will not be doable in a streaming fashion due to the ordering of the "issue" column. These considerations lead me to conclude that derived signals should not be supported for requests with the issue parameter (note that as_of can still be supported).

krivard:
> reporting max/min statistics on all the signals; if we shift over to dropping smoothed/incidence signal from the table, this would require performing a full time-series calculation for each derived signal, which would be really expensive; one option would be to perform this computation once daily and cache it, otherwise we just might not support it

two alternatives:
* compute variant stats at acquisition time, since [we're reading it into a pandas data frame anyhow](https://github.com/cmu-delphi/delphi-epidata/blob/2e62d6c336768659f0d34962c982a3932d6c1569/src/acquisition/covidcast/csv_importer.py#L337)
* revert variant stats to base stats & include a warning message that this was done

krivard:
> These considerations lead me to conclude that derived signals should not be supported for requests with the issue parameter

I we desperately wanted single issue, we could do the following:
* Run a database query as-of the requested issue
* Run a second database query as-of the previous issue
* Generate a derived stream for each
* Only return rows where the derived data for as-of the requested issue is different from the derived data for as-of the previous issue

it's utter 🗑️  and I have no idea how you'd stream it, but not impossible. definitely wait and see if we have a compelling use case request first.

krivard:
& ftr I'm okay with making the backfill interface unavailable for derived signals

dshemetov:
> > reporting max/min statistics on all the signals; if we shift over to dropping smoothed/incidence signal from the table, this would require performing a full time-series calculation for each derived signal, which would be really expensive; one option would be to perform this computation once daily and cache it, otherwise we just might not support it
> 
> two alternatives:
> 
> * compute variant stats at acquisition time, since [we're reading it into a pandas data frame anyhow](https://github.com/cmu-delphi/delphi-epidata/blob/2e62d6c336768659f0d34962c982a3932d6c1569/src/acquisition/covidcast/csv_importer.py#L337)
> * revert variant stats to base stats & include a warning message that this was done

It occurs to me that these statistics may vary depending on the parameters given to the computation - i.e. max for smoothing with which kernel? with which window length? max for diffing with what boundary conditions? with what nan filling choice? We could provide the statistics only for the default settings.

dshemetov:
Thinking more about correlations computations, those would be hard to do in a streaming fashion. But it looks like our current implementation of correlations just loads the query into a dataframe anyway.

krivard:
is there already an upper limit on the number of points for correlations? if not we should add one

dshemetov:
I don't see one. Btw, [`pd.read_sql_query`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_sql_query.html#pandas.read_sql_query) seems very relevant. @sgratzl do you know how intelligent this function is about doing lazy evaluations? Depending on if it can handle streaming or not, it could obsolete my custom-brew code. Which would be simultaneously a bummer and a relief?

krivard:
@dshemetov would you switch this to draft until it's ready to review?

dshemetov:
@krivard so trying to get this into an MVP state I need to handle a few more endpoints. My thinking is: we can handle trend, trendseries, and possibly csv with redirects to `/`. For `/meta` and `/coverage`, the simplest thing to do now would be to not report the smoothed / diffed signals. We can add periodic caching of the results or another solution at a later time. What do you think?

krivard:
> the simplest thing to do now would be to not report the smoothed / diffed signals

I'm okay with calling `/meta` and `/coverage` out of scope for this PR.

That will, however, affect rollout. Before we delete the derivable cases/deaths signals from the database, we need to check how www-covidcast determines valid dates for cases/deaths. If www-covidcast expects meta information for the derived signals, then we'll need to put in a shim first depending on what exactly the constraints are;
* If www-covidcast just needs min and max time_value for all variants, then we can amend metadata at query time just by copying the entry from the base signal. 
* If it needs legit mean/min/max/stdev value, that gets trickier, but in that case we probably just do a post-pass during the meta update to fire up an API client and request the derived time series. 

Regardless, out of scope for this PR, and possibly for the MVP in general

melange396 (dismissed):
there are a few things to double check, but overall it looks pretty damn good to me.  whew, sorry that took so long!  my excuse is that theres a lot here and there are some convoluted patterns scattered in it.  im not in love with a lot of the style and paradigms used here, but youve just echoed that of the surrounding code.  it makes sense from a maintainability point of view, so i really cant complain.  you also included excellent docstrings to make sense of the complicated routines which helps assuage the cringe i feel when work with this part of the codebase.  there is a long rant in there, im not sure why i put so much into it, but whatevs

dshemetov:
Thank you for the thorough review @melange396! I'm grateful to get a thorough check from someone experienced, since this is my first release that touches a major artery of our API, it's admittedly complicated, and I am a bit nervous that I overlooked something.

I'm currently prioritizing some modeling work, but will chip away slowly at your comments and suggestions over the next week.

dshemetov:
Merged #903 and latest `dev` in.

dshemetov:
Made a few new updates. CI tests will likely be failing for the same reason that relative imports fail here #949. If we can fix those there, we can fix them here. Otherwise, the relative import refactor might need to wait for a fix for #939 (and related versions not just for `acquisition`).

src/server/endpoints/covidcast.py
```diff
@@ -216,8 +270,20 @@ def handle_trend():
...
+            transformed_rows = row_transform_generator(parsed_rows=parsed_rows, time_pairs=[TimePair("day", [time_window])], transform_args=transform_args)
+            for row in transformed_rows:
+                yield row
+    else:
+        def gen_transform(rows):
+            parsed_rows = (parse_row(row, fields_string, fields_int, fields_float) for row in rows)
+            for row in parsed_rows:
+                yield row
+
+    def gen_trend(rows):
```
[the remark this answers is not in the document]
dshemetov: don't expect a response since this review is almost a year old, but i think it is a [generator](https://docs.python.org/3/glossary.html#term-generator): 
> [A generator] looks like a normal function except that it contains [yield](https://docs.python.org/3/reference/simple_stmts.html#yield) expressions [...]

src/server/endpoints/covidcast.py
```diff
@@ -431,14 +539,15 @@ def gen(first_row, rows):
         raise DatabaseErrorException(str(e))
 
     # special case for no data to be compatible with the CSV server
-    first_row = next(r, None)
+    transformed_query = peekable(gen_transform(r))
```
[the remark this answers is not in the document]
dshemetov: made me so happy to find it! [`more_itertools`](https://pypi.org/project/more-itertools/) is super handy!

integrations/server/test_covidcast_endpoints.py
```diff
@@ -530,19 +530,51 @@ def test_correlation(self):
     def test_csv(self):
         """Request a signal the /csv endpoint."""
 
+        expected_columns = ["geo_value", "signal", "time_value", "issue", "lag", "value", "stderr", "sample_size", "geo_type", "data_source"]
```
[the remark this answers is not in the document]
dshemetov: it's reflecting the same arbitrary order specified in `parse_csv_row` ¯\(ツ)/¯

i didn't know if there was a reason for it to be that way, so i just left it alone (just in case someone had already built their CSV parsing from us to expect a certain row order; one of those things that was arbitrary during early early dev, but is now unfortunately locked in stone)

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -299,3 +311,263 @@ def map_row(source: str, signal: str) -> str:
...
+    Example:
+    [TimePair("day", [20210407])] with pad_length 6 would return [TimePair("day", [20210407]), TimePair("day", [(20210401, 20210407)])].
+
+    Used to pad the user DB query with extra days.
+    """
+    if pad_length < 0:
+        raise ValueError("pad_length should non-negative.")
+    if pad_length == 0:
+        return time_pairs.copy()
+    min_time = min(time_value if isinstance(time_value, int) else time_value[0] for time_pair in time_pairs if not isinstance(time_pair.time_values, bool) for time_value in time_pair.time_values)
```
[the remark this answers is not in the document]
dshemetov: Sometimes I feel so clever with one-liners and comprehensions

...and then sometimes I come back to my code a year later and I wonder how I ever thought it was clear xD

Thanks for catching!

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -299,3 +311,263 @@ def map_row(source: str, signal: str) -> str:
...
     return transformed_pairs, map_row
+
+
+def _resolve_all_signals(source_signals: Union[SourceSignalPair, List[SourceSignalPair]], data_sources_by_id: Dict[str, DataSource]) -> Union[SourceSignalPair, List[SourceSignalPair]]:
+    """Expand a request for all signals to an explicit list of signal names.
+
+    Example: SourceSignalPair("jhu-csse", signal=True) would return SourceSignalPair("jhu-csse", [<list of all JHU signals>]).
+    """
+    if isinstance(source_signals, SourceSignalPair):
+        if source_signals.signal == True:
```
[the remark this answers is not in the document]
dshemetov: I suppose it doesn't matter, because of the way `_parse_common_multi_arg` is coded, signal is a bool if and only if it is equal to True. 

This happens if someone makes a request for `signal=jhu-csse:*` (in theory, though currently the API seems to be spinning endlessly on this simple query `https://api.covidcast.cmu.edu/epidata/covidcast/?signal=jhu-csse:*&time=day:20220102&geo_type=state&geo_values=ca`)

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -299,3 +311,263 @@ def map_row(source: str, signal: str) -> str:
...
+        if not all(time_pair.time_type == "day" for time_pair in time_pairs):
+            raise ValueError("get_day_range only supports day time_type pairs.")
+        time_values = sorted(list(set().union(*(get_day_range(time_pair).time_values for time_pair in time_pairs))))
+        return TimePair("day", time_values)
+    else:
+        raise ValueError("get_day_range received an unsupported type as input.")
+
+
+def _generate_transformed_rows(
+    parsed_rows: Iterable[Dict], time_pairs: Optional[List[TimePair]] = None, transform_dict: Optional[Dict[Tuple[str, str], List[Tuple[str, str]]]] = None, transform_args: Optional[Dict] = None, group_keyfunc: Optional[Callable] = None, data_signals_by_key: Dict[Tuple[str, str], DataSignal] = data_signals_by_key,
```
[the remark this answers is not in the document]
dshemetov: So the whole relationship between generator, generator iterator, and iterator feels annoyingly murky to me, but I think I'm settling down on this interpretation of the docs:
- a [generator](https://docs.python.org/3/glossary.html#term-generator) is a function with a yield statement; when instantiated it returns a generator iterator
- a [generator iterator](https://docs.python.org/3/glossary.html#term-generator-iterator) is an iterator made by a generator function
- an [iterator](https://docs.python.org/3/glossary.html#term-iterator) is an object with a `__next__` method defined

So since `parsed_rows` is not a function that makes an iterator, but instead is the result of the [generator expression](https://docs.python.org/3/glossary.html#term-generator-expression) `parsed_rows = (parse_row(row, fields_string, fields_int, fields_float) for row in rows)` in `covidcast.py`, I'm going to mark this as `Iterator` and call it a day. Technically this function should work regardless of where parsed_rows came from, as long as it is an iterator that returns dicts (so it could work with a list or an iterator or a generator iterator).

I will mention in the docs though that we're assuming that `parsed_rows` is an iterable over rows from a query in the database.

dshemetov:
## A/B Testing Summary

I have completed a large batch of tests. Summary of findings:

1. JIT derived signals match exactly with Pandas-recomputed derived signals
2. JIT derived signals does not match exactly with live data, but in those cases
	- the live data contains bugs (e.g. possibly desync issues)
	- the live data contains artificial values (e.g. incidence earliest reference dates match cumulative, but it should be 1 day later)

(A **derived signal** is a signal obtained from a base signal via a transformation. A **base signal** is a signal that is not a derived signal.)

These findings give me confidence that we can roll this code out.

## Test Setup

### Data

I downloaded the latest data from:
- `sources=jhu-csse,usa-facts` 
- `signals=confirmed_cumulative_num,confirmed_incidence_num,confirmed_7dav_incidence_num`
- `geo_values=state:*,county:*` 
- `time_values=20200122-20220910`

### Test Procedure

The testing procedure went as follows:
1. I loaded all these signals into a local database
2. I obtained `confirmed_incidence_num` from JIT
	- I did this by querying the local API server using JIT (which computes the incidence values from `confirmed_cumulative_num` server-side)
3. I obtained `confirmed_incidence_num` from live data
	- I did this by querying the local API server *without* using JIT
4. I obtained `confirmed_incidence_num` by computing it from `confirmed_cumulative_num` with Pandas
	- I did this by reading the CSV file with `confirmed_cumulative_num` of all the USAFacts downloaded data into a dataframe, I reindexed it onto a contiguous time index, grouped by geo_value, and used [`diff`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.diff.html?highlight=diff#pandas.core.groupby.DataFrameGroupBy.diff) to take differences

### Test Comparisons and Results

I joined these three series into a single dataframe and compared:
1. I searched for all the rows where Pandas and JIT were different in the `value` column; I found 0 rows where they were different!
2. I searched for all the rows where JIT and live data were different and broke these down into two cases
	1. JIT has a nan and live data doesn't; in all these cases JIT was right to have a nan (there was a nan in the 2-day window for incidence or a nan in the 7-day window for smoothed incidence)
	2. JIT and live data both have value; since JIT matched with Pandas and live data didn't, this is an error in the live data (possibly a cache desync)

dshemetov:
Took a final rebase pass and cleaned the commit log
- all the codebase commits that aren't related to JIT are now accurately marked and frontloaded
- the JIT system itself is now in one big commit
  - I didn't find it helpful for readability to have it in many separate commits, especially given that most of the separate commits wouldn't run on their own

src/server/endpoints/covidcast.py
```diff
@@ -111,50 +123,115 @@ def _handle_lag_issues_as_of(q: QueryBuilder, issues: Optional[List[Union[Tuple[
...
     else:
-        # transfer also the new detail columns
-        fields_string.extend(["source", "geo_type", "time_type"])
-        q.set_order("source", "signal", "time_type", "time_value", "geo_type", "geo_value", "issue")
+        def gen_transform(rows):
+            parsed_rows = (parse_row(row, fields_string, fields_int, fields_float) for row in rows)
+            for row in parsed_rows:
+                yield alias_row(row)
+
+    q.set_order("geo_type", "geo_value", "source", "signal", "time_type", "time_value", "issue")
```
melange396, line 213: i think this change of ordering in the db query is causing the slowdown we saw in QA.  the previous order should be able to use an index to essentially get the ordering for free, but this one causes an extra sort step.  can we change it back and then ill run the locust tests again?

PR #659, opened 2021-07-26
Document declining survey response rate

nmdefries:
Added general overview of declining response rate/sample size in limitations section


capnrefsmmat (approved):
Looks good

PR #661, opened 2021-07-28
fix: handle cornercases where source is not transmitted for aliases

sgratzl:
**Prerequisites**:

- [x] Build is successful

### Summary

fix a cornercases in the alias resolver which didn't consider when the source is not about to be returned


krivard (approved):
Very tidy

PR #665, opened 2021-08-02
Signals dash generator: store a longer time series by using covidcast/coverage endpoint

krivard:
**Prerequisites**:

- [x] Build is successful

### Summary

Most of this PR is for switching the signals dashboard data generator over to the `covidcast/coverage` endpoint instead of pulling the regional data and counting sizes in a massive groupby. Since `coverage` isn't available natively in the covidcast library (and won't be; that's all going into the new delphi_epidata library/ies) we have to do this by constructing the URL, calling `pandas.read_csv`, and parsing the `time_value` column.

With that change, it makes it less of a big deal to double the length of the coverage time series -- the true goal of this PR.

### DIFF
--- a/src/acquisition/covidcast/signal_dash_data_generator.py
+++ b/src/acquisition/covidcast/signal_dash_data_generator.py
@@ -18,7 +18,7 @@ import delphi.operations.secrets as secrets
 from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
 
 
-LOOKBACK_DAYS_FOR_COVERAGE = 52
+LOOKBACK_DAYS_FOR_COVERAGE = 56
 BASE_COVIDCAST = covidcast.covidcast.Epidata.BASE_URL[:-len("api.php")] + "covidcast"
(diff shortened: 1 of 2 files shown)


src/acquisition/covidcast/signal_dash_data_generator.py
```diff
@@ -17,7 +18,9 @@
 from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
 
 
-LOOKBACK_DAYS_FOR_COVERAGE = 28
+LOOKBACK_DAYS_FOR_COVERAGE = 52
```
melange396, line 21: out of curiousity, why 52?

melange396 (approved):
LGTM

src/acquisition/covidcast/signal_dash_data_generator.py
krivard: uhhh because I can't multiple 28 by 2, apparently

krivard, line 21:
```suggestion
LOOKBACK_DAYS_FOR_COVERAGE = 56
```

PR #667, opened 2021-08-05
feat: move active flag to signal level instead of source level

### LINKED ISSUE #666: Permit advanced covidcast metadata to specify when some signals of an otherwise-active source are no longer being updated.
We currently track "active" status as a characteristic of the source. However, when sources have some signals that are still active, but others which have been discontinued, the metadata cannot help a downstream agent determine whether an individual signal is still being updated or not. For example, the `hospital-admissions` source is still updating `smoothed_covid19_from_claims` and `smoothed_adj_covid19_from_claims`, but not `smoothed_covid19` or `smoothed_adj_covid19`. [A query for meta data about the `smoothed_covid19` signal](https://api.covidcast.cmu.edu/epidata/covidcast/meta?signal=hospital-admissions:smoothed_covid19) claims it is active, but that is in reference to the overall source, not the signal requested.

Not sure how we want to resolve this -- we could:

* Split these sources into subdivisions, and throw all the inactive signals into a subdivision with active=False. Do something clever so that queries using the db source name get routed to the correct subdivision.
* Move active status from per-source to per-signal. Do something clever so that someone iterating over sources trying to only display the active ones doesn't have to iterate through all the child signals just in case.
* Keep active status per-source, but permit signals to override. Use the following rule: the first active=False wins.
* Something else

sgratzl:
closes #666

**Prerequisites**:

- [x] Build is successful

### Summary

update to the latest meta doc with the active flag moved to the signal level instead of source level


krivard (approved):
👍

PR #684, opened 2021-08-18
Add pairwise logging for meta

krivard:
Additional diagnostic information for 4-hour meta cache update runs the last few days.

**Prerequisites**:

- [x] Build is successful

### Summary
We're getting duplicate "thread started" logs for some reason, and the long runtimes seem to be because one of the threads takes ages to finish.

* Log number of threads before we start (to maybe see what is up with duplicates)
* Log after each pair is completed (i tried timing each pair but tests hung -- i suspect time() isn't thread-friendly)

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -315,13 +315,13 @@ class Database:
       try:
         while True:
           (source, signal) = srcsigs.get_nowait() # this will throw the Empty caught below
+          logger.info("starting pair", thread=name, pair=f"({source}, {signal})")
           w_cursor.execute(inner_sql, (source, signal))
           with meta_lock:
             meta.extend(list(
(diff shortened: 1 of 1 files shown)


melange396 (approved):
👍

src/acquisition/covidcast/database.py
```diff
@@ -319,8 +321,9 @@ def worker():
               dict(zip(w_cursor.column_names, x)) for x in w_cursor
             ))
           srcsigs.task_done()
+          logger.info("completed pair", thread=name, pair=f"({source}, {signal})")
```
melange396, line 324: this might be good to print before the sql is evaluated?

krivard: as in a "starting pair" instead?

PR #704, opened 2021-09-03
Limit rows returned for correlations

dshemetov:
Addresses this comment https://github.com/cmu-delphi/delphi-epidata/pull/646#issuecomment-889457081.

**Prerequisites**:

- [x] Build is successful

### Summary

Currently, the `/correlations` endpoint in covidcast bypasses the row limits requirements by passing the query directly to `pd.read_sql_query`. This PR abstracts row limiting from `run_query` and adds it to `as_pandas`. A test to make sure it works is provided. 

There may be a possible issue where someone requests too many rows and the row limit leads to the two correlated time series being mismatched, which would likely lead to an Exception later in the correlation code. We could try and send an error message if the number of returned rows == MAX_RESULTS instead of proceeding, getting an error, and sending nothing.

### DIFF
--- a/tests/server/test_pandas.py
+++ b/tests/server/test_pandas.py
@@ -6,14 +6,13 @@ from typing import Any, Dict, Iterable
 import unittest
 
 import pandas as pd
-from pandas.core.frame import DataFrame
-from sqlalchemy import text
-import mysql.connector
+from sqlalchemy import create_engine
 
 # from flask.testing import FlaskClient
 from delphi_utils import Nans
+from delphi.epidata.server.main import app
 from delphi.epidata.server._pandas import as_pandas
-from delphi.epidata.server._query import limit_query
+from delphi.epidata.server._query import QueryBuilder
 
 # py3tester coverage target
 __test_target__ = "delphi.epidata.server._query"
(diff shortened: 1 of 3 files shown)


tests/server/test_pandas.py
```diff
@@ -0,0 +1,176 @@
...
+        columns = [
+            'id', 'source', 'signal', 'time_type', 'geo_type', 'time_value',
+            'geo_value', 'value_updated_timestamp', 'value', 'stderr',
+            'sample_size', 'direction_updated_timestamp', 'direction', 'issue',
+            'lag', 'is_latest_issue', 'is_wip', 'missing_value', 'missing_stderr',
+            'missing_sample_size'
+        ]
+        return pd.DataFrame.from_records([[i] + row.aslist for i, row in enumerate(rows, start=1)], columns=columns)
+
+    def test_as_pandas(self):
```
krivard, line 159: given the name, I expected this test to call `as_pandas`, but it does not. on purpose?

dshemetov: I agree it's weird that I just duplicated the innards of `as_pandas` here, but I couldn't think of a way to pass in my own `engine` variable into it.

dshemetov: I modified this with some Flask testing magic, but now I'm running into an issue with SQLAlchemy 
```
    def as_pandas(query: str, params: Dict[str, Any], db_engine: Engine = engine, parse_dates: Optional[Dict[str, str]] = None, limit_rows = MAX_RESULTS+1) -> pd.DataFrame:
        try:
            query = limit_query(query, limit_rows)
            return pd.read_sql_query(text(str(query)), db_engine, params=params, parse_dates=parse_dates)
        except Exception as e:
>           raise DatabaseErrorException(str(e))
E           delphi.epidata.server._exceptions.DatabaseErrorException: 200 OK: database error: Execution failed on sql 'SELECT t.id, t.source, t.signal, t.time_type, t.geo_type, t.time_value, t.geo_value, t.value_updated_timestamp, t.value, t.stderr, t.sample_size, t.direction_updated_timestamp, t.direction, t.issue, t.lag, t.is_latest_issue, t.is_wip, t.missing_value, t.missing_stderr, t.missing_sample_size FROM covidcast t      LIMIT 10000001': Boolean value of this clause is not defined
```
`Boolean value of this clause is not defined` is pretty cryptic and the [documentation ](https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql) hasn't helped so far.

dshemetov: Turned out the issue was that the `mysql.connector` used in the tests is different from the `sqlalchemy.engine` used in the actual server and `pd.read_sql_query` is only happy with the latter.

krivard (approved):
👍

PR #674, opened 2021-08-06
[survey] Describe new lower-priority Wave 11 indicators for API and contingency

nmdefries:
### Description
Doc updates accompanying [[fb-package] Split `vaccine_barriers` indicator into vaccinated and tried versions](https://github.com/cmu-delphi/covidcast-indicators/pull/1147), [[fb package] Add a bunch more Wave 11 indicators to API](https://github.com/cmu-delphi/covidcast-indicators/pull/1153), and [[fb-survey] Expand contingency with lower-priority Wave 11 indicators](https://github.com/cmu-delphi/covidcast-indicators/pull/1175).

PR #706, opened 2021-09-08
Update Wave 11 indicators dates after backfill

nmdefries:
### Summary
Update dates from #674 after backfilling indicators to Wave 11 release.


docs/symptom-survey/contingency-tables.md
```diff
@@ -66,9 +66,6 @@ Note: If a survey item was introduced in the middle of an aggregation period,
 derived indicators will be included in aggregations for that period but will
 only use a partial week or month of data.
 
-CSVs for the month of January 2021 only use data from January 6-31 due to a
-[definitional change in a major vaccine item on January 6](./coding.md#new-items-2).
```
capnrefsmmat, line 70: Is this no longer true?

capnrefsmmat (commented):
So the API signals say May 20 and the contingency tables say May 16; is that just because how the weeks fall?

One question and then I think this is good to go

docs/symptom-survey/contingency-tables.md
nmdefries: That's correct.

Display logic for V3 was changed on Jan 6, mid-wave 6, to be shown only to respondents indicating that they had not been vaccinated. For the purposes of the contingency tables, we ignore responses to V3 from before the change. During the initial runs we enforced this by only including input data for Jan 6 or later, resulting in non-V3-derived indicators also only using data after Jan 6, hence the note in the docs.

The logic to ignore pre-Jan 6 V3 data is now included in the variable definitions, so I backfilled to the beginning of January so we don't have any partial weeks or months.

nmdefries:
> So the API signals say May 20 and the contingency tables say May 16; is that just because how the weeks fall?

Yep.

docs/symptom-survey/contingency-tables.md
capnrefsmmat: Makes sense, thanks.

capnrefsmmat (approved):

PR #730, opened 2021-10-04
Hotfix acquisition: fix dtypes of missing columns

### LINKED ISSUE #728: covidcast acquisition rejects valid nan-coded rows
We're getting acquisition failures that may be the primary cause of the dramatic drop in added CHNG rows.

Sample error:
```
{
  "detail": [
    "Pandas(geo_id='17000', val=nan, se=nan, sample_size=nan, missing_val='4.0', missing_se='4.0', missing_sample_size='4.0')",
    "missing_val"
  ],
  "file": "/common/covidcast/receiving/chng/20210926_county_smoothed_outpatient_covid.csv",
  "event": "invalid value for row",
  "logger": "load_csv",
  "level": "warning",
  "timestamp": "2021-10-04T03:09:38.804771Z"
}
```

Looking at our tests, it appears that we always instantiate data frames as python objects and not by parsing string data. `pandas` may be inferring the wrong data type for the missingness columns (here looks like it thinks it's a string), causing the error.

dshemetov:
Closes #728 

**Prerequisites**:

- [x] Build is successful

### Summary

Missingness code validation was no ready to receive `str(float(missing_code))` and therefore caused an error. It should now convert from string to float to int. Added tests to cover this case.


src/acquisition/covidcast/csv_importer.py
```diff
@@ -224,7 +224,7 @@ def validate_missing_code(row, attr_quantity, attr_name):
     if hasattr(row, "missing_" + attr_name):
       missing_entry = getattr(row, "missing_" + attr_name)
       try:
-        missing_entry = int(missing_entry)
+        missing_entry = int(float(missing_entry)) # convert from string to float to int
```
krivard, line 227: Would it make more sense to set the dtype for these when the file is read?

https://github.com/cmu-delphi/delphi-epidata/blob/cecb6c950e2e2f189d4b901557d9c54833eb85e6/src/acquisition/covidcast/csv_importer.py#L348

dshemetov: Yea, but I think that would be a slightly larger refactor. A lot of `extract_and_check_row` is written around having everything be a string. I didn't want to go down that road right now, in favor of a quick fix.

krivard (changes requested):
minor adjustments but otherwise looks good!

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -203,40 +203,30 @@ def make_row(
       (make_row(missing_val='missing_val'), 'missing_val'),
       (make_row(missing_se='missing_val'), 'missing_se'),
       (make_row(missing_sample_size='missing_val'), 'missing_sample_size'),
-      (make_row(val='1.2', missing_val=Nans.OTHER), 'missing_val'),
-      (make_row(se='1.2', missing_se=Nans.OTHER), 'missing_se'),
-      (make_row(sample_size='1.2', missing_sample_size=Nans.OTHER), 'missing_sample_size')
+      (make_row(val='1.2', missing_val=str(float(Nans.OTHER))), 'missing_val'),
+      (make_row(se='1.2', missing_se=str(float(Nans.OTHER))), 'missing_se'),
+      (make_row(sample_size='1.2', missing_sample_size=str(float(Nans.OTHER))), 'missing_sample_size'),
```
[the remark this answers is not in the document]
dshemetov: I think this will work, but I was deleting those test cases because they were feeding `missing_*` with int values into `extract_and_check_row`, which never happens since the code converts everything to str...

(Minor vent: grrrrr this code is so unclear about the data types it's dealing with and therefore has to be written in such a defensive way, I really want to blow it up and impose stricter data types on it, but even though that will likely work for covidcast, I don't know about the other data sources delphi-epidata handles >:^0!!!).

krivard:
Superseded by #733

PR #733, opened 2021-10-05
Repair and log nancodes instead of rejecting files

### LINKED ISSUE #731: Acquisition and error-checking
> Normally I'd agree with you, but the failed files archive is lossy (no way to reconstruct previous issues of files that have been failed more than once). Prefer fixing them here, but complaining loudly so we can detect errors in the pipeline.

_Originally posted by @krivard in https://github.com/cmu-delphi/delphi-epidata/issues/729#issuecomment-933809247_

Acquisition (i.e. `csv_importer.py`) currently refuses to upload files with errors in them, even for some minor type errors. Since it's hard to recover rejected data, we should instead upload what we can, but write `logger.errors` for the issues we see.

dshemetov:
Closes #731

**Prerequisites**:

- [x] Build is successful

### Summary

* repair and log instead of rejecting files with invalid missing codes
* introduce dtype inference code in acquisition to see if further refactors possible (see [this comment](https://github.com/cmu-delphi/delphi-epidata/pull/730#discussion_r721686153))
* refactor test_csv_uploading for readability

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -224,7 +224,7 @@ class CsvImporter:
     if hasattr(row, "missing_" + attr_name):
       missing_entry = getattr(row, "missing_" + attr_name)
       try:
-        missing_entry = int(missing_entry)
+        missing_entry = int(float(missing_entry)) # convert from string to float to int
       except ValueError:
         return None
       # A missing code should never contradict the quantity being present,
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -212,7 +212,7 @@ class CsvImporter:
       return "Error"
 
   @staticmethod
-  def validate_missing_code(row, attr_quantity, attr_name):
+  def validate_missing_code(row, attr_quantity, attr_name, filepath=None, logger=None):
     """Take a row and validate the missing code associated with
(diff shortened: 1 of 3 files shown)


src/acquisition/covidcast/csv_importer.py
```diff
@@ -221,27 +221,30 @@ def validate_missing_code(row, attr_quantity, attr_name):
...
-        return None
-      return missing_entry
-    else:
-      if attr_quantity is None:
-        return Nans.OTHER.value
+    logger = get_structured_logger('load_csv') if logger is None else logger
+    missing_entry = getattr(row, "missing_" + attr_name, None)
+
+    try:
+      missing_entry = CsvImporter.floaty_int(missing_entry) # convert from string to float to int
```
krivard, line 237: good ol' floaty_int

krivard (commented):
Looks good; I modified to pull dtype into a constant and some other small adjustments. take a look and let me know if they're friendly?

src/acquisition/covidcast/csv_importer.py
dshemetov: that ol' floaty_int we all know and love

dshemetov:
lgtm. the int conversion at the end of floaty_int is probably not needed, since round returns an int type, but it is more clear that way (I didn't expect round to return an int until i checked).

krivard (approved):
👍

PR #761, opened 2021-11-02
fix: add compatibildity mode for non_standard_message

### LINKED ISSUE #760: `delphi` endpoint returns wrong result code
When there are no results, Epidata should return
```
{"result":-2, "message":"no results", "epidata": []}
```

But for the `delphi` endpoint, we're returning the following, causing `Epidata.check` to fail to identify the no-results condition:
```
$ curl -sL "http://api.covidcast.cmu.edu/epidata/delphi/?system=epic&epiweek=202132"
{"epidata":[],"message":"success","result":1}
$ curl -sL "http://api.covidcast.cmu.edu/epidata/api.php?source=delphi&system=epic&epiweek=202132"
{"epidata":[],"message":"success","result":1}
```

This in turn is causing excessive (multi-hour) runtimes on some of the nowcasting sensor updates.

sgratzl:
closes #760

**Prerequisites**:

- [x] Build is successful

### Summary

fixes non standard messages and also include the compatibility mode

image

image

### DIFF
--- a/src/server/_printer.py
+++ b/src/server/_printer.py
@@ -18,8 +18,16 @@ def print_non_standard(data):
     format = request.values.get("format", "classic")
     if format == "json":
         return jsonify(data)
+
+    if not data:
+        message = "no results"
+        result = -1
     else:
-        return jsonify(dict(result=1, message="success", epidata=data))
+        message = "success"
+        result = 1
+    if result == -1 and is_compatibility_mode():
+        return jsonify(dict(result=result, message=message))
+    return jsonify(dict(result=result, message=message, epidata=data))
 
 
 class APrinter:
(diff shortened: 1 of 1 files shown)


src/server/_printer.py
```diff
@@ -18,8 +18,16 @@ def print_non_standard(data):
     format = request.values.get("format", "classic")
     if format == "json":
         return jsonify(data)
+
+    if not data:
+        message = "no results"
+        result = -1
```
krivard, line 24: [as specified in the API docs](https://cmu-delphi.github.io/delphi-epidata/api/delphi.html#response) (may need to be propagated to the compatibility mode conditional as well)

```suggestion
        result = -2
```

krivard (approved):
👍

PR #786, opened 2021-12-13
Update google-symptoms geo aggregation

krivard:
Addresses part of https://github.com/cmu-delphi/covidcast-indicators/issues/1406

**Prerequisites**:

- [x] Build is successful

### Summary

We're changing the way google-symptoms aggregates hhs and nation to be based on population-weighted average. It turns out neither of these were added to the documentation when they were added to the API 😕  so I've refactored that section a bit to be more clear about what source region is being used.

TODO: verify that this is, in fact, how the new weighted averages are being computed...


jingjtang (approved):
lgtm

PR #806, opened 2022-01-13
Implement API key scrubbing and structured logging

lee14257:
### Summary

- [x] Adds structured logging module to parse server logs in JSON format
- [x] Applies API key scrubbing / masking for Flask logs according to User authentication and no-track status
- [x] Adds additional getter functions for User class: is_authenticated() and is_tracking()
- [x] Elevates SQLAlchemy logging level to "ERROR" to prevent API keys to show up in query logs

### Notes

This PR implements API key scrubbing for logs in the application layer (Flask + gunicorn). If the user is authorized and tracked, the API key will be logged whenever the user makes an API request. If the user is not tracked, the API key parameter will be masked / scrubbed out. 

For logs in the web server layer (NGINX), the API key will be shown as a query string during client requests. This will need to be filtered by creating a custom filter or ingestion pipeline on Filebeat before ingesting to Elasticsearch.

### TODO in separate PR:
- Create an ingestion pipeline / filter in Filebeat to scrub the "api_key" query string from the incoming NGINX logs.
- Make Flask logs more useful (analytics-friendly) for Kibana by adding more parameters such as list of COVIDcast data sources, etc.

### DIFF
--- a/src/server/_logger.py
+++ b/src/server/_logger.py
@@ -1,0 +1,94 @@
+"""Structured logger utility for creating JSON logs in Delphi pipelines."""
+import logging
+import sys
+import threading
+import structlog
+
+
+def handle_exceptions(logger):
+    """Handle exceptions using the provided logger."""
+    def exception_handler(etype, value, traceback):
+        logger.exception("Top-level exception occurred",
+                         exc_info=(etype, value, traceback))
+
+    def multithread_exception_handler(args):
+        exception_handler(args.exc_type, args.exc_value, args.exc_traceback)
+
+    sys.excepthook = exception_handler
+    threading.excepthook = multithread_exception_handler
+
+
+def get_structured_logger(name=__name__,
+                          filename=None,
+                          log_exceptions=True):
+    """Create a new structlog logger.
+
+    Use the logger returned from this in indicator code using the standard
+    wrapper calls, e.g.:
+
+    logger = get_structured_logger(__name__)
+    logger.warning("Error", type="Signal too low").
+
+    The output will be rendered as JSON which can easily be consumed by logs
+    processors.
+
+    See the structlog documentation for details.
+
+    Parameters
+    ---------
+    name: Name to use for logger (included in log lines), __name__ from caller
+    is a good choice.
+    filename: An (optional) file to write log output.
+    """
+    # Configure the underlying logging configuration
+    handlers = [logging.StreamHandler()]
+    if filename:
+        handlers.append(logging.FileHandler(filename))
+
+    logging.basicConfig(
+        format="%(message)s",
+        level=logging.INFO,
+        handlers=handlers
+        )
+    
+    
+
+    # Configure structlog. This uses many of the standard suggestions from
+    # the structlog documentation.
+    structlog.configure(
+        processors=[
+            # Filter out log levels we are not tracking.
+            structlog.stdlib.filter_by_level,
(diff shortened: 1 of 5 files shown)


src/server/_logger.py
```diff
@@ -0,0 +1,94 @@
...
+    sys.excepthook = exception_handler
+    threading.excepthook = multithread_exception_handler
+
+
+def get_structured_logger(name=__name__,
+                          filename=None,
+                          log_exceptions=True):
+    """Create a new structlog logger.
+
+    Use the logger returned from this in indicator code using the standard
```
krivard, line 26:
```suggestion
    Use the logger returned from this in server code using the standard
```

src/server/_query.py
```diff
@@ -235,7 +235,7 @@ def run_query(p: APrinter, query_tuple: Tuple[str, Dict[str, Any]]):
     query, params = query_tuple
     # limit rows + 1 for detecting whether we would have more
     full_query = text(f"{query} LIMIT {p.remaining_rows + 1}")
-    current_user.log_info("full_query: %s, params: %s", full_query, params)
+    current_user.log_info({"full_query" : str(full_query), "params" : params})
```
krivard, line 238: Use kwargs instead of logging a dictionary for all structlog logging

```suggestion
    current_user.log_info("Running query", full_query=str(full_query), params=params)
```

(& please propagate this change to other instances below)

krivard (changes requested):
good start -- but structlog takes kwargs instead of a dict

src/server/_query.py
lee14257: Thanks, changed the parameters into kwargs format and propagated to below

lee14257:
Thanks for the feedback, I've made further changes below:
- [x] Changed all struct log calls to kwargs form.
- [x] Further applied API key scrubbing for logs containing URL or request path - typically during the resolve_user() call that happens before query runs.

src/server/main.py
```diff
@@ -69,4 +69,5 @@ def send_lib_file(path: str):
     app.logger.setLevel(gunicorn_logger.level)
     sqlalchemy_logger = logging.getLogger("sqlalchemy")
     sqlalchemy_logger.handlers = gunicorn_logger.handlers
-    sqlalchemy_logger.setLevel(gunicorn_logger.level)
+    sqlalchemy_logger.setLevel(logging.ERROR)
+    #sqlalchemy_logger.setLevel(gunicorn_logger.level)
```
krivard, line 73: If switching to ERROR is intended to be permanent, drop the commented-out line and add a comment explaining why we can't use the gunicorn setting.

If this was temporary for testing, revert.

lee14257: SQLAlchemy logs at gunicorn settings show all queries, and changing this to ERROR would prevent it from showing API keys when the server is querying the _api_user_ table (regardless of the no-track option).

The structured logger already stores queries that access data source tables (covidcast, fluview, etc.), so having SQLAlchemy logs to do that seems redundant.

Unless we specifically need all queries to be logged for analytics / Kibana, this setting seems to be more appropriate to hide API key details.

krivard (approved):
👍  nice work! do you have merge permissions or should I click the big button?

lee14257:
> 👍 nice work! do you have merge permissions or should I click the big button?

yup! I do have merge permissions

PR #830, opened 2022-02-02
Wave 13 codebook

### LINKED ISSUE #823: Update codebook CSV for Wave 13
We'll want to publish this in the next week or two, since Wave 13 is being deployed.

nmdefries:
### Summary
Update CTIS codebook to include Wave 13 items.

Closes #823.


capnrefsmmat:
Hmm. We appear to have reused item number V2a for a different item than it used to be before wave 10. Will that cause any problems in the codebook descriptions?

nmdefries:
Uh oh. No problems in the pipeline -- we don't use either V2a for anything. The short description in the codebook will be the same for both versions, which is only a minor problem since the questions are fairly similar. But it's definitely misleading to have the two questions use the same name.

I wonder if it would be straightforward to change the name to "V2d"... @WichadaLMK?

WichadaLMK:
Should be pretty straight forward. I can publish the change on Qualtrics but we will have to go back and relabel all the previous micro data for that question. Sorry about that. 

Best - W.LMK

On Feb 3, 2022, at 9:40 AM, nmdefries ***@***.***> wrote:

﻿
Uh oh. No problems in the pipeline -- we don't use either V2a for anything. The short description in the codebook will be the same for both versions, which is only a minor problem since the questions are fairly similar. But it's definitely misleading to have the two questions use the same name.

I wonder if it would be straightforward to change the name to "V2d"... @WichadaLMK?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

WichadaLMK:
Alternatively, we can have the label change on the back end and update the variable name in Qualtrics at Wave 14? I think we had to do this for something else I goofed on 😬

Best - W.LMK

On Feb 3, 2022, at 1:57 PM, Wichada La Motte-Kerr ***@***.***> wrote:

﻿Should be pretty straight forward. I can publish the change on Qualtrics but we will have to go back and relabel all the previous micro data for that question. Sorry about that. 

Best - W.LMK

On Feb 3, 2022, at 9:40 AM, nmdefries ***@***.***> wrote:

﻿
Uh oh. No problems in the pipeline -- we don't use either V2a for anything. The short description in the codebook will be the same for both versions, which is only a minor problem since the questions are fairly similar. But it's definitely misleading to have the two questions use the same name.

I wonder if it would be straightforward to change the name to "V2d"... @WichadaLMK?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

capnrefsmmat:
@nmdefries, do you think implementing something like the existing `bodge_C6_C8` would be easy here, and allow us to not have to reissue a bunch of data?

nmdefries:
Yes, a `bodge` function is a great idea, and then we can officially change the name in the next wave (or just do a mid-wave change and count on the `bodge` function to make the pre and post-change data look the same).

nmdefries:
@capnrefsmmat This is ready.

capnrefsmmat (approved):

PR #831, opened 2022-02-03
Reformat CPR signals table

nmdefries:
### Summary
"Earliest date available" was in its own cell in the table. Replace newline with html line break.


krivard (approved):
👍

PR #837, opened 2022-02-07
Copy `dbjobs` directory contents

korlaxxalrok:
**Prerequisites**:

- [ ] Build is successful

### Summary

Configure deploy.json to copy `.py`, `.txt`, and the directory structure for the `dbjobs` tools.

Also renames `db_config.py` so that we can preserve it as a useful example while also moving it out of the way to allow us to deploy a secrets file with the same name without clobbering.


krivard (approved):
seems legit, though oof txt files really?

korlaxxalrok:
There is a `requirements.txt` file in there. I'm not sure exactly how we will use it, but it was added so that we could remove some full package libraries that shouldn't have been included.

PR #862, opened 2022-02-28
Allow NAs in missing columns

dshemetov:
Part of fixing https://github.com/cmu-delphi/covidcast-indicators/issues/1520. This should be merged first, as it will be a no-op (aside from a memory usage increase on the acquisition server). Then we can merge the archivediffer fix (upcoming).

**Prerequisites**:

- [x] Build is successful

### Summary

Allows NAs in missing columns, so the validation can fix them. This will allow archive differ to easily add deletion codes, not have to do any validation on the rest of the missing codes, and let the validation here in acquisition take care of it.


krivard (approved):
Nice work!

PR #863, opened 2022-03-01
Fix bugs identified in staging tests

krivard:
While testing quidel deletions in staging, it became clear that we missed a few things in review...

**Prerequisites**:

- [x] Build is successful

### Summary

* add missing rowcount handling to `database.py:Database.delete_batch()`
* add missing rowcount-unsupported logic to `delete_batch.py:main()`
* add missing imports to `delete_batch.py`
* add missing argument to `delete_batch.py:handle_file()`
* add missing `time_type` logic to `delete_batch.py:handle_file()`
* add missing arguments to `handle_file()` call in `delete_batch.py:main()`

### DIFF
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -342,7 +342,7 @@ SET `covidcast`.`is_latest_issue`=1;
       self._cursor.execute(fix_latest_issue_sql)
       self._connection.commit()
 
-      if total is None or total == -1:
+      if total == -1:
         # the SQL connector does not support returning number of rows affected (see PEP 249)
         total = None
     except Exception as e:
--- a/src/acquisition/covidcast/database.py
+++ b/src/acquisition/covidcast/database.py
@@ -326,6 +326,7 @@ SET `covidcast`.`is_latest_issue`=1;
 '''
 
     drop_tmp_table_sql = f'DROP TABLE {tmp_table_name}'
+    total = None
     try:
       self._cursor.execute(create_tmp_table_sql)
       self._cursor.execute(amend_tmp_table_sql)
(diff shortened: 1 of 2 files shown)


src/acquisition/covidcast/database.py
```diff
@@ -337,12 +338,18 @@ def delete_batch(self, cc_deletions):
         raise Exception(f"Bad deletions argument: need a filename or a list of tuples; got a {type(cc_deletions)}")
       self._cursor.execute(add_id_sql)
       self._cursor.execute(delete_sql)
+      total = self._cursor.rowcount
       self._cursor.execute(fix_latest_issue_sql)
       self._connection.commit()
+
+      if total is None or total == -1:
```
chinandrew, line 345: Isn't this `is None` condition redundant?

src/acquisition/covidcast/delete_batch.py
```diff
@@ -22,19 +23,21 @@ def get_argument_parser():
...
 
-def handle_file(deletion_file, database):
+def handle_file(deletion_file, database, logger):
     logger.info("Deleting from csv file", filename=deletion_file)
     rows = []
     with open(deletion_file) as f:
         for line in f:
-            rows.append(line.strip().split(","))
+            fields = line.strip().split(",")
+            if len(field) < 9: continue
```
chinandrew, line 32:
```suggestion
            if len(fields) < 9: continue
```
??

src/acquisition/covidcast/database.py
krivard: No; we want this block to run only when total is -1 or when total is None, and not when total is 0. 0 is a falsy value in Python.

src/acquisition/covidcast/delete_batch.py
krivard: thank you 🤦‍♀️

src/acquisition/covidcast/database.py
krivard: Hah! Good catch. This was pattern-replicating from
https://github.com/cmu-delphi/delphi-epidata/blob/442b3b9ff476c3a1d0c5fe13d16d91458a98624c/src/acquisition/covidcast/database.py#L236-L238

where `is None` is necessary to avoid the else clause... but we don't have one of those here, so it's purely vestigial.

```suggestion
      if total == -1:
```

chinandrew (approved):
:shipit: :shipit: :shipit:

PR #876, opened 2022-03-23
state the limitation of the availability

### LINKED ISSUE #875: Update COVIDcast:Quidel docs to describe state-level dropout in raw age group signals on weekends
The age groups we added to quidel for omicron have low availability at the county and msa levels, which is to be expected since those population sizes are quite small.

However, we've also seen availability drop out at the state level on weekends for some age groups:

| signal | age group | February 19-20 | February 26-27 | March 5-6 | March 12-13 |
| - | - | - | - | - | - |
| raw | 0-4 | missing (backfilled) | | | missing |
| raw | 65+ | missing | missing |  | missing |

We should clearly document this behavior in the Limitations section of the API documentation so that users know what to expect.

https://github.com/cmu-delphi/delphi-epidata/blob/dev/docs/api/covidcast-signals/quidel.md

jingjtang:
closes #875 

### Summary

Add description to the limitation of availability for some age groups.

### DIFF
--- a/docs/api/covidcast-signals/quidel.md
+++ b/docs/api/covidcast-signals/quidel.md
@@ -144,7 +144,7 @@ June 14th and subsequently revised on June 16th.
 
 ### Limitations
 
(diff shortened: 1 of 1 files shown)


krivard (approved):
👍

PR #886, opened 2022-03-29
[CTIS] V17 appears twice, remove entry with less info

nmdefries:
### Summary
V17 appears twice, remove entry with less info.


capnrefsmmat (approved):
oops

PR #888, opened 2022-04-01
Set test secrets in test_covidcast_meta

krivard:
**Prerequisites**:

- [x] Part of v4: merge against the primary v4 branch
- [x] Branch is up-to-date with the branch to be merged with
- [ ] Build is successful (v4 tests don't pass yet; these PRs will bring us closer to a successful build)

### Summary

test_covidcast_meta used to depend on prior tests to set secrets.db.host instead of setting them itself, creating an ordering dependency. By setting secrets in each test in which they are used, we remove the ordering dependency and ensure tests pass regardless of whether they're run alone or as part of a suite.


melange396 (approved):
excellent catch, thank you very much!

PR #889, opened 2022-04-04
[CTIS] Wave 13 V2d incorrectly referred to as V2a

nmdefries:
### Summary
Wave 13 V2d incorrectly referred to as V2a


capnrefsmmat (approved):
Thanks for the quick fix

PR #891, opened 2022-04-07
Fix Flask Jinja dependency

korlaxxalrok:
**Prerequisites**:

- [ ] Build is successful

### Summary
Jinja is a dependency of Flask. In Jinga 3.1.0, `escape` is deprecated and should be imported from MarkupSafe. We can pin `jinja<3.1.0` to work around this for now.

Probably the better fix will be to update Flask, but since we are already pinning it to an older version it seems likely that we can't do this easily.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 itsdangerous<2.1
+jinja2<3.1.0
 Flask==1.1.2
 SQLAlchemy==1.3.22
 mysqlclient==2.0.2


korlaxxalrok:
Oh, yep, looks like it 👍 Maybe we could merge `dev` here? Last time I tried that there were conflicts that seemed like they were going to be hard/blocking.

krivard:
There are major conflicts, largely due to the "delete from CSV" work we needed for Quidel. This PR will add another, but it's minor, so I'm not opposed

krivard (approved):
👍

korlaxxalrok:
Ok, good deal. It is a necessity for now as the test env is failing to start because of the jinja deprecation 💥

PR #908, opened 2022-05-16
[CTIS] Add note that current SEs are unweighted

nmdefries:
### Summary
- Add banner with note that current SEs are unweighted.
- Change weighted indicators link in contingency tables page to properly point to weighting discussion.

### DIFF
--- a/docs/api/covidcast-signals/fb-survey.md
+++ b/docs/api/covidcast-signals/fb-survey.md
@@ -726,6 +726,12 @@ of zero, and in simulations improves the quality of the standard error
 estimates. See the [Appendix](#appendix) for further motivation for these
 estimators.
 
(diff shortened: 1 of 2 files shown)


capnrefsmmat (approved):
Huh, we haven't used `\(` as a math delimiter in this file before, but KaTeX appears to support it, so I guess we can. (And probably should, since apparently it's for inline math and `$$` is for display math)

nmdefries:
I was struggling a lot with getting `$$` to display as in-line rather than as a block. Something about being inside of a `div`?

capnrefsmmat:
My read of the [KaTeX manual](https://katex.org/docs/autorender.html#api) is that it treats `$$` to mean display (block) math by default. Somehow that works fine in most of the docs. In any case I rendered the documentation with Jekyll and your update looks fine, so I'm fine leaving it as you wrote it.

PR #917, opened 2022-05-26
Remove remaining wip pieces in tests

dshemetov:
A few tests in #903 still had WIP references. Removed those.


tests/acquisition/covidcast/test_csv_to_database.py
```diff
@@ -87,30 +83,28 @@ def load_csv_impl(path, *args):
       mock_logger,
       csv_importer_impl=mock_csv_importer)
 
-    self.assertEqual(modified_row_count, 4)
+    self.assertEqual(modified_row_count, 1)
```
melange396, line 86: i know this isnt exactly recent in your memory anymore, but why did this change to 1 instead of to 3??

melange396 (commented):
this looks great, but im confused by the one line change that i pointed out...

tests/acquisition/covidcast/test_csv_to_database.py
dshemetov: good question, let me look around. since tests pass, there must've been some change to this test in v4.

dshemetov: thanks for pointing this out! i made a fix, so it's 3 now. here's an explanation:

- we have a database mock object, which overwrites `insert_or_update_bulk`
- we didn't even count how many rows this function was passed, we just updated the mock `return_value` argument
- so i rewrote this so `insert_or_update_bulk = MagicMock(wraps=iter_len)` where `iter_len` is just a function that counts the number of items in the iterable passed to `insert_or_update_bulk`. so now we correctly get `modified_row_count == 3`

melange396 (approved):
👍

PR #922, opened 2022-06-01
schema refinement, dbjobs speed increase, and added migration routine

krivard:
This is the branch with the "constraintz" variant of v4.

* schema refinement
* dbjobs speed increase
* added migration routine


src/acquisition/covidcast/database.py
```diff
@@ -232,52 +217,37 @@ def insert_or_update_batch(self, cc_rows, batch_size=2**20, commit_partial=False
...
     signal_dim_add_new_load = f'''
-        INSERT INTO signal_dim (`source`, `signal`, `compressed_signal_key`) 
-            SELECT DISTINCT `source`, `signal`, compressed_signal_key 
-                FROM `{self.load_table}` 
-                WHERE compressed_signal_key NOT IN 
-                    (SELECT DISTINCT compressed_signal_key 
-                     FROM signal_dim)
+        INSERT INTO signal_dim (`source`, `signal`)
+            SELECT DISTINCT sl.source, sl.signal
+                FROM {self.load_table} AS sl LEFT JOIN signal_dim AS sd
```
nolangormley, line 225: Should we be using with (nolock) on our tables here? Or are we sure that no process will ever have the tables locked?

https://www.sqlshack.com/understanding-impact-clr-strict-security-configuration-setting-sql-server-2017/

nolangormley: question stands for all queries in this file.

src/acquisition/covidcast/database.py
```diff
@@ -318,32 +283,44 @@ def run_dbjobs(self):
...
+
+      self._cursor.execute(signal_latest_load)
+      time_q.append(time.time())
+      logger.debug('signal_latest_load', rows=self._cursor.rowcount, elapsed=time_q[-1]-time_q[-2])
+
+      self._cursor.execute(signal_load_delete_processed)
+      time_q.append(time.time())
+      logger.debug('signal_load_delete_processed', rows=self._cursor.rowcount, elapsed=time_q[-1]-time_q[-2])
+    except Exception as e:
+      raise e
```
nolangormley, line 319: Should this exception be added to the logger so it's in the log file?

src/acquisition/covidcast/migrate_epidata_to_v4.py
```diff
@@ -0,0 +1,188 @@
...
+      FROM epidata.covidcast AS cc
+      USE INDEX(`PRIMARY`)
+      WHERE {batch_lower} <= cc.id AND cc.id < {batch_upper}; """
+    # TODO: use LIMIT instead of id range??
+    # TODO: might it be worth adding "ORDER BY id ASC" ?
+
+    if use_transaction_wrappers:
+      start_tx(db._cursor)
+
+    print(f"-=-=-=-=-=-=-=- RUNNING BATCH STARTING AT {batch_lower} -=-=-=-=-=-=-=-")
```
nolangormley, line 103: Is print okay here since it's the migrate script? Or should this be changed to a log file?

src/ddl/v4_schema.sql
```diff
@@ -15,77 +15,63 @@ GRANT ALL ON covid.* TO 'user';
...
-    `signal_key_id` BIGINT(20) UNSIGNED,
-    `geo_key_id` BIGINT(20) UNSIGNED,
-    `demog_key_id` BIGINT(20) UNSIGNED,  -- TODO: for future use ; also rename s/demog/stratification/
-    `issue` INT(11),
+    `signal_data_id` BIGINT(20) UNSIGNED NOT NULL PRIMARY KEY,
+    `signal_key_id` BIGINT(20) UNSIGNED NOT NULL,
+    `geo_key_id` BIGINT(20) UNSIGNED NOT NULL,
+    `strat_key_id` BIGINT(20) UNSIGNED NOT NULL DEFAULT 1,  -- TODO: for future use
+    `issue` INT(11) NOT NULL,
     `data_as_of_dt` DATETIME(0),  -- TODO: for future use ; also "as_of" is problematic and should be renamed
```
nolangormley, line 47: rename before migrating the schema?

nolangormley (commented):
Left some comments. Big question I have is whether or not we should use nolock in the Acquisition database.py file or check in some way that there are no current locks on the tables in question.

Some other general questions about logging.

src/ddl/v4_schema.sql
melange396: Renaming a column is cheap and easy, and it isnt a priority that we do this right now.  It is important that we have the "column space" reserved with the appropriate datatype because adding it later isnt so simple.  TBQH i would like to rename several of the columns (and we have talked about it before) but the hope was that we finish the real work first and deal with the naming later.

src/acquisition/covidcast/database.py
melange396: in practice, locks from these statements should only last a very short time, on the order of (or less) than the time it takes to process one csv file (which is quite fast).  using this can also lead to inconsistent data reads for outside transactions, so its not worth any small reductions in time waiting for locks to be released.  (also, the article you referred to is for MS SQL Server though an equivalent exists for mysql)

src/acquisition/covidcast/migrate_epidata_to_v4.py
melange396: I think its ok to keep it as print() because this is only going to be run ~2 times total in the process of migrating the data to v4, and there isnt much to be gained from structured or searchable logs.

src/acquisition/covidcast/database.py
nolangormley: Gotcha, makes sense!

nolangormley (approved):
George answered my questions on some of the code. All looks good!

src/acquisition/covidcast/database.py
melange396: well...

this method is only really called by `insert_or_update_batch` which in turn will again catch and re-raise this exception, and it states "rollback is handled in csv_to_database; if you're calling this yourself, handle your own rollback".

also, instantiating the [structured_logger](https://github.com/cmu-delphi/delphi-epidata/blob/v4-srrpp-migrations/src/acquisition/covidcast/logger.py#L21) registers a handler by default that should log any uncaught exceptions.

looking at this now (and unless im missing something), its really just a no-op to catch an unspecific exception and simply re-raise it.  the same situation effectively exists in the above mentioned `insert_or_update_batch`, but its presence coupled with the comment hints to a maintainer that exception handling is not an afterthought and is being dealt with elsewhere.

and to be fair, i cribbed the pattern in this method from @krivard 's version [here](https://github.com/cmu-delphi/delphi-epidata/blob/krivard/v4-rpp-mergeddim-leftjoin/src/acquisition/covidcast/database.py#L305) (because im currently in the middle of incorporating lots of her testing fixes from that branch into this one).  she probably had a reason for this that got lost along the way; perhaps other code was removed, like an Exception subclass case that was handled differently, or a finally block.

SO i think the correct thing to do here is to remove these no-op catch+reraise block boundaries, BUT i am going to leave them for now because they are harmless and we can discuss their efficacy later (or preferably sooner!).  i made a note to address this that i will either act on in the very near future, either by fixing it or turning it into a github issue.  i also made a note for some things i noticed in the logger class while investigating this, and it will have 

im not sure why i wrote such a novel for this.  ¯\\\_(ツ)_/¯

melange396:
Whoops, sorry if i assaulted anyones inboxes with edits i made to comments above!!!  I forgot that i was responding to individual review comments and not submitting a single unified review myself (i thought i had to hit submit for them to be sent all at once).

melange396:
This changeset slightly outgrew the scope of what this branch was supposed to be, so i want to summarize the totality of the delta up to this point:

 - simplifed schema definitions (and added placeholder stratification dimension table)
 - sped up various db operations by adding comprehensive indexes and by removing: bad index hints in the views, process_status, and compressed keys
 - integrated dbjobs() call into the end of insert_or_update_batch
 - created script to perform migration of data from v3 schema to v4 schema in batches
 - added method to fix autoincrement counter if it gets reset

melange396 (approved):

src/acquisition/covidcast/database.py
krivard: > i think the correct thing to do here is to remove these no-op catch+reraise block boundaries, BUT i am going to leave them for now because they are harmless and we can discuss their efficacy later (or preferably sooner!)

+1, agree on both counts

PR #923, opened 2022-06-01
v4 variant: Merged dimensions

### LINKED ISSUE #897: Refactor integration tests of server
The integration tests of the server code are arduous to maintain, largely because they violate best practices:

* Avoid copy-pasted code
* Use the least complex example that adequately exercises the code under test

We want to refactor it to address these problems by adding a helper module. This helper module should include (a hierarchy of?) testcase superclasses to support:

* setting Epidata.BASE_URL and secrets
* truncating tables
* loading data
    * should isolate variations so they are easy to read and reason about at the point of use
    * should return a dict in epidata response format using same ordering
    * should feel free to use the associated database.Database as a helper -- these are server tests; database.Database should have been tested elsewhere

The files affected:

* integrations/server/test_covid_hosp.py - has a reasonable example of what loading data might look like
* integrations/server/test_covidcast.py
* integrations/server/test_covidcast_endpoints.py
* integrations/server/test_covidcast_meta.py
* integrations/server/test_covidcast_nowcast.py
* integrations/server/test_fluview.py - setup only
* integrations/server/test_fluview_meta.py - setup only

krivard:
### Summary

The MySQL query planner doesn't know how to do a merged join across the `signal_dim` and `geo_dim` tables, which resulted in some runtime performance problems on certain queries. This PR manually implements a merged dimension table `merged_dim` which should be used instead for greater runtime efficiency.

This PR also includes a refactor of the covidcast unit and integration tests to reduce duplication and make test cases easier to decipher on reading.

## Fixes

Fixes #897

### DIFF
--- a/dev/docker/database/epidata/Dockerfile
+++ b/dev/docker/database/epidata/Dockerfile
@@ -1,5 +1,14 @@
-# start with the `delphi_database` image
-FROM delphi_database
+# start with a standard percona mysql image
+FROM percona:ps-8
+
+# percona exits with the mysql user but we need root for additional setup
+USER root
+
+# use delphi's timezome
+RUN ln -s -f /usr/share/zoneinfo/America/New_York /etc/localtime
+
+# specify a development-only password for the database user "root"
+ENV MYSQL_ROOT_PASSWORD pass
 
 # create the `epidata` database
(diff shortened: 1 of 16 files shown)


krivard:
Invalidated by #954

Issue #896, opened 2022-04-11
Write integration tests for src/acquisition/covidcast/database.py


krivard:
Integration tests because we do want to exercise the actual database engine here, and not a mock.

* [ ] make sure is_latest_issue is being set correctly - 
    * when adding new daily data (data in signal_load has > issue than data in signal_latest)
    * when uploading data patches (data in signal load has < issue than data in signal_latest)
* [ ] make sure data is being correctly transferred from signal_load into signal_* and *_dim
   * make sure signal_latest is only overwritten when signal_load issue is >=
   * make sure rows are added to *_dim only when needed

+Any other corner cases that worry you.

Use integrations/acquisition/covidcast/test_*.py as templates.

melange396:
the associated pr #925 is a good start but it will, at the very least, need to update the table names as they have changed

PR #925, opened 2022-06-06
Integration Test Case for is_latest_issue using DB

### LINKED ISSUE #926: [Refactor] `insert_or_update_bulk` is a thin wrapper around `insert_or_update_batch` without a purpose?
in `src/acquisition/covidcast/database.py` we have:

```py
  def insert_or_update_bulk(self, cc_rows):
    return self.insert_or_update_batch(cc_rows)
```

minor nit: this wrapper doesn't seem to do anything...

_Originally posted by @dshemetov in https://github.com/cmu-delphi/delphi-epidata/pull/903#discussion_r883791096_

xavier-xia-99:
Summary

This is a fix for issue #896.

The earlier integration tests did not consist of a connection to the database, instead using a MagicMock() object. This did not emulate the actual behaviour of the system. This PR includes a test file `is_latest_issue.py` where the database was actively used to query to ensure that the is_latest_issue flag is accurately reflected in signal_latest and signal_history. It also checks for the addition of new `src` , `sig` and `geo_value`, `geo_type` into respective *_dim tables.

It also includes an edge case that was identified with regards to the db pipeline.

Fixes:
- syntax error failing CI
- fixes #926

### DIFF
--- a/integrations/acquisition/covidcast/test_dim_tables.py
+++ b/integrations/acquisition/covidcast/test_dim_tables.py
@@ -75,9 +75,11 @@ class CovidcastDimensionTablesTests(unittest.TestCase):
         ]
         self._db.insert_or_update_batch(rows)
         self._db.run_dbjobs()
+        
+        #initializing local variables to be used throughout later
         self._db._cursor.execute(self.viewGeoDim)
         record = self._db._cursor.fetchall()
-        geoDimRowCount = len(list(record))
+        geoDimRowCount = len(list(record)) 
 
         self._db._cursor.execute(self.viewSignalDim)
         record = self._db._cursor.fetchall()
@@ -107,8 +109,8 @@ class CovidcastDimensionTablesTests(unittest.TestCase):
             #ensure new entries are added in latest
             self._db._cursor.execute(self.viewSignalLatest)
             record = self._db._cursor.fetchall()
-            sigLatestRowCount = len(list(record)) #updating sigLatestRowCount
-            self.assertEqual(4 , sigLatestRowCount) #diff timevalue and issue, 2 older + 2 newer = 4
+            sigLatestRowCount = len(list(record)) 
+            self.assertEqual(4 , sigLatestRowCount) #added diff timevalue and issue, 2 older + 2 newer = 4
             
             #ensure nothing changed in geoDim
             self._db._cursor.execute(self.viewGeoDim)
@@ -137,8 +139,8 @@ class CovidcastDimensionTablesTests(unittest.TestCase):
             #ensure new entries are added in latest
             self._db._cursor.execute(self.viewSignalLatest)
             record = self._db._cursor.fetchall()
-            sigLatestRowCount = len(list(record)) #updating sigLatestRowCount
-            self.assertEqual(6 , sigLatestRowCount) #diff timevalue and issue, 2 more added ontop of 4 previously
+            sigLatestRowCount = len(list(record)) 
+            self.assertEqual(6 , sigLatestRowCount) #added diff timevalue and issue, 2 more added ontop of 4 previously
 
             #ensure nothing changed in geoDim
             self._db._cursor.execute(self.viewGeoDim)
@@ -158,7 +160,7 @@ class CovidcastDimensionTablesTests(unittest.TestCase):
 
             self._db._cursor.execute(self.viewSignalLatest)
             record = self._db._cursor.fetchall()
-            sigLatestRowCount = len(list(record)) #update
+            sigLatestRowCount = len(list(record)) 
             self.assertEqual(8, sigLatestRowCount) #total entries = 2 + 6 in previous subtest
 
             #ensure nothing changed in geoDim with repeated geo_type, geo_value pairs
--- a/integrations/acquisition/covidcast/test_is_latest_issue.py
+++ b/integrations/acquisition/covidcast/test_is_latest_issue.py
@@ -98,7 +98,7 @@ class CovidcastLatestIssueTests(unittest.TestCase):
 
     updateRow = [
       CovidcastRow('src', 'sig', 'day', 'state', 20200414, 'pa', 
-                  6.5, 2.2, 11.5, nmv, nmv, nmv, 20200414, 2)]  #should not showup
+                  6.5, 2.2, 11.5, nmv, nmv, nmv, 20200414, 2)]  #update previous entry
     self._db.insert_or_update_batch(updateRow)
     self._db.run_dbjobs()
     
@@ -111,7 +111,7 @@ class CovidcastLatestIssueTests(unittest.TestCase):
     #dynamic check for signal_history's list of issue
     self._db._cursor.execute(f'SELECT `issue` FROM {Database.history_table}')
     record3 = self._db._cursor.fetchall()
-    totalRows = len(list(record3)) #updating totalRows
+    totalRows = len(list(record3)) 
     self.assertEqual(2, totalRows) #added 1 new row, updated old row. Total = 2
     self.assertEqual(20200416,max(record3)[0]) #max of the outputs is 20200416 , extracting from tuple
     


integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+      CovidcastRow('src', 'sig', 'day', 'state', 20200414, 'pa', 
+                   1.5, 2.5, 3.5, nmv, nmv, nmv, 20200414, 0),
+      CovidcastRow('src', 'sig', 'day', 'state', 20200415, 'wa', 
+                   3, 2, 1, nmv, nmv, nmv, 20200415, 0)
+    ]
+    self._db.insert_or_update_bulk(rows)
+    self._db.run_dbjobs()
+    #preview
+    self._db._cursor.execute('''SELECT * FROM `signal_history`''')
+    self.totalRows = len(list(self._db._cursor.fetchall()))
```
melange396, line 98: `totalRows` doesnt need to be an object member (test methods probbly shouldnt alter the TestCase object without a good reason)

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+    for row in record:
+      x.add_row(list(row)[:len(x.field_names)])
+    print(x)
+    print("Finish with 1st Set of Data")
+
+    #sanity check for adding dummy data
+    sql = '''SELECT `issue` FROM `signal_latest` where `time_value` = 20200414'''
+    self._db._cursor.execute(sql)
+    record = self._db._cursor.fetchall()
+    self.assertEqual(record[0][0], 20200414)
```
melange396, line 127: probbly worth checking that there is only one record in the results

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+    sql = '''SELECT `issue` FROM `signal_latest` where `time_value` = 20200414  '''
+    self._db._cursor.execute(sql)
+    record2 = self._db._cursor.fetchall()
+    self.assertEqual(record, record2) #same as previous as is_latest did not change
+
+    #dynamic check
+    sql2 = '''SELECT `issue` FROM `signal_history` where `time_value` '''
+    self._db._cursor.execute(sql2)
+    record3 = self._db._cursor.fetchall()
+    self.assertEqual(3,self.totalRows + 1) #ensure 3 added (1 of which refreshed)
```
melange396, line 161: i dont think this is checking what you want...  `totalRows` doesnt get updated and is still equal to 2, which you set at the very top of this test

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+    self.assertEqual(20200416,max(list(record3))[0]) #max of the outputs is 20200416 , extracting from tuple
+    
+    #check older issue not inside latest, empty field
+    sql = '''SELECT * FROM `signal_latest` where `time_value` = 20200414 and `issue` = 20200415 '''
```
melange396, line 165: did you mean `issue=14`?  theres never an ` issue=15` with `time_value=14`

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+      2, 2, 2, nmv, nmv, nmv, 20200414, 0),
+      CovidcastRow('src', 'sig', 'day', 'county', 20200414, 'ca', # updating previous entry
+      3, 3, 3, nmv, nmv, nmv, 20200414, 0)
+    ]
+    self._db.insert_or_update_bulk(rows)
+    self._db.run_dbjobs()
+    #setting baseline variables
+    self._db._cursor.execute('''SELECT * FROM `geo_dim`''')
+    record = self._db._cursor.fetchall()
+    self.geoDimRows = len(list(record))
```
melange396, line 190: like before, these variables dont need to be attached to the `self` object

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+    self._db._cursor.execute('''SELECT * FROM `signal_latest`''')
+    record = self._db._cursor.fetchall()
+    self.sigLatestRows = len(list(record))
+
+    #test not added first
+    with self.subTest(name='older src and sig not added into sig_dim'):
+      oldSrcSig = [
+        CovidcastRow('src', 'sig', 'day', 'state', 20211111, 'pa', #new src, new sig
+                    99, 99, 99, nmv, nmv, nmv, 20211111, 1),
+        CovidcastRow('src', 'sig', 'day', 'county', 20211111, 'ca', #new src, new sig
```
melange396, line 205: maybe use a `geo_value` that doesnt look like a state abbreviation when you have a `geo_type` of "county" (just for reader clarity, it doesnt actually matter for the test)

melange396 (changes requested):
you should revert `test_covidcast_meta_caching.py` (its changes are minuscule and not part of this issue) and delete `dev/local/output.txt`

there are a lot of repeated strings in the SQL statements.  they can be saved to variables for re-use -- which can also mean less reading for anyone doing maintenance on this.

i think i would remove the `print()` statements and `prettytable` and `maxDiff=None` stuff, unless you think theyre particularly useful.  or maybe just reduce them to one call to your `view_table()` helper fn in the `test_*()` methods?

clean up these and the other comments, and we can take another pass at it!

integrations/acquisition/covidcast/test_is_latest_issue.py
xavier-xia-99: That's a good point! What are some sample `geo_value` normally used?

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,346 @@
...
+    
+    # sig_load = PrettyTable()
+    # sig_load.field_names = [('signal_data_id',), ('signal_key_id',), ('geo_key_id',), ('demog_key_id',), ('issue',), ('data_as_of_dt',), ('source',), ('signal',), ('geo_type',), ('geo_value',), ('time_type',), ('time_value',), ('reference_dt',), ('value',), ('stderr',), ('sample_size',), ('lag',), ('value_updated_timestamp',), ('computation_as_of_dt',), ('is_latest_issue',), ('missing_value',), ('missing_stderr',), ('missing_sample_size',), ('legacy_id',), ('compressed_signal_key',), ('compressed_geo_key',), ('compressed_demog_key',), ('process_status',)]
+    
+    #when uploading data patches (data in signal load has < issue than data in signal_latest)
+    #when signal_load is older than signal_latest, we patch old data (i.e changed some old entries)
+    pass
+
+  @unittest.skip("Having different (time_value,issue) pairs in one call to db pipeline does not happen in practice")
```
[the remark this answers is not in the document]
krivard: Can one of you clarify for me? I thought the whole point of batch issue uploads was that they included multiple issues.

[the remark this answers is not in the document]
xavier-xia-99: Oh! I was writing tests with multiple issues but it was not performing as intended so I sought George out for some help. Would it be possible to find a time to sit down together to go through this quickly?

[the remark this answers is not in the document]
krivard: I found it -- `insert_or_update_bulk` is called from within a loop, such that it is called separately for each CSV file. Each CSV contains only one time_value and only one issue.

https://github.com/cmu-delphi/delphi-epidata/blob/61a3d8bcd8fe3bb0b2d362eee8ce5984fbc9ea4d/src/acquisition/covidcast/csv_to_database.py#L91-L109

However, this test also includes `run_dbjobs`, which is not called as part of the `csv_to_database` script or as part of `insert_or_update_bulk`. We cannot therefore expect that `run_dbjobs` will only deal with one time_value and issue at a time. So it depends on what this test is trying to verify:

* correct behavior of `insert_or_update_bulk` alone? --> then it should not call run_dbjobs
* correct behavior of `insert_or_update_bulk` and `run_dbjobs` when run separately with each CSV file? --> then this test doesn't tell us whether run_dbjobs will run correctly as part of the intended load pipeline
* correct behavior of `insert_or_update_bulk` when run separately with each CSV file, and of `run_dbjobs` when run after all available CSVs have been loaded into `signal_load`? --> then this test should run `insert_or_update_bulk` once for each time_value x issue configuration, and then `run_dbjobs` once at the end.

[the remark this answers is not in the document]
melange396: It is a flaw (for some definition of 'flaw') that we cannot properly process multiple issues per datapoint in a single run of `insert_or_update[+dbjobs]` if the issues are not provided in increasing order (or really, if the 'latest' issue is not last in the series).  In practice, `insert_or_update_bulk()` is only called by `csv_to_database:upload_archive()` and each call only specifies a single issue (in fact, the only part of the composite key that differs is the `geo_value`).  This is not going to cause us problems for the foreseeable future, but it is still worth discussing to determine if/how we want to account for possibilities later (adding/changing logic in `insert_or_update` (in the SQL or in the python), carrying some of those key-field arguments further down the call stack to prevent unintended consequences, refactoring the larger ingestion pipeline, or otherwise).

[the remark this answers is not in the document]
krivard: Results of zoom discussion:
Options:
- one CSV per insert, many CSVs per dbjobs
- **Decision**: one CSV per insert+dbjobs. Have insert call dbjobs itself so csv_to_database doesn't have to bother about it.
  - if someone chooses to run dbjobs on an untrusted signal_load, it's their responsibility to ensure that issues are ordered correctly.
  - could add a guard to dbjobs to check if signal_load is sorted correctly but that check would take a long time so let's not. **comment it obsessively though at the head of dbjobs.**

Options for fixing at dbjobs time: none of them are good so let's not.
- get max before loading into latest
- sort while loading into latest

Batch issue uploads do traverse issues in the correct order, but since we're having insert call dbjobs we don't need to care.

integrations/acquisition/covidcast/test_is_latest_issue.py
xavier-xia-99: Here was what I was trying to test: Initially, there are 2 rows and I update one of them and added a new one, so I wanted to make sure that there were only 3 rows in the end. If this is confusing, I am trying to think of better ways to test it.
(1 old & not updated)
(1 old but updated)
(1 added newly)

integrations/acquisition/covidcast/test_is_latest_issue.py
xavier-xia-99: Thank you for that! Removed issue = 15 totally too from the dummy data

melange396 (changes requested):
Looks pretty good, just a few little things:
- there are still some unused imports (including `prettytable`, so you can also remove that from `requirements.txt`)
- you should revert the changes in `integrations/acquisition/covidcast/test_covidcast_meta_caching.py`
- check your comments on the CovidcastRow() lines, i think some need to be updated
- some spacing nits (usually want an empty line right before a class or method definition line, usually dont want an empty line right after;  probably dont want double empty lines unless separating important or otherwise large sections of code)

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,262 @@
...
+
+    # use the local instance of the epidata database
+    secrets.db.host = 'delphi_database_epidata'
+    secrets.db.epi = ('user', 'pass')
+
+    #Commonly used SQL commands:
+    self.viewSignalLatest = '''SELECT * FROM `signal_latest`'''
+    self.viewSignalHistory = '''SELECT * FROM `signal_history`'''
+    self.viewSignalDim = '''SELECT `source`, `signal` FROM `signal_dim`'''
+    self.viewGeoDim = '''SELECT `geo_type`,`geo_value` FROM `geo_dim`'''
```
krivard, line 70: For easier maintainability, use e.g. `Database.history_table` instead of "signal_history" in these

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
+"""Integration tests for covidcast's is_latest_issue boolean."""
+# standard library
+import unittest
+import time
```
melange396, line 4:
```suggestion
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
...
+import threading
```
melange396, line 5:
```suggestion
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+# third party
+from aiohttp.client_exceptions import ClientResponseError
```
melange396, line 9:
```suggestion
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
...
+import mysql.connector
+import pytest
+# first party
+from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
```
melange396, line 13:
```suggestion
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
...
+from delphi_utils import Nans
+from delphi.epidata.client.delphi_epidata import Epidata
+from delphi.epidata.acquisition.covidcast.database import Database, CovidcastRow
+from delphi.epidata.acquisition.covidcast.covidcast_meta_cache_updater import main as update_covidcast_meta_cache
```
melange396, line 17:
```suggestion
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,256 @@
...
+# standard library
+import unittest
+import time
+import threading
+
+
+# third party
+from aiohttp.client_exceptions import ClientResponseError
+import mysql.connector
+import pytest
```
melange396, line 11:
```suggestion
```

melange396 (changes requested):

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+      #testing src, sig 
+      self._db._cursor.execute(self.viewSignalDim)
+      record = self._db._cursor.fetchall()
+      res = [('src','sig')] #output
+      self.assertEqual(res , list(record))
+
+      #ensure new entries are added in latest
+      self._db._cursor.execute(self.viewSignalLatest)
+      record = self._db._cursor.fetchall()
+      self.assertEqual(len(list(record)), sigLatestRows + 2) #2 original, 2 added
```
krivard, line 173: A pattern like this will be easier to maintain -- we'll be able to add new tests without having to update all subsequent expressions.
```suggestion
      record_length = len(list(record))
      self.assertEqual(record_length, sigLatestRows + 2) #2 original, 2 added
      sigLatestRows = record_length
```

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+                      2, 2, 2, nmv, nmv, nmv, 20200414, 0)
+      ]  
+      self._db.insert_or_update_bulk(newSrcSig)
+      self._db.run_dbjobs()
+
+      #testing src, sig
+      self._db._cursor.execute(self.viewSignalDim)
+      record = self._db._cursor.fetchall()
+      res = [('src', 'sig'), ('new_src', 'sig'), ('src', 'new_sig')]
+      self.assertEqual(res , (record))
```
krivard, line 194: Update sigDimRows here

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+      self._db.insert_or_update_bulk(repeatedGeoValues)
+      self._db.run_dbjobs()
+
+      self._db._cursor.execute(self.viewGeoDim)
+      record = self._db._cursor.fetchall()
+      self.assertEqual(len(list(record)),geoDimRows) #geoDimRows unchanged
+
+      self._db._cursor.execute(self.viewSignalLatest)
+      record = self._db._cursor.fetchall()
+      self.assertEqual(len(list(record)),sigLatestRows + 6) #total entries = 2(initial) + 6(test)
```
krivard, line 216: Update sigLatestRows here

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+              3, 3, 3, nmv, nmv, nmv, 20200414, 0),
+        CovidcastRow('src', 'sig', 'day', 'county', 20200414, '15451', # everything same except, county = nj
+              3, 3, 3, nmv, nmv, nmv, 20200414, 0)
+          ]  
+      self._db.insert_or_update_bulk(newGeoValues)
+      self._db.run_dbjobs()
+
+      self._db._cursor.execute(f'SELECT `geo_type`,`geo_value` FROM `geo_dim`')
+      record = self._db._cursor.fetchall()
+      self.assertEqual(len(list(record)),geoDimRows + 3) #2 + 3 new
```
krivard, line 232: Update geoDimRows here

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+
+      self._db._cursor.execute(self.viewSignalLatest)
+      record = self._db._cursor.fetchall()
+      self.assertEqual(len(list(record)),sigLatestRows + 6 + 3) #total entries = 2(initial) + 6(test)
```
krivard, line 236: Update sigLatestRows here

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -0,0 +1,250 @@
...
+      CovidcastRow('src', 'sig', 'day', 'state', 20200414, 'pa', #updating old issue, should be seen in latest
+                17, 17, 17, nmv, nmv, nmv, 20200417, 3),
+      CovidcastRow('src', 'sig', 'day', 'state', 20200414, 'pa', # updating previous entry
+                2, 2, 3, nmv, nmv, nmv, 20200416, 2)
+    ]
+    self._db.insert_or_update_bulk(rows) 
+    self._db.run_dbjobs()
+    self._db._cursor.execute(f'SELECT `issue` FROM {Database.latest_table} ')
+    record = self._db._cursor.fetchall()  
+    self.assertEqual(record[0][0], 20200417) #20200416 != 20200417
```
krivard, line 250:
```suggestion
    # Make sure the 4/17 issue is listed even though 4/16 was imported after it
    self.assertEqual(record[0][0], 20200417)
```

krivard (changes requested):
Changes for maintainability and a question

integrations/acquisition/covidcast/test_dim_tables.py
```diff
@@ -125,8 +125,16 @@ def test_src_sig(self):
...
+<<<<<<< HEAD
             res = [('src', 'sig'), ('new_src', 'sig'), ('src', 'new_sig')]
             self.assertEqual(res , (record))
+=======
+            self.sigDimRows = len(list(record))
+            
+            res = set([('new_src', 'sig'), ('src', 'new_sig'), ('src', 'sig')])
+            self.assertEqual(res , set(record))
+            self.assertEqual(3, self.sigDimRows)
+>>>>>>> 8a39f36b (used set to remove ordering of elements in return list)
```
krivard, line 137: Looks like the conflict edit didn't stick -- are you okay to fix?

integrations/acquisition/covidcast/test_dim_tables.py
```diff
@@ -71,19 +73,19 @@ def test_src_sig(self):
         CovidcastRow('src', 'sig', 'day', 'county', 20200414, '11111',
         3, 3, 3, nmv, nmv, nmv, 20200414, 0)
         ]
-        self._db.insert_or_update_bulk(rows)
+        self._db.insert_or_update_batch(rows)
         self._db.run_dbjobs()
         self._db._cursor.execute(self.viewGeoDim)
         record = self._db._cursor.fetchall()
-        geoDimRows = len(list(record))
+        self.geoDimRows = len(list(record))
```
krivard, line 80: Does geoDimRows need to be an object member, or can it be a local variable instead? Remember, [test methods probbly shouldnt alter the TestCase object without a good reason](https://github.com/cmu-delphi/delphi-epidata/pull/925/files/18a27582d8056659a7afa251c78e9e5fef6c7d35..d250f980e070679bcd5f2dfaff70fd1336475ad2#r891518731)

integrations/acquisition/covidcast/test_is_latest_issue.py
```diff
@@ -68,17 +67,17 @@ def test_signal_latest(self):
...
-    totalRows = len(list(self._db._cursor.fetchall()))
+    self.totalRows = len(list(self._db._cursor.fetchall()))
 
     #sanity check for adding dummy data
     sql = f'SELECT `issue` FROM {Database.latest_table} where `time_value` = 20200414'
     self._db._cursor.execute(sql)
     record = self._db._cursor.fetchall()
     self.assertEqual(record[0][0], 20200414)
-    self.assertEqual(len(record), 1) #check 1 entry
+    self.assertEqual(len(record), 1) #check 1 entry present
```
krivard, line 80: line comments should add context for future readers, beyond just restating the code -- for example, this modification explains why we only expect 1 entry for this query: 
```suggestion
    self.assertEqual(len(record), 1) #placeholder data only has one issue for 20200414
```

krivard (changes requested):
Needs a thorough pass through to:
* [ ] change object members to local variables (except where they reduce duplication, like in `self.viewSignalLatest`)
* [ ] either strip out redundant line comments or add context that explains the reasoning behind the operation.

integrations/acquisition/covidcast/test_is_latest_issue.py
xavier-xia-99: I edited them accordingly, let me know if they showed up!

PR #752, opened 2021-10-24
Local dev scripts: installation, docker refresh

krivard:
Installation script ensures nothing but code winds up in the resulting Docker images.

Refresh script automatically keeps track of what's running, shutdown/build/run procedures, and test bindmounts. We've got people copy-pasting whole paragraphs of shell commands and losing track of which copypastes do what, and that's just not necessary.

**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

Issue #742, opened 2021-10-14
Possibly add retry to async tests


dshemetov:
Some of the async tests sporadically fail, due to a suspected race condition, but fail after a few more retries of the tests. We may want to add a retry condition to automatically do this.

dshemetov:
> These are the local integration tests?

Yup! 

Here are some error messages:
```
E           aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error', url=URL('http://delphi_web_epidata/epidata/api.php?source=covidcast&data_source=src&signals=sig&time_type=day&geo_type=county&geo_value=11111&time_values=20200414')

/usr/local/lib/python3.8/site-packages/aiohttp/client_reqrep.py:1000: ClientResponseError
=========================== short test summary info ============================
FAILED repos/delphi/delphi-epidata/integrations/acquisition/covid_hosp/facility/test_scenarios.py::AcquisitionTests::test_acquire_dataset
FAILED repos/delphi/delphi-epidata/integrations/client/test_delphi_epidata.py::DelphiEpidataPythonClientTests::test_async_epidata
========================= 2 failed, 47 passed in 5.38s =========================
```
and
```
E                        AssertionError: Lists differ: [{'ep[273 chars]}, {'epidata': [], 'message': "database error:[7911 chars]ts'}] != [{'ep[273 chars]}, {'result': -2, 'message': 'no results'}, {'[3523 chars]ts'}]
E
E                        First differing element 1:
E                        {'epidata': [], 'message': "database error:[659 chars]: -1}
E                        {'result': -2, 'message': 'no results'}
E
E                        Diff is 13820 characters long. Set self.maxDiff to None to see it.
```

This is what the docker web-server log shows when it occurs
```
[2021-10-05 13:21:02 -0400] [42] [INFO] Invalidate connection <_mysql.connection open to 'delphi_database_epidata' at 0x564cb285d120> (reason: OperationalError:(2006, 'MySQL server has gone away'))
```

melange396:
does this still happen?  i cant remember the last time i saw this (though absence of evidence is not evidence of absence, and my memory can be faulty)...  changes elsewhere in the server code may have resolved the issue, and im p[ushing to deprecate the async stuff](https://github.com/cmu-delphi/delphi-epidata/pull/1295), so this can probably be closed.

PR #936, opened 2022-06-18
Add bootstrap installer and a Docker controller Makefile

dshemetov:
**Prerequisites**:

- [ ] Build is successful

### Summary

I had some local devtools laying around and figured I would try merging them with #752. The differences aren't large
- no need for a `driver` directory by symlinking a `.dockerignore` file that only passes `repos` through to the build context
- use a [Makefile controller](https://washpost.engineering/seamless-docker-control-with-make-c1a1827592e) for the script (up to the reader if it's clearer than a bash script)
- added in some missing (though [unused](https://github.com/cmu-delphi/operations/pull/70)) dependencies to `install.sh`

Just figured I would see if I could help push this work along. Though I didn't add much, maybe consider it a thorough review, so we can get closer to merging? After this, we can really simplify our dev docs and speed up engineer onboarding time.


krivard:
Did you rebase commits from another branch? I merged #742 earlier this week

xavier-xia-99 (commented):
Everything works for setting up the environment from a new directory to testing. 

Would also just love to include the additional flags when running `make test` in the markdown!

dshemetov:
I rebased to `dev`.

xavier-xia-99:
All good! Thanks

xavier-xia-99 (approved):
All seems good now! Thank you!

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+	test-subdir=repos/delphi/delphi-epidata/tests repos/delphi/delphi-epidata/integrations
+endif
+
+# Set your shell here
+SHELL:=/bin/sh
+
+# Get the Makefile's absolute path: https://stackoverflow.com/a/324782/4784655
+# (if called from symlink, the path is the location of the symlink)
+NOW:=$(shell date "+%Y-%m-%d")
+CWD:=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
krivard, line 63: Assuming the comment goes with CWD and not NOW:

```suggestion
# Get the Makefile's absolute path: https://stackoverflow.com/a/324782/4784655
# (if called from symlink, the path is the location of the symlink)
CWD:=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
NOW:=$(shell date "+%Y-%m-%d")
```

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+LOG_WEB:=$(CWD)docker-logs/delphi_web_epidata_$(NOW).log
+LOG_DB:=$(CWD)docker-logs/delphi_database_epidata_$(NOW).log
```
krivard, line 65: Why do these need to specify CWD?

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+	-f repos/delphi/delphi-epidata/dev/docker/python/Dockerfile .
+
+.PHONY=all
+all: web db py
+
+.PHONY=test
+test: 
+	@docker run -i --rm --network delphi-net \
+		--mount type=bind,source=$(CWD)repos/delphi/delphi-epidata,target=/usr/src/app/repos/delphi/delphi-epidata,readonly \
+		--mount type=bind,source=$(CWD)repos/delphi/delphi-epidata/src,target=/usr/src/app/delphi/epidata,readonly \
```
krivard, line 129: Why is CWD needed here?

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+		--env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" \
+		--env "FLASK_SECRET=abc" \
+		delphi_web_python python -m pytest --import-mode importlib $(pdb) $(test-subdir) | tee -a output.txt
```
krivard, line 132: This will append infinitely to output.txt; are we relying on the user to periodically clear this file?

dev/local/install.sh
```diff
@@ -1,9 +1,26 @@
...
-mkdir -p driver/repos/delphi driver-logs/delphi_database_epidata driver-logs/delphi_web_epidata
-cd driver/repos/delphi
+mkdir -p repos/delphi
+cd repos/delphi
 git clone https://github.com/cmu-delphi/operations
 git clone https://github.com/cmu-delphi/delphi-epidata
 git clone https://github.com/cmu-delphi/utils
+git clone https://github.com/cmu-delphi/flu-contest
+git clone https://github.com/cmu-delphi/nowcast
+git clone https://github.com/cmu-delphi/github-deploy-repo
```
krivard, line 14: Where are these used?

dev/local/install.sh
```diff
@@ -1,9 +1,26 @@
 #!/bin/bash
+# Bootstrap delphi-epidata
+#
+# Downloads the repos needed for local delphi-epidata development into current dir 
+# and provides a Makefile with Docker control commands.
 
-mkdir -p driver/repos/delphi driver-logs/delphi_database_epidata driver-logs/delphi_web_epidata
-cd driver/repos/delphi
+mkdir -p repos/delphi
```
krivard, line 7: I still recommend building into a `driver` directory. Reasoning:

1. scripts developed in local dev will be more easily adapted to run on staging or prod, where this directory is required
2. Dangling a bare Makefile in a random directory that may include other files is risky practice

We can however put the logs directory into driver, which will be more tidy.

docs/epidata_development.md
```diff
@@ -7,10 +7,16 @@ nav_order: 4
...
 ```
 $ curl "https://raw.githubusercontent.com/cmu-delphi/delphi-epidata/dev/dev/local/install.sh" | bash
-$ cd driver
-$ [sudo] ./epidata-refresh.sh database web python
+$ [sudo] make all
+$ make test
+# To drop into debugger on error
+$ make test pdb=1
+# To test only a subset of tests
+$ make test test-subdir=repos/delphi/delphi-epidata/integrations/acquisition
```
krivard, line 19: sudo requirement holds for running images as well as building them, alas

```suggestion
$ [sudo] make test
# To drop into debugger on error
$ [sudo] make test pdb=1
# To test only a subset of tests
$ [sudo] make test test-subdir=repos/delphi/delphi-epidata/integrations/acquisition
```

krivard (changes requested):
Great work! Changes:

* Be consistent about whether CWD is specified or not (& decide if it's even needed)
* Double-check uses of @ to hide output
* Beware creating a neverending output file
* Probably don't put a random Makefile on someone's filesystem
* Verify if all prepandemic delphi packages are really needed

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+NOW:=$(shell date "+%Y-%m-%d")
+CWD:=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+LOG_WEB:=$(CWD)docker-logs/delphi_web_epidata_$(NOW).log
+LOG_DB:=$(CWD)docker-logs/delphi_database_epidata_$(NOW).log
+
+
+.PHONY=web
+web:
+	@# Stop container if running
+	@-docker ps | grep delphi_web_epidata && docker stop delphi_web_epidata
```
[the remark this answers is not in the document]
dshemetov: The `-` makes sure that the commands continue on "failure". If `delphi_web_epidata` is not found, the command is considered a "failure". I think the technical term is non zero exit status or something like that.

dev/local/Makefile
dshemetov: I'm just in the habit of specifying full paths at this point. Some tools need full, some don't 🤷.

dev/local/Makefile
```diff
@@ -0,0 +1,136 @@
...
+
+	@# Setup virtual network if it doesn't exist
+	@-docker network ls | grep delphi-net || docker network create --driver bridge delphi-net
+
+	@# Build the web_epidata image
+	@cd repos/delphi/delphi-epidata;\
+	docker build -t delphi_web_epidata -f ./devops/Dockerfile .;\
+	cd ../../../
```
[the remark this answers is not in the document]
dshemetov: They aren't, since they're part of the command above. I'll fix the indentation to make that clearer.

dev/local/Makefile
dshemetov: This may have been one of those things that got me in the habit of using absolute paths 
```
docker: Error response from daemon: invalid mount config for type "bind": invalid mount path: 'repos/delphi/delphi-epidata' mount path must be absolute.
```

dev/local/Makefile
dshemetov: Yea, I probably should make this auto-clear.

dev/local/install.sh
dshemetov: [Here](https://github.com/cmu-delphi/operations/blob/main/dev/docker/python/assets/setup.sh). I made some issues and PRs to remove at least some of them [here](https://github.com/cmu-delphi/operations/pull/70), but I figured we'll get there on low-priority speed.

dev/local/install.sh
dshemetov: My (unstated) assumption was that the user is already in a clean directory like `../delphi-epidata-dev/` when they call this script. I agree now that it's not a good assumption to make.

My main pickiness is that the directory tree already seems unnecessarily deep:
```
delphi-epidata-dev
└── driver
    └── repos
        └── delphi
            └── delphi-epidata
```
and the `delphi` level is where finally get to any code. So it's really about just saving one level of depth for me
```
delphi-epidata-dev
└── repos
    └── delphi
        └── delphi-epidata
```
If it were up to me, I would get rid of the `delphi` directory in there too, since it's not doing much anymore (especially so when we get rid of the legacy `undefx` tools). 

But yea, it's just a minor aesthetic thing. And if prod requires `driver`, then it makes sense to make local dev reflect that. So I'll bring `driver` back and get rid of `docker-logs`.

dev/local/install.sh
dshemetov: (Relatedly, I started a very long-term, low-priority project to redo the directory structure changes that `setup.sh` file makes to the Docker image [e.g. `mv repos/delphi/operations/src delphi/operations`]. That little change makes it so that all my IDE tools can't track imports at all and breaks my "Go To Reference" tools. Could be a worthwhile experiment.)

dev/local/Makefile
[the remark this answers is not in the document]
krivard: if the container is running and `docker stop` fails, don't we want execution to halt?

dev/local/install.sh
krivard: > I wonder if those changes made in that setup script reflect the way the directory structure is on prod?

Yes -- they let the epidata acquisition system treat `delphi.epidata.acquisition` as a package without it having to be a valid (ie installable, with an `__init__.py` and `setup.py` and everything) Python package. Agree that this makes IDEs angry and angry IDEs make development needlessly difficult and error-prone, and that we should just make `delphi.epidata.acquisition` a package. Same with all other delphi "packages" that aren't really packages and only import because we bodge them on the filesystem like this.

Until we make that change, we need the CI and local dev environments to closely match production, so that we don't end up with changes that work in local dev but fail in prod.

krivard: https://github.com/cmu-delphi/delphi-epidata/issues/939

dev/local/Makefile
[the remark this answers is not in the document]
dshemetov: Hm, I suppose it could fail, though is that likely? Do containers just refuse to stop sometimes?

dshemetov:
Going to leave the refactor discussions unresolved so they're easier to find later.

dev/local/Makefile
```diff
@@ -0,0 +1,140 @@
...
+WEB_CONTAINER_ID:=$(shell docker ps -q --filter 'name=delphi_web_epidata')
+DATABASE_CONTAINER_ID:=$(shell docker ps -q --filter 'name=delphi_database_epidata')
+
+
+.PHONY=web
+web:
+	@# Stop container if running
+	@if [ $(WEB_CONTAINER_ID) ]; then\
+		docker stop $(WEB_CONTAINER_ID);\
+	fi
```
krivard, line 73: 💯 

thoughts on converting this to a one-liner? We wouldn't need the backslashes but maybe it's less readable.
```suggestion
	@[ $(WEB_CONTAINER_ID) ] && docker stop $(WEB_CONTAINER_ID)
```

krivard:
> Going to leave the refactor discussions unresolved so they're easier to find later.

Oh sorry, I resolved them. They're listed in issues and each one links back to this PR in the relevant thread for future discussion purposes.

dev/local/Makefile
dshemetov: I would, but that doesn't work (just tested). If the left expression fails, the whole expression fails.

krivard (approved):
👍

PR #633, opened 2021-07-13
API key server logic

sgratzl:
**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

implements the server logic to require API keys and roles for the endpoints. 

tasks
- [x] check for api key
- [x] show a soft warning
- [x] show a hard warning
- [x] adapt tests
- [ ] create tests
- [x] admin interface? 
- [?] log / track api_key + query if allowed
- [x] create basic google form
- [x] connect google form with api server handling -> webhook
- [x] create a basic request a key form: /admin/create_key
- [x] use [flask-limited](https://flask-limiter.readthedocs.io) for rate limiting
- [ ] setup a Redis DB for shared rate limit tracking
- [ ] define a good default rate limit
- [ ] consider using Redis DB also for account management instead of SQL server

PR #937, opened 2022-06-22
[Prototype] Add mandatory `meta_key` check to server

dshemetov:
**Prerequisites**:

- [x] Build is successful

### Summary

This adds a prototype of a basic authentication system to the API server for meta cache computations. Since the meta computations run on a multi-user machine, with read and write privileges to the main DB, we need to make it secure. The idea here is to run the API server with `user1` and run the meta cache computations with `user2`. Both users will share a secret key, which will make sure the API server only responds to requests made by `user2`. The basic logic is to expect a secret meta_key parameter in every request.

Many tests had to be fixed. I went with a simple patching route that added the missing fake `meta_key` parameter in most requests.

This will likely be deprecated by #633, but the review and roll out of that work is still a way into the future.

Creating this PR for visibility and review, though it definitely should not be merged into `dev`. I envision that this code will continue to live on a separate branch and the meta cache computations server will just run its API server from that branch. Any updates to the JIT computations will need to be propagated to this branch too, whenever they are pushed to `dev`. The diffs shouldn't be too hard to manage. We should probably automate pulling into this branch when `dev` gets updated. We will also want to automate updating the meta cache computations API server whenever this code gets updated.

For the sake of tests passing, the `meta_key` comparison only expects a `meta_key` if [more than zero parameters were requested](https://github.com/cmu-delphi/delphi-epidata/blob/c823943338975e46ca55eb40f3f9720d2ef4ee8a/src/server/_security.py#L6) from the API. So for example, requests like `https://api.covidcast.cmu.edu/epidata/covidcast/` and `https://api.covidcast.cmu.edu/epidata/` pass. For these two, the result returns no actual data, so it's fine, but I need to double check that this won't expose information in other endpoints.

### DIFF
--- a/integrations/acquisition/covid_hosp/facility/test_scenarios.py
+++ b/integrations/acquisition/covid_hosp/facility/test_scenarios.py
@@ -2,7 +2,7 @@
 
 # standard library
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 # first party
 from delphi.epidata.acquisition.covid_hosp.common.database import Database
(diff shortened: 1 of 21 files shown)


dshemetov:
Misunderstood the attack vector. A big concern is making sure that [`FLASK_SECRET`](https://github.com/cmu-delphi/delphi-epidata/blob/a3cf65b6f43202263a35401b1f51f3f984c37e39/.github/workflows/ci.yaml#L66) isn't exposed to other users. Will ping @korlaxxalrok later for help with using local secret overrides.

PR #951, opened 2022-08-19
#798 Fix bad table

### LINKED ISSUE #798: Fix bad table formatting in documentation
A lot of the tables in the Epidata documentation for the COVIDcast endpoint are badly formatted. (They may also be badly formatted in other places in the Epidata documentation, I haven't looked as carefully.)

For example, looking at the HHS data source this is what I see: 

Screen Shot 2022-01-06 at 12 09 13 PM

And when I scroll right, I get this: 

Screen Shot 2022-01-06 at 12 09 06 PM

We should fix this.

LeonLu2:
Fixes #798

one table in the documentation for the COVIDcast endpoint are badly formatted

Changes:
1. Reformat the table code
2. combine the signal and 7-day signal columns into one column
3. remove the geography and resolution columns and add a universal explanation below the table to make the description column text more visible


krivard:
@LeonLu2 it looks like this branch has received updates unrelated to #798. Was that intentional?

LeonLu2:
Closing pull request, new pr coming

PR #927, opened 2022-06-07
[Refactor] Clean server integration tests

dshemetov:
**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

### Summary

Extracts the really nice test refactors Katie did in #923. 

Why? That branch is experimental and might not be merged. We definitely want to keep these refactors, however.

Addresses #897

PR #959, opened 2022-08-26
incorporation of test improvements from parallel branch

melange396:
Updates to include improvements from "merged key dimension table" branch (`krivard/v4-rpp-mergeddim-leftjoin`), specifically at commit hash `fbf878e`.
Changes include:
 - unit/integration testing refactorization and other improvements.
 - percona dbms now used in db docker image, plus changes for resulting compatibility issues.
 - db schema names are specified in ddl files.
 - removed of obsolete index hint guessing method.
 - updated comments.

This changeset is essentially just a port of the excellent work @krivard did to refactor and otherwise improve the test architecture, as she applied it to branch mentioned above.  Most of the files were simply copied over from the other branch to this one to create this PR; I only really made edits to these files: (and most edits were to strip out "mergedkey" stuff)
 - src/ddl/v4_schema.sql
 - src/acquisition/covidcast/database.py
 - src/acquisition/covidcast/test_utils.py
 - integrations/acquisition/covidcast/test_covidcast_meta_caching.py
 - integrations/server/covidcast/test_covidcast_meta.py


integrations/server/test_covidcast_meta.py
```diff
@@ -33,15 +33,18 @@ class CovidcastMetaTests(unittest.TestCase):
...
+      INSERT INTO `signal_latest` (
+        `signal_data_id`, `signal_key_id`, `geo_key_id`,
+        `time_type`, `time_value`, `value_updated_timestamp`,
         `value`, `stderr`, `sample_size`,
         `issue`, `lag`, `missing_value`,
         `missing_stderr`,`missing_sample_size`)
       VALUES
-        (%d, %d, %d, "%s", %d, 123,
-        %d, 0, 0, %d, 0, %d, %d, %d)
+        (%d, %d, %d,
```
nolangormley, line 43: Do we have code standards set for line length and how to format things like this? Seems to change from file to file and it would be nice to have a unified truth for the project.

src/acquisition/covidcast/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        self._db.connect()
+
+        # empty all of the data tables
+        for table in "signal_load  signal_latest  signal_history  geo_dim  signal_dim".split():
+            self._db._cursor.execute(f"TRUNCATE TABLE {table};")
+        self.localSetUp()
+        self._db._connection.commit()
+
+    def localSetUp(self):
+        pass
```
nolangormley, line 26: Todo or local needs no setup? If the latter, why is it called in setup?

Also, in the case that a test fails or a test doesn't clean up after itself, it seems like it may be beneficial for the truncate to happen on the local DB too.

nolangormley (approved):
Looks good. Left 2 comments, one of them is a vague code formatting question that doesn't need to be answered. The other one about local environments is the only important one imo.

dshemetov:
Is this duplicating some of the work in #927?

integrations/server/test_covidcast_meta.py
krivard: Not in this repo, and we should! Outside the scope of this PR but [I made an issue for it. ](https://github.com/cmu-delphi/delphi-epidata/issues/962)

src/acquisition/covidcast/test_utils.py
krivard: it's a stub, available for overriding in subclasses.

krivard:
```suggestion
        # stub; override in subclasses to perform custom setup. 
        # runs after tables have been truncated but before database changes have been committed
        pass
```

krivard:
@dshemetov half maybe? it will supersede it. target of #927 is v3; target of this PR is v4

melange396:
@dshemetov kinda sorta in-a-way but not really, in that #927 duplicated work in #923, then #923 evolved into #931, and this is essentially merging all that from #931 back into the v4 development branch (without the merged key stuff).

i did give #927 a good look-through (though not in a very thorough or methodical way) but didnt see anything that seemed too different from what i got from #931.  if theres something else there that you think i need to explore, im happy to give it a look!

integrations/server/test_covidcast_meta.py
melange396: we dont have any hard rules.  generally, imnsho, the format should be something that "looks clean" (or at least is not sloppy), preferably with a structure that makes it easy to read, identifies/highlights important parts, and otherwise means something semantically (whenever possible).

this section you highlighted includes sql with column names that need to follow the same order as the values provided shortly after; thus the formatting here has line breaks that match up 1:1, so that each row of column names corresponds to a later row for its values.  that way, its easy to see (for example) that the last item of the second row of column names (`value_updated_timestamp`) refers to the last item of the second row of values (`123`).

all this test code has come a long way, but there is still more room for improvement...  see #897 

for this pr, i tried not to change anything where i didnt have to, to keep the [diff between this and source material](https://github.com/cmu-delphi/delphi-epidata/compare/fbf878e..0d4a2e7) as small as possible.  the intent was to preserve the spirit of the test refactoring & improvements from the other branch without complicating it by introducing other changes.  however, i goofed a couple times and introduced some whitespace changes but ill claim they were for the best ¯\\\_(ツ)_/¯

src/acquisition/covidcast/test_utils.py
melange396: this method is a placeholder, and subclasses of this base class can redefine/override it if they choose.  that way, those subclasses can specify "local" set-up procedures that are applicable only to themselves and would be meaningless (or even inappropriate) for a different test class.  its sort of like an abstract method, but optional.

any expectation for "cleaning up" the state of the tables is really covered in `setUp()`: it runs before each test case and provides a clean slate for that test.  anything that test leaves behind (populated rows) will be removed for the next test when `setUp()` runs again.

melange396: @krivard im fine with adding that suggested change, especially because its small and only documentation, but no more [riders](https://en.wikipedia.org/wiki/Rider_(legislation))!

dshemetov:
Gotcha, yea, I didn't add anything extra in #927 that wasn't in #923. I'll close that one.

PR #960, opened 2022-08-26
#797 Other endpoints issue

### LINKED ISSUE #797: Other Diseases --> Other Endpoints?
One of the headings in the table of contents on the left in the [Epidata documentation](https://cmu-delphi.github.io/delphi-epidata/api/README.html) says "Epidata API (Other Diseases)".  

I'm not a fan of "Diseases" because there is indeed COVID data here, e.g., all the HHS endpoints. And not every endpoint is its own disease. So how about just "Other Endpoints"? @RoniRos @capnrefsmmat @krivard

LeonLu2:
Issue: Some headings in the table of contents in the Epidata Documentation are inappropriate, need changes and re-arrange the sequence of the table.

Fixes #797 

Changelog:

1. change the headings in the table, and all of the usages of the headings in their children's file
2. re-arrange their sequence
3. waiting for the coming version of Just the Docs, with the feature of unlimited nested relationship which is still under development

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1,6 +1,6 @@
 ---
 title: Other Endpoints (COVID-19 and Other Diseases)
-nav_order: 3
+nav_order: 2
 has_children: true
 ---
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1,5 +1,5 @@
 ---
-title: Epidata API (Other Diseases)
+title: Other Endpoints (COVID-19 and Other Diseases)
 nav_order: 3
 has_children: true
 ---
--- a/docs/api/afhsb.md
+++ b/docs/api/afhsb.md
@@ -1,6 +1,6 @@
 ---
 title: AFHSB
-parent: Epidata API (Other Diseases)
+parent: Other Endpoints (COVID-19 and Other Diseases)
 ---
 
 # AFHSB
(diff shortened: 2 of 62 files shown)


dshemetov:
Hey thanks for doing this work!

I'm not sure how to compile / host this. Could you share a screenshot of the new nav order?

LeonLu2:
Sure! Here it is: 
image

dshemetov:
Thanks @LeonLu2. So this PR mostly looks good to me. My personal preference would be to have the "COVID-19 Trends and Impact Survey" section be higher, either below "Other Endpoints" or keep it below "Main Endpoint". The reason being that, even though it's no longer being updated, I suspect that it will still generate a lot of retrospective interest from researchers. So we want to keep it with the "data sections" and above the "development" sections.

krivard:
> So we want to keep it with the "data sections" and above the "development" sections.

I agree the current placement isn't ideal, but I'm not sure the change you propose is better -- the CTIS section is about the survey itself and doesn't mention the API at all. That would suggest that API development and API endpoint documentation should be clustered together, with CTIS set apart. You can see the effect of that in Roni's comment in #797, but since we can only have three levels in the hierarchy, this is what Leon and I came up with as a next-best.

dshemetov:
Ah, thanks for the context! Sounds good to me then.

LeonLu2:
1. merged dev into current branch, conflicts resolved. Ready to pull

krivard (approved):
👍

Issue #910, opened 2022-05-23
CHNG documentation should indicate sample size is not provided


capnrefsmmat:
The CHNG documentation (and documentation for any other source with suppressed sample size) should document that the `sample_size` column is left deliberately blank, and give whatever reason we can share for that.

krivard:
@LeonLu2 these are all covidcast signals, so you know where to find the documentation sources for them already.

Match disclaimer and reason with doctor visits: "Standard errors and sample sizes are not available for this data source."

- [ ] CHNG
- [ ] Hospital Admissions
- [ ] Google Search Trends

Disclaimer and reason: "Standard errors and sample sizes are not applicable to these metrics."

- [ ] Data Strategy and Execution Workgroup (DSEW)
- [ ] JHU
- [ ] USAFacts
- [ ] Department of Health and Human Services (HHS)
- [ ] NCHS Mortality

krivard:
Fixed in #961

PR #961, opened 2022-08-26
#910 CHNG-documentation

LeonLu2:
Issue: The CHNG documentation (and documentation for any other source with suppressed sample size) should document that the sample_size column is left deliberately blank, and provide reasons.

Fixes:
1. add the reason of "Standard errors and sample sizes are not available/applicable for this data source/metrics." in all there pages limitation sections.


krivard (approved):
🎉 👍

PR #963, opened 2022-08-29
renamed v4 db objects: load, latest, and history tables, and their id columns

melange396:
src/ddl/migrations/v4_renaming.sql - contains the SQL to do the renaming of a live v4 system

source code changes were done by these 4 shell commands:
```{bash}
  find ./src ./tests ./integrations -type f -exec sed -i 's/signal_history/epimetric_full/g' {} \;
  find ./src ./tests ./integrations -type f -exec sed -i 's/signal_latest/epimetric_latest/g' {} \;
  find ./src ./tests ./integrations -type f -exec sed -i 's/signal_load/epimetric_load/g' {} \;
  find ./src ./tests ./integrations -type f -exec sed -i 's/signal_data_id/epimetric_id/g' {} \;
```

all tests pass on the codebase as it stands in this commit;
they also pass on a `delphi_web_epidata` docker image that was created before this commit and then had the migration file changes applied to it.


melange396:
note: the `find` commands were designed to avoid the `.git/` directory

src/ddl/migrations/v4_renaming.sql
```diff
@@ -0,0 +1,99 @@
+
+-- drop VIEWs in `epidata` that act as aliases to (ie, they reference) VIEWs in `covid`
+USE epidata;
+DROP VIEW
+    epimetric_full_v,
```
krivard, line 5: Should this be `signal_history_v`?

melange396: hahahaha, good catch!  i ran the renaming commands on that migration script.  fixed now.  *facepalm*

src/ddl/v4_schema_aliases.sql
```diff
@@ -5,6 +5,6 @@
 --       frontend api code still uses `epidata` but has these relevant tables/views "aliased" to use covid.blah when referred to as epidata.blah in context.
 -- ----------------------------------
 
-CREATE VIEW `epidata`.`signal_history_v`     AS SELECT * FROM `covid`.`signal_history_v`;
-CREATE VIEW `epidata`.`signal_latest_v`      AS SELECT * FROM `covid`.`signal_latest_v`;
+CREATE VIEW `epidata`.`epimetric_full_v`     AS SELECT * FROM `covid`.`epimetric_full_v`;
+CREATE VIEW `epidata`.`epimetric_latest_v`      AS SELECT * FROM `covid`.`epimetric_latest_v`;
 CREATE VIEW `epidata`.`covidcast_meta_cache` AS SELECT * FROM `covid`.`covidcast_meta_cache`;
```
krivard, line 10: whitespace nit
```suggestion
CREATE VIEW `epidata`.`epimetric_full_v`     AS SELECT * FROM `covid`.`epimetric_full_v`;
CREATE VIEW `epidata`.`epimetric_latest_v`   AS SELECT * FROM `covid`.`epimetric_latest_v`;
CREATE VIEW `epidata`.`covidcast_meta_cache` AS SELECT * FROM `covid`.`covidcast_meta_cache`;
```

krivard (approved):
lgtm! 👍

PR #964, opened 2022-08-30
Metadata threads and ANALYZE and etc

melange396:
4 commits, each for:
 - [cli option for specifying # of metadata worker threads](https://github.com/cmu-delphi/delphi-epidata/commit/0d4188b95bde9c4e162c7ec902b7c6c0ff9f569e)
 - [added flag to Makefile for new percona db image (and some whitespace fixes)](https://github.com/cmu-delphi/delphi-epidata/commit/5e3eb51c005720a9da1e52ee78fb1cbfc0339516)
 - [converted JOINs in VIEWs to 'USING' instead of 'ON'](https://github.com/cmu-delphi/delphi-epidata/commit/b9ab3f3fe09fb6c015baece120e68fa591c06a4b)
 - [added ANALYZE TABLE to the end of acquisition runs (and some whitespace fixes)](https://github.com/cmu-delphi/delphi-epidata/commit/a61a5db5b830cc6c0c124c6ee3374264970c1976)

### DIFF
--- a/src/acquisition/covidcast/covidcast_meta_cache_updater.py
+++ b/src/acquisition/covidcast/covidcast_meta_cache_updater.py
@@ -15,6 +15,7 @@ def get_argument_parser():
 
   parser = argparse.ArgumentParser()
   parser.add_argument("--log_file", help="filename for log output")
+  parser.add_argument("--num_threads", type=int, help="number of worker threads to spawn for processing of each source/signal pair")
   return parser
 
 
@@ -24,8 +25,10 @@ def main(args, epidata_impl=Epidata, database_impl=Database):
   `args`: parsed command-line arguments
   """
   log_file = None
+  num_threads = None
   if (args):
     log_file = args.log_file
+    num_threads = args.num_threads
 
   logger = get_structured_logger(
       "metadata_cache_updater",
@@ -37,7 +40,7 @@ def main(args, epidata_impl=Epidata, database_impl=Database):
   # fetch metadata
   try:
     metadata_calculation_start_time = time.time()
-    metadata = database.compute_covidcast_meta()
+    metadata = database.compute_covidcast_meta(n_threads=num_threads)
     metadata_calculation_interval_in_seconds = time.time() - metadata_calculation_start_time
   except:
     # clean up before failing
--- a/src/acquisition/covidcast/covidcast_meta_cache_updater.py
+++ b/src/acquisition/covidcast/covidcast_meta_cache_updater.py
@@ -15,7 +15,7 @@ def get_argument_parser():
 
   parser = argparse.ArgumentParser()
   parser.add_argument("--log_file", help="filename for log output")
-  parser.add_argument("--num_threads", type=int, help="number of worker threads to spawn for processing of each source/signal pair")
+  parser.add_argument("--num_threads", type=int, help="number of worker threads to spawn for processing source/signal pairs")
   return parser
 
 
--- a/src/acquisition/covidcast/csv_to_database.py
+++ b/src/acquisition/covidcast/csv_to_database.py
@@ -77,12 +77,12 @@ def upload_archive(
     csv_importer_impl=CsvImporter):
   """Upload CSVs to the database and archive them using the specified handlers.
 
-  :path_details: output from CsvImporter.find*_csv_files 
-  
+  :path_details: output from CsvImporter.find*_csv_files
+
   :database: an open connection to the epidata database
 
   :handlers: functions for archiving (successful, failed) files
-  
+
   :return: the number of modified rows
   """
   archive_as_successful, archive_as_failed = handlers
@@ -130,7 +130,7 @@ def upload_archive(
       archive_as_successful(path_src, filename, source, logger)
     else:
       archive_as_failed(path_src, filename, source,logger)
-  
+
   return total_modified_row_count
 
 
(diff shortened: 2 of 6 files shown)


src/acquisition/covidcast/covidcast_meta_cache_updater.py
```diff
@@ -15,6 +15,7 @@ def get_argument_parser():
 
   parser = argparse.ArgumentParser()
   parser.add_argument("--log_file", help="filename for log output")
+  parser.add_argument("--num_threads", type=int, help="number of worker threads to spawn for processing of each source/signal pair")
```
krivard, line 18: (number of threads for the group of pairs, not number of threads per pair)
```suggestion
  parser.add_argument("--num_threads", type=int, help="number of worker threads to spawn for processing source/signal pairs")
```

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -161,13 +161,12 @@ def main(
       database,
       make_handlers(args.data_dir, args.specific_issue_date),
       logger)
-    logger.info("Finished inserting database rows", row_count = modified_row_count)
-    # the following print statement serves the same function as the logger.info call above
-    # print('inserted/updated %d rows' % modified_row_count)
+    logger.info("Finished inserting/updating database rows", row_count = modified_row_count)
   finally:
+    database.do_analyze()
```
krivard, line 166: Do we want the database to run analyze when there's been an exception?

src/acquisition/covidcast/database.py
```diff
@@ -138,6 +138,17 @@ def _reset_load_table_ai_counter(self):
...
+  def do_analyze(self):
+    """performs and stores key distribution analyses, used for join order and index selection"""
+    # TODO: consider expanding this to update columns' histograms
+    #       https://dev.mysql.com/doc/refman/8.0/en/analyze-table.html#analyze-table-histogram-statistics-analysis
+    self._cursor.execute(
+      f'''ANALYZE TABLE
+          signal_dim, geo_dim,
+          {self.load_table}, {self.history_table}, {self.latest_table}''')
+    output = [self._cursor.column_names] + self._cursor.fetchall()
+    get_structured_logger('do_analyze').info("ANALYZE results: "+str(output))
```
krivard, line 150: does it make sense to make use of structure here, a la the following?
```suggestion
    get_structured_logger('do_analyze').info("ANALYZE results", results=str(output))
```

src/ddl/migrations/v4_renaming.sql
```diff
@@ -54,10 +54,8 @@ CREATE OR REPLACE VIEW epimetric_full_v AS
         `t1`.`signal_key_id` AS `signal_key_id`,
         `t1`.`geo_key_id` AS `geo_key_id`
     FROM `epimetric_full` `t1`
-        JOIN `signal_dim` `t2`
-            ON `t1`.`signal_key_id` = `t2`.`signal_key_id`
-        JOIN `geo_dim` `t3`
-            ON `t1`.`geo_key_id` = `t3`.`geo_key_id`;
+        JOIN `signal_dim` `t2` USING (`signal_key_id`)
+        JOIN `geo_dim` `t3` USING (`geo_key_id`);
```
krivard, line 58: 🏆

krivard (commented):
clarification questions

src/acquisition/covidcast/csv_to_database.py
melange396: in short, "yes".  `upload_archive` calls `commit` or `rollback` for each file, so an exception does not indicate that rows were not added (note the double negative there).  thus, running an ANALYZE is still prudent.  ANALYZE is also fast/inexpensive to run, just in case.

relatedly, there is also some older technical debt here that i was kicking further down the road -- the call below does not need the `True` argument (to force a call to `commit()`) and the associated comment about committing is a misnomer.  i left them there for the sake of getting this ANALYZE call into the acquisition process faster.  because of the aforementioned commit/rollback calls in `upload_archive`, the True+comment bellow are irrelevant and can be removed.  i am happy to do that.

src/ddl/migrations/v4_renaming.sql
melange396: 🦾

src/acquisition/covidcast/database.py
melange396: probably?  its mostly cursory though.  it will look like `[('Table', 'Op', 'Msg_type', 'Msg_text'), ('covid.signal_dim', 'analyze', 'status', 'OK'), ('covid.geo_dim', 'analyze', 'status', 'OK'), ('covid.epimetric_load', 'analyze', 'status', 'OK'), ('covid.epimetric_full', 'analyze', 'status', 'OK'), ('covid.epimetric_latest', 'analyze', 'status', 'OK')] ` in the standard/typical case.  i dont know what it means if it does not output that, and i dont know how to test it to make it output anything but that.  if there is some sequence of events that would cause a different output, it would be indicative of a deeper issue or error that we will hopefully be notified of somewhere other than here.

src/acquisition/covidcast/csv_to_database.py
krivard: that helps! no changes needed

krivard (approved):
👍 nice work!

nolangormley (approved):
Looks good!

Issue #941, opened 2022-06-28
v4: insert_batch should fail if you try to run it with something already in signal_load


krivard:
[We recently repaired the mergedkey branch of v4 to force it to run dbjobs automatically after insert_batch](https://github.com/cmu-delphi/delphi-epidata/pull/940). This is because [the fix_islatest query](https://github.com/cmu-delphi/delphi-epidata/blob/fbf878e70d3f4756fc3b7289f980799f7e06f0d4/src/acquisition/covidcast/database.py#L153-L163), having been stripped of the process_status column, performs quite poorly when you wait to run dbjobs until all the CSVs have been run through insert_batch. Each additional CSV file added to signal_load increases the number of rows participating in the fix_islatest join, but a smaller and smaller percentage of rows are updated.

We want to:
* Add a check to insert_batch so that if signal_load has anything in it, it throws an exception
* Add a test (could be an integration test if you did the database setup, or a unit test if you mocked out the database responses) to confirm that insert_batch throws an exception if you run it with something in signal_load

melange396:
i have a call to `logger.fatal` [here](https://github.com/cmu-delphi/delphi-epidata/blob/1fe5b9c88cc315e9c6b60c0491dc48ca7e31dce4/src/acquisition/covidcast/database.py#L176) which i somehow boneheadedly thought would actually halt execution.  this needs to raise an exception at the very least.  it looks like @xavier-xia-99 might have gotten this done in #948 but i cant make sense of the mass of changes in there...  this should be included in v4 before it goes live, so ill get to adding the exception shortly.

krivard:
Fixed in #967

PR #967, opened 2022-08-31
DBLoadStateException: halt acquisition when unexpected data found in load table

melange396:
addresses #941

a question remains about how to handle the new exception in `csv_to_database.py:upload_archive` -- i would argue the behavior i have here is correct: it stops processing any more files and leaves them where they are.  the exception condition would arise because of a problem with the database (and not with the files) so it should not handle unprocessed files with `archive_as_failed`.  if this breaks some filesystem maintenance paradigm though, im fine with changing it.

### DIFF
--- a/integrations/acquisition/covidcast/test_db.py
+++ b/integrations/acquisition/covidcast/test_db.py
@@ -1,7 +1,7 @@
 import unittest
 
 from delphi_utils import Nans
-from delphi.epidata.acquisition.covidcast.database import Database, CovidcastRow
+from delphi.epidata.acquisition.covidcast.database import Database, CovidcastRow, DBLoadStateException
 from delphi.epidata.acquisition.covidcast.test_utils import CovidcastBase
 import delphi.operations.secrets as secrets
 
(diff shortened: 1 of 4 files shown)


integrations/acquisition/covidcast/test_db.py
```diff
@@ -29,6 +29,28 @@ def _find_matches_for_row(self, row):
...
+        a_row = self._make_placeholder_row()[0]
+        another_row = self._make_placeholder_row(time_value=self.DEFAULT_TIME_VALUE+1, issue=self.DEFAULT_ISSUE+1)[0]
+        # insert one
+        self._db.insert_or_update_bulk([a_row])
+        # put something into the load table
+        self._db._cursor.execute(f'''
+            INSERT INTO {self._db.load_table}
+            (source, `signal`, geo_type, geo_value, time_type, time_value, issue, `lag`, value_updated_timestamp)
+            VALUES
+            ('sr',   'si',     'gt',     'gv',      'tt',      3,          8,     5,     13)''')
```
krivard, line 43: hmmm i kinda wanna find a way to refactor this out just so we don't have to continually mess with it when we make more changes. did you already look at that?

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -120,6 +120,10 @@ def upload_archive(
         if modified_row_count is None or modified_row_count: # else would indicate zero rows inserted
           total_modified_row_count += (modified_row_count if modified_row_count else 0)
           database.commit()
+      except DBLoadStateException as e:
+        # if the db is in a state that is not fit for loading new data,
+        # then we should stop processing any more files
+        raise e
```
krivard, line 126: Noting for posterity: we don't need a `database.rollback()` here because if this exception is thrown, no changes have yet been made to the database.

tests/acquisition/covidcast/test_database.py
```diff
@@ -94,6 +94,7 @@ def test_insert_or_update_batch_row_count_returned(self):
     """Test that the row count is returned"""
     mock_connector = MagicMock()
     database = Database()
+    database.count_all_load_rows = lambda:0 # simulate an empty load table
```
krivard, line 97: good catch + good fix

krivard (approved):
Assuming the SQL in test_db is begrudging, 🚀

xavier-xia-99 (approved):

tests/acquisition/covidcast/test_database.py
xavier-xia-99: This was neat!

integrations/acquisition/covidcast/test_db.py
melange396: twas indeed quite begrudging.  ive thought about a better way to handle it, but i cant really come up with something that bothers me less.  i also had to do something similar with the AI counter fix method [here](https://github.com/cmu-delphi/delphi-epidata/blob/c890712333cbfa2a519bb2644db0517a869b2b78/src/acquisition/covidcast/database.py#L131)

src/acquisition/covidcast/csv_to_database.py
melange396: *yet been made to the db by the file in question, files before this _may_ have been imported correctly

krivard: wait, really? in theory this exception can only be thrown if acquisition is started when the database is in an inconsistent state. under what scenarios would some files be uploaded successfully before this exception is thrown?

melange396: its exceedingly unlikely.  but for example, multiple acquisition processes could be running simultaneously

melange396: ...though transaction boundaries would make that virtually impossible

PR #977, opened 2022-09-22
Use BETWEEN queries in the API as much as possible

### LINKED ISSUE #790: API should use 'BETWEEN' queries as much as possible
The API retrieves data from the database by specifying date criteria in two different ways: in an enumerated list of days 'OR'ed together as well as a range of dates using the SQL 'BETWEEN' syntax (ie, `t.time_value = 20201201 OR t.time_value = 20201202 OR ... OR t.time_value = 20201230 OR t.time_value = 20201231` versus `t.time_value BETWEEN 20201201 AND 20201231`).  Analysis shows the 'BETWEEN' queries appear to run 1.5x-2x faster than those with enumerated days.  We should refactor to remove the enumeration as much as possible.

rzats:
Closes #790.

**Prerequisites**:

- [x] Build is successful

### Summary

Currently, the API works with mixed lists of dates and date ranges for date-based filtering, e.g.:
```
[20200101, 20200102, (20200101, 20200104), 20200106, 20200107, 20200108]
```

This PR adds a utility function, `dates_to_ranges`, to optimize those lists, grouping consecutive dates into ranges (which in SQL are represented by a single `BETWEEN` filter rather than multiple `OR` filters) and removing duplicate dates.
e.g., after the function is applied, the mixed list above becomes:
```
[(20200101, 20200104), (20200106, 20200108)]
```
The first two values from the original list are merged into a range, and so are the last three.

### DIFF
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -22,6 +22,7 @@ from ._printer import create_printer, APrinter
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
+from .utils import dates_to_ranges
 
 
 def date_string(value: int) -> str:
@@ -89,7 +90,8 @@ def filter_dates(
     param_key: str,
     params: Dict[str, Any],
 ):
-    return filter_values(field, values, param_key, params, date_string)
+    ranges = dates_to_ranges(values)
+    return filter_values(field, ranges, param_key, params, date_string)
 
 
 def filter_fields(generator: Iterable[Dict[str, Any]]):
@@ -185,7 +187,8 @@ def filter_time_pairs(
         params[type_param] = pair.time_type
         if isinstance(pair.time_values, bool) and pair.time_values:
             return f"{type_field} = :{type_param}"
-        return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], pair.time_values), type_param, params)})"
+        ranges = dates_to_ranges(pair.time_values)
+        return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], ranges), type_param, params)})"
 
     parts = [filter_pair(p, i) for i, p in enumerate(values)]
 
--- a/src/server/utils/dates.py
+++ b/src/server/utils/dates.py
@@ -5,8 +5,6 @@ from typing import (
     Union
 )
 from datetime import date, timedelta
-from operator import itemgetter
-from itertools import groupby
 from epiweeks import Week, Year
 
 
--- a/src/server/utils/dates.py
+++ b/src/server/utils/dates.py
@@ -85,11 +85,14 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) -> 
         return values
 
     # determine whether the list is of days (YYYYMMDD) or weeks (YYYYWW) based on first element
-    if (isinstance(values[0], tuple) and guess_time_value_is_day(values[0][0]))\
-        or (isinstance(values[0], int) and guess_time_value_is_day(values[0])):
-        return days_to_ranges(values)
-    else:
-        return weeks_to_ranges(values)
+    try:
+        if (isinstance(values[0], tuple) and guess_time_value_is_day(values[0][0]))\
+            or (isinstance(values[0], int) and guess_time_value_is_day(values[0])):
+            return days_to_ranges(values)
+        else:
+            return weeks_to_ranges(values)
+    except:
+        return values
 
 def days_to_ranges(values: Sequence[Union[Tuple[int, int], int]]) -> Sequence[Union[Tuple[int, int], int]]:
     dates = []
--- a/src/server/utils/dates.py
+++ b/src/server/utils/dates.py
@@ -95,61 +95,71 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) -> 
         return values
 
 def days_to_ranges(values: Sequence[Union[Tuple[int, int], int]]) -> Sequence[Union[Tuple[int, int], int]]:
-    dates = []
+    intervals = []
 
-    # populate list of days
+    # populate list of intervals based on original values
     for v in values:
         if isinstance(v, int):
-            dates.append(time_value_to_date(v))
+            # 20200101 -> [20200101, 20200101]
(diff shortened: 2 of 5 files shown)


src/server/utils/dates.py
```diff
@@ -67,3 +74,82 @@ def weeks_in_range(week_range: Tuple[int, int]) -> int:
...
+        if isinstance(v, int):
+            dates.append(time_value_to_date(v))
+        else: # tuple
+            start_date = time_value_to_date(v[0])
+            end_date = time_value_to_date(v[1])
+            # add all dates between start and end, inclusive of end date
+            dates.extend([start_date + timedelta(n) for n in range(int((end_date - start_date).days) + 1)])
+
+    # remove duplicates then sort
+    dates = sorted(list(set(dates)))
```
dshemetov, line 111: issue: this could be potentially slow if someone requests a range like 19000101-20200101 (and there is a place in the [codebase](https://github.com/cmu-delphi/delphi-epidata/blob/ds/jit-update-meta/src/acquisition/covidcast/database_meta.py#L330-L348) where I have to do that right now, because I can't use `time_value=*` in JIT for complicated reasons...).

suggestion: have you thought about a slightly different approach where you convert everything to a range (e.g. 20220101 -> (20220101, 20220101)), sort, and then write some logic to merge overlapping / neighboring ranges? there could probably be an interesting approach with `groupby` there.

if it seems too complicated, I'm ok with filing this as a separate enhancement issue and coming back to it. up to you.

src/server/utils/dates.py
```diff
@@ -67,3 +74,82 @@ def weeks_in_range(week_range: Tuple[int, int]) -> int:
...
+        else: # tuple
+            start_week = week_value_to_week(v[0])
+            end_week = week_value_to_week(v[1])
+            # add all weeks between start and end, inclusive of end week
+            while start_week <= end_week:
+                weeks.append(start_week)
+                start_week += 1
+    
+    # remove duplicates then sort
+    weeks = sorted(list(set(weeks)))
```
dshemetov, line 142: efficiency concerns probably less important here, but would be good to have the logic in these functions be the same (if we do change the function above) for symmetry.

dshemetov (changes requested):
looks good! have a small efficiency question / alternative route suggestion, but other than that, looks great 👍

src/server/utils/dates.py
rzats: That's an excellent suggestion! I implemented your alternate approach and reworking the logic this way lowered the execution time for your edge case (1900-2050) from ~89 ms to 0.1 ms :)

dshemetov (approved):
excellent! 🚀

src/server/utils/dates.py
```diff
@@ -67,3 +76,95 @@ def weeks_in_range(week_range: Tuple[int, int]) -> int:
...
+            return days_to_ranges(values)
+        elif (isinstance(values[0], tuple) and guess_time_value_is_week(values[0][0]))\
+            or (isinstance(values[0], int) and guess_time_value_is_week(values[0])):
+            return weeks_to_ranges(values)
+        else:
+            return values
+    except:
+        return values
+
+def days_to_ranges(values: Sequence[Union[Tuple[int, int], int]]) -> Sequence[Union[Tuple[int, int], int]]:
```
melange396, line 102: theres enough complexity and similarity in both `xs_to_ranges()` that it makes sense to refactor -- have them simply pass in handles to the encoder+decoder (`time_value_to_date`+`date_to_time_value` vs `week_value_to_week`+`week_to_time_value`) and the "increment" (`timedelta(days=1)` vs `1`) to a base method.

src/server/utils/dates.py
```diff
@@ -67,3 +76,95 @@ def weeks_in_range(week_range: Tuple[int, int]) -> int:
...
+    # populate list of intervals based on original values
+    for v in values:
+        if isinstance(v, int):
+            # 20200101 -> [20200101, 20200101]
+            intervals.append([time_value_to_date(v), time_value_to_date(v)])
+        else: # tuple
+            # (20200101, 20200102) -> [20200101, 20200102]
+            intervals.append([time_value_to_date(v[0]), time_value_to_date(v[1])])
+
+    intervals.sort(key=lambda x: x[0])
```
melange396, line 114: you can get rid of that argument to `sort` and itll do a proper lexicographic ordering.  im pretty sure you should then be able set `merged[-1][1] = interval[1]` below instead of doing the max...  removing the max wont always work if you keep that sort key there though!

melange396:
great job!  i just have a few things to add.  please forgive me if im being a stickler, but i have a lot of concerns with this part of the codebase and i want to help make it better.  this is a great segue to getting some improvements put in there.

src/server/utils/dates.py
```diff
@@ -23,6 +28,10 @@ def guess_time_value_is_day(value: int) -> bool:
     # YYYYMMDD type and not YYYYMM
     return len(str(value)) > 6
 
+def guess_time_value_is_week(value: int) -> bool:
```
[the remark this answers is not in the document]
rzats: I've noticed some mentions of a `YYYY-MM-DD` format in the codebase, which is 10 characters - guessing that this is the reason for the "greater than 6" check.

src/server/utils/dates.py
rzats: Tried this, broke some tests. I think it's safe to keep the logic as-is (:

src/server/_query.py
```diff
@@ -185,7 +187,8 @@ def filter_pair(pair: TimePair, i) -> str:
         params[type_param] = pair.time_type
         if isinstance(pair.time_values, bool) and pair.time_values:
             return f"{type_field} = :{type_param}"
-        return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], pair.time_values), type_param, params)})"
+        ranges = dates_to_ranges(pair.time_values)
```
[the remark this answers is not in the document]
rzats: I tried this:
```
ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)
```
but it fails on one of the integration tests, which has inputs that are neither valid weeks nor valid ranges (`0-999999`). `dates_to_ranges` has some extra validation to account for that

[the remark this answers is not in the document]
melange396: `dates_to_ranges` works for that strange integration test case not because of extra validation, but because it passes back the input values through as-is when validation fails.  if you wrap the guts of `_to_ranges()` in a `try` (or at least wrap the part that populates the `intervals` list) and return the unmodified input on an exception, i think the test should pass.  you can also remove the try from `dates_to_ranges` after making that change.

PR #988, opened 2022-10-03
Tweaks to datetime optimization

rzats:
**Prerequisites**:

- [x] Build is successful

### Summary

A followup PR to https://github.com/cmu-delphi/delphi-epidata/pull/977, this makes a few changes to the new datetime optimization code:
- merges `days_to_ranges` and `weeks_to_ranges` into a single `_to_ranges` function with `Callable` arguments
- adds extra tests for some edge cases, as well as `TimePair` objects
- other minor changes

### DIFF
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -22,7 +22,7 @@ from ._printer import create_printer, APrinter
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
-from .utils import time_values_to_ranges
+from .utils import time_values_to_ranges, days_to_ranges, weeks_to_ranges
 
 
 def date_string(value: int) -> str:
@@ -187,7 +187,7 @@ def filter_time_pairs(
         params[type_param] = pair.time_type
         if isinstance(pair.time_values, bool) and pair.time_values:
             return f"{type_field} = :{type_param}"
-        ranges = time_values_to_ranges(pair.time_values)
+        ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)
         return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], ranges), type_param, params)})"
 
     parts = [filter_pair(p, i) for i, p in enumerate(values)]
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -187,7 +187,7 @@ def filter_time_pairs(
         params[type_param] = pair.time_type
         if isinstance(pair.time_values, bool) and pair.time_values:
             return f"{type_field} = :{type_param}"
-        ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)
+        ranges = dates_to_ranges(pair.time_values)
         return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], ranges), type_param, params)})"
 
     parts = [filter_pair(p, i) for i, p in enumerate(values)]
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -22,7 +22,7 @@ from ._printer import create_printer, APrinter
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
-from .utils import dates_to_ranges, days_to_ranges, weeks_to_ranges
+from .utils import time_values_to_ranges
 
 
 def date_string(value: int) -> str:
@@ -90,7 +90,7 @@ def filter_dates(
     param_key: str,
     params: Dict[str, Any],
 ):
-    ranges = dates_to_ranges(values)
+    ranges = time_values_to_ranges(values)
     return filter_values(field, ranges, param_key, params, date_string)
 
 
@@ -187,7 +187,7 @@ def filter_time_pairs(
         params[type_param] = pair.time_type
         if isinstance(pair.time_values, bool) and pair.time_values:
             return f"{type_field} = :{type_param}"
-        ranges = dates_to_ranges(pair.time_values)
+        ranges = time_values_to_ranges(pair.time_values)
         return f"({type_field} = :{type_param} AND {filter_integers(time_field, cast(Sequence[Union[int, Tuple[int,int]]], ranges), type_param, params)})"
 
     parts = [filter_pair(p, i) for i, p in enumerate(values)]
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -22,7 +22,7 @@ from ._printer import create_printer, APrinter
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
-from .utils import dates_to_ranges
+from .utils import dates_to_ranges, days_to_ranges, weeks_to_ranges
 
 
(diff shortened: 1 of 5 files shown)


src/server/utils/dates.py
```diff
@@ -88,63 +89,33 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
...
-        if m[0] == m[1]:
-            ranges.append(date_to_time_value(m[0]))
-        else:
-            ranges.append((date_to_time_value(m[0]), date_to_time_value(m[1])))
-
-    return ranges
+    return _to_ranges(values, time_value_to_date, date_to_time_value, lambda x: x - timedelta(days=1))
 
 def weeks_to_ranges(values: Sequence[Union[Tuple[int, int], int]]) -> Sequence[Union[Tuple[int, int], int]]:
+    return _to_ranges(values, week_value_to_week, week_to_time_value, lambda x: x - 1)
```
dshemetov, line 106: Can this decrement function correctly handle the first week of they year, like `202201`?

melange396: if x is a `Week` object, then it should

src/server/utils/dates.py
```diff
@@ -88,63 +89,33 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
...
+
+def _to_ranges(values: Sequence[Union[Tuple[int, int], int]], value_to_date: Callable, date_to_value: Callable, decrement: Callable) -> Sequence[Union[Tuple[int, int], int]]:
```
melange396, line 108:
```suggestion
def _to_ranges(values: Sequence[Union[Tuple[int, int], int]], value_to_date: Callable, date_to_value: Callable, time_unit: Union[int, timedelta]) -> Sequence[Union[Tuple[int, int], int]]:
```

the idea is to put as much common code into `_to_ranges` as possible, and to simplify the methods that call it as much as possible.  putting lambdas in the one-liners takes away from readability, and specifying just the unit to subtract is clearer.

src/server/utils/dates.py
```diff
@@ -153,18 +124,18 @@ def weeks_to_ranges(values: Sequence[Union[Tuple[int, int], int]]) -> Sequence[U
     for interval in intervals:
         # no overlap, append the interval
         # caveat: we subtract 1 from interval[0] so that contiguous intervals are considered overlapping. i.e. [1, 1], [2, 2] -> [1, 2]
-        if not merged or merged[-1][1] < interval[0] - 1:
+        if not merged or merged[-1][1] < decrement(interval[0]):
```
melange396, line 127:
```suggestion
        if not merged or merged[-1][1] < interval[0] - time_unit:
```

(continuation of suggestion above)

src/server/utils/dates.py
```diff
@@ -88,63 +89,33 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
...
-            intervals.append([week_value_to_week(v), week_value_to_week(v)])
+            # 20200101 -> [20200101, 20200101]
+            intervals.append([value_to_date(v), value_to_date(v)])
         else: # tuple
-            # (202001, 202002) -> [202001, 202002]
-            intervals.append([week_value_to_week(v[0]), week_value_to_week(v[1])])
+            # (20200101, 20200102) -> [20200101, 20200102]
+            intervals.append([value_to_date(v[0]), value_to_date(v[1])])
 
     intervals.sort(key=lambda x: x[0])
```
melange396, line 120:
```suggestion
    intervals.sort()
```

providing the sort key here is an unnecessary complication which can confuse the reader.

src/server/_query.py
```diff
@@ -22,7 +22,7 @@
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
-from .utils import dates_to_ranges
+from .utils import dates_to_ranges, days_to_ranges, weeks_to_ranges
```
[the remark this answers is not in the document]
melange396: also, if we arent going to use the other two methods here, they should come out of the import statement.

melange396 (changes requested):
lookin good!  im sorry if it seems like im picking at nits, but this is a nucleation point for changes that sweep through this call stack, and we will be able to make a lot of changes for the better soon!

i also have a [new comment](https://github.com/cmu-delphi/delphi-epidata/pull/977/files#r986447003) regarding changes to `src/server/_query.py` in the previous PR.

src/server/utils/dates.py
rzats: Yes, the "-" here is a logical operation in the `epiweeks` package, not naive integer subtraction - see https://epiweeks.readthedocs.io/en/stable/usage.html#rich-comparison-and-logical-operations

src/server/utils/dates.py
```diff
@@ -25,11 +26,11 @@ def week_value_to_week(value: int) -> Week:
     return Week(year=year, week=week)
 
 def guess_time_value_is_day(value: int) -> bool:
-    # YYYYMMDD type and not YYYYMM
+    # YYYYMMDD or YYYY-MM-DD type and not YYYYMM
```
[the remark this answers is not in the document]
rzats: Good point, and changing it to `== 8` doesn't break any tests. I'll make that change

src/server/utils/dates.py
rzats: Looked it up and it's indeed the default behavior, good catch!

melange396:
good stuff, theres just the comment i left in the previous PR about not guessing at the date type when we dont have to: https://github.com/cmu-delphi/delphi-epidata/pull/977/files#r986447003

src/server/utils/dates.py
```diff
@@ -88,83 +89,56 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
...
+        # convert intervals from dates/weeks back to integers
+        ranges = []
+        for m in merged:
+            if m[0] == m[1]:
+                ranges.append(date_to_value(m[0]))
+            else:
+                ranges.append((date_to_value(m[0]), date_to_value(m[1])))
+
+        return ranges
+    except:
```
melange396, line 143: a `logger.info` that reports the reason for this fall-through would be good.  itll almost certainly be a bad conversion in `value_to_date()`, in which case we might wanna see what bad inputs we are being handed.

src/server/utils/dates.py
```diff
@@ -88,83 +89,56 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
 
     # determine whether the list is of days (YYYYMMDD) or weeks (YYYYWW) based on first element
     try:
```
melange396, line 91: can we get rid of this `try/except` now too?

melange396 (changes requested):
quick little things and i think itll be good to go!

src/server/utils/dates.py
dshemetov: The type hints are misleading here then? `values: Sequence[Union[Tuple[int, int], int]]`

src/server/utils/dates.py
```diff
@@ -87,84 +88,55 @@ def dates_to_ranges(values: Optional[Sequence[Union[Tuple[int, int], int]]]) ->
...
+        ranges = []
+        for m in merged:
+            if m[0] == m[1]:
+                ranges.append(date_to_value(m[0]))
+            else:
+                ranges.append((date_to_value(m[0]), date_to_value(m[1])))
+
+        return ranges
+    except:
+        logging.info('bad input to date ranges', detail=values)
```
melange396, line 141: we should add the actual exception and its message to this too

melange396 (approved):

PR #994, opened 2022-10-04
evaluate today only when  is run

melange396:
this change makes it so that when `issue` is not specified in the argument, is determined to be "today" whenever the method is run, and not bound/stuck to whatever "today" was when the file was loaded.  (in the existing pattern, the default value for the argument gets assigned once, at the time the method definition is evaluated.)  in practice, this would probably never come up, but it was bugging me.  i looked for other instances of this pattern but found none: `grep -rl '^\s*def .*(.*(' .`

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -108,7 +108,7 @@ class CsvImporter:
           logger.warning(event='invalid issue directory day', detail=issue_date_value, file=path)
 
   @staticmethod
-  def find_csv_files(scan_dir, issue=(date.today(), epi.Week.fromdate(date.today())), glob=glob):
+  def find_csv_files(scan_dir, issue=None, glob=glob):
     """Recursively search for and yield covidcast-format CSV files.
 
     scan_dir: the directory to scan (recursively)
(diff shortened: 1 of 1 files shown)


dshemetov (approved):
yikes, good catch!

krivard:
Don't we *want* issue to only be assigned once, at the earliest possible time when the acquisition job begins?

krivard:
> that is certainly a valid potential use case. if we want that to be explicit, it would be better to make that behavior more obvious than where it it is now.

It's not a potential use case, it's the _current_ use case, and I think we disagree about what is obvious -- I'm more likely to look at the default value in the function definition than scan through the function body looking for where a None default is re-set.

However, I am okay with this change ("setting the default issue should not be an import side effect") so long as we make it explicit that `csv_to_database` in daily mode will generate only one issue per run. I've opened https://github.com/cmu-delphi/delphi-epidata/pull/996 as a suggested fix but i'm also happy to hear other approaches.

krivard:
I found another reason to do this: [freezegun doesn't work on default arguments](https://github.com/spulec/freezegun#default-arguments) :(

PR #1004, opened 2022-10-13
Add IDE bindings for epidata development

rzats:
**Prerequisites**:

- [x] Build is successful

### Summary

Previously, most IDE features - autocompletion, linting, functioning imports - did not work in epidata due to the codebase's nonstandard structure (pulling code from multiple repos, then reordering them in the Docker container, so the compile-time structure does not match the imports at runtime).

This PR adds a solution that makes IDE development a lot less painful. The idea behind it is to use an "editable install", which is a minimalistic pip package that symlinks to a Python project such that any changes to the package's files are reflected in the dev environment and vice versa. To that end, it

- adds `pyproject.toml` and `setup.cfg` files that define how our local package is laid out

- adds some code to the `install.sh` setup script that symlinks those files to `driver/repos`

- adds instructions on using these to the `README`

**Caveats:**

- `pip install -e . --config-settings editable_mode=strict` needs to be re-ran when any Python files are added or deleted locally. (But edits to existing files will be reflected in the editable install.)
- The bindings in `pyproject.toml` need to be updated whenever we add new packages and subfolders. This is due to the following [limitation](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages):
> Although package_dir applies recursively, you must explicitly list all packages in packages: the Distutils will not recursively scan your source tree looking for any directory with an `__init__.py` file


dev/local/install.sh
```diff
@@ -43,3 +46,6 @@ cd ../../
 
 ln -s repos/delphi/delphi-epidata/dev/local/Makefile
 ln -s repos/delphi/delphi-epidata/dev/local/.dockerignore
+cd repos
+ln -s delphi/delphi-epidata/dev/local/pyproject.toml
+ln -s delphi/delphi-epidata/dev/local/setup.cfg
```
dshemetov, line 51:
```suggestion
ln -s delphi/delphi-epidata/dev/local/setup.cfg
cd -
```

Make the install script back out to the original. Should we install the package here too or is that assuming too much for the user?

dshemetov (changes requested):
one small nit, but otherwise lgtm 🚀 

good work tracking this all the way to the root issue in `setuptools`!

dev/local/install.sh
rzats: re:installing the package for the user - I'd suggest not going for that, since
- we can't really guess what environment they want the package installed into - `python` vs `python3` vs `/path/to/python3` vs `/path/to/my/venv` etc.
- it requires a fairly recent version of setuptools, so we'd have to `install --upgrade` that for the user too

dshemetov (approved):
🚢 🚢 🚢

melange396:
very nice!

PR #1010, opened 2022-10-26
Code improvements to time processing routines

### LINKED ISSUE #999: API server code improvements surrounding some "time" processing routines
Persuant to #998,  theres a lot of needlessly complicated stuff in the api server code...  

This is a section of this, focused mostly around time-related objects and routines

- a `TimePair` is a bit of a misnomer in that the second object of the "pair" is a date type (either 'day' or 'week'), and the other component is a possibly mixed list of pairs of dates (ranges) or individual dates.  "dates" and "TimePair" objects should probably be refactored/combined into one to reduce confusion.
- `guess_time_value_is_*` is a potentially dangerous cruch and should only be used if time type was not specified in the request
- find unused methods like `where_dates` in server code -- linting can probably help us locate these
- where_time_pairs() is the only thing that calls filter_time_pairs() (and theyre nowhere near each other in the code), and filter_time_pairs() is just like filter_dates() which is only called by where_dates() [which itself is never used] and by the twitter and wiki endpoints....  similarities probbly also exist for geos and for srcsigs.  remove/refactor!
- rename inconsistently named methods: `time_value_to_date` <==> `date_to_time_value`, but `week_to_time_value` <==> *** `week_value_to_week` ***
- make an alias for the long typing sequence "`Sequence[Union[Tuple[int, int], int]]`" -- something like "`TimeValues`", add "`TimeValues`" to "`TimePair`"...  then use TimePair wherever possible instead of `S[U[T[i,i,i]]`
- once `TimePair` has replaced similar object instances, pull the `_to_ranges` code into it as member methods
- is this typing alias useful or even necessary?  do we ever do BETWEEN type operations on strings? "`Sequence[Union[Tuple[str, str], str]`"
- there should probably never be a List[] or Sequence[] of GeoPair or TimePair or SourceSignalPair, because they have sequences inside of them, and in our schema/structure, that would require like a cross product of multiple time_types or geo_types or sources

rzats:
Closes #999.

**Prerequisites**:

- [x] Build is successful

### Summary

Refactors several parts of the datetime code as specified in #999. Specifically:

- [x] Use `TimePair` objects as much as possible instead of lists of dates or individual dates
- [x] Only use `guess_time_value_is_*` when absolutely necessary
- [X] Remove unused methods like `where_dates`
- [X] Rename inconsistently named methods: `week_value_to_week`, etc.
- [X] Make an alias for the long typing sequence "Sequence[Union[Tuple[int, int], int]]", use it as much as possible

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ scipy==1.6.2
 tenacity==7.0.0
 newrelic
 epiweeks==2.1.2
-typing-extensions==4.4.0
+typing-extensions
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ scipy==1.6.2
 tenacity==7.0.0
 newrelic
 epiweeks==2.1.2
+typing-extensions==4.4.0
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -174,29 +174,25 @@ def filter_source_signal_pairs(
 def filter_time_pairs(
     type_field: str,
     time_field: str,
-    values: Sequence[TimePair],
+    pair: Optional[TimePair],
     param_key: str,
     params: Dict[str, Any],
 ) -> str:
     """
-    returns the SQL sub query to filter by the given time pairs
+    returns the SQL sub query to filter by the given time pair
     """
-
-    def filter_pair(pair: TimePair, i) -> str:
-        type_param = f"{param_key}_{i}t"
-        params[type_param] = pair.time_type
-        if isinstance(pair.time_values, bool) and pair.time_values:
-            return f"{type_field} = :{type_param}"
-        ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)
-        return f"({type_field} = :{type_param} AND {filter_integers(time_field, ranges, type_param, params)})"
-
-    parts = [filter_pair(p, i) for i, p in enumerate(values)]
-
-    if not parts:
-        # something has to be selected
+    if not pair:
         return "FALSE"
 
-    return f"({' OR '.join(parts)})"
+    type_param = f"{param_key}_0t"
+    params[type_param] = pair.time_type
+    if isinstance(pair.time_values, bool) and pair.time_values:
+        parts =  f"{type_field} = :{type_param}"
+    else:
+        ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)
+        parts = f"({type_field} = :{type_param} AND {filter_integers(time_field, ranges, type_param, params)})"
+
+    return f"({parts})"
 
 
 def parse_row(
@@ -457,7 +453,7 @@ class QueryBuilder:
         self,
         type_field: str,
         value_field: str,
-        values: Sequence[TimePair],
+        values: Optional[TimePair],
         param_key: Optional[str] = None,
     ) -> "QueryBuilder":
         fq_type_field = self._fq_field(type_field)
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -22,7 +22,7 @@ from ._printer import create_printer, APrinter
 from ._exceptions import DatabaseErrorException
 from ._validate import DateRange, extract_strings
 from ._params import GeoPair, SourceSignalPair, TimePair
-from .utils import time_values_to_ranges, days_to_ranges, weeks_to_ranges
+from .utils import time_values_to_ranges, days_to_ranges, weeks_to_ranges, TimeValues
 
 
 def date_string(value: int) -> str:
@@ -86,7 +86,7 @@ def filter_integers(
 
 def filter_dates(
     field: str,
-    values: Optional[Sequence[Union[Tuple[int, int], int]]],
+    values: Optional[TimeValues],
     param_key: str,
     params: Dict[str, Any],
 ):
@@ -411,16 +411,6 @@ class QueryBuilder:
     ) -> "QueryBuilder":
         fq_field = self._fq_field(field)
         self.conditions.append(filter_integers(fq_field, values, param_key or field, self.params))
-        return self
-
-    def where_dates(
-        self,
-        field: str,
-        values: Optional[Sequence[Union[Tuple[int, int], int]]],
-        param_key: Optional[str] = None,
-    ) -> "QueryBuilder":
-        fq_field = self._fq_field(field)
-        self.conditions.append(filter_dates(fq_field, values, param_key or field, self.params))
         return self
 
     def where_geo_pairs(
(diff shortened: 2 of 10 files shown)


melange396:
This is teriffic!  Everything looks great, especially the changes in `src/server/endpoints/covidcast.py` which make me happy :)

There might be a couple more changes w/ the `TimeValues` alias that could be made:
  - i think you can do a replacement to `TimeValues` at `src/server/_query.py:191`, or maybe even entirely remove the call to `cast()`
  - a couple `List[Union[Tuple[int, int], int]]`s might be able to come out too:
    - can the `issues` argument to `src/server/endpoints/covidcast.py:_handle_lag_issues_as_of()` be a `TimeValues`?
    - can `values` in `src/server/_validate.py:extract_dates()` be hinted to be type `TimeValues`?  i think you can also remove `DateRange` `from src/server/_validate.py:144` and then change `src/server/_validate.py:extract_dates()` to return `Optional[TimeValues]`

Then there are a few remaining bullets in https://github.com/cmu-delphi/delphi-epidata/issues/999...
  - Can you dig around and see if we really need the instances of `Sequence[Union[Tuple[str, str], str]` or can we just use `Sequence[str]` in its place instead?
  - There is also the discussion of the lists/sequences of `TimePair`s -- if changing that stuff out looks like a big job, we could move it to a different PR.

melange396 (commented):
whoops, i had these comments pending and forgot to send them!

requirements.txt
```diff
@@ -11,3 +11,4 @@ scipy==1.6.2
 tenacity==7.0.0
 newrelic
 epiweeks==2.1.2
+typing-extensions==4.4.0
```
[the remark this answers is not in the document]
rzats: Hm, isn't best practice to pin the version in case newer updates unexpectedly break?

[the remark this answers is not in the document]
melange396: for the most part, yeah, but this is a fairly fundamental and python-authored package that i expect to "do the right thing" as it matures.  pinning can also have negative consequences where you can get stuck with vulns or bugs or poor performance implementations.  so... im not actually certain on what the best thing to do here, hence the "maybe" ¯\_(ツ)_/¯

rzats:
@melange396 made a few more changes to the PR:
- All of your changes to `TimeValues` can be implemented, and `parse_day_or_week_arg` can be converted into `TimePair` from a tuple too :)
- `Tuple[str, str]` can be safely removed; I've taken a look at all of the calls to `where_strings`/`filter_strings` and those only deal with `Sequence[str]`. (Also, SQL's `BETWEEN` behaves a bit oddly with strings - `BETWEEN 'A' AND 'B'` will include all the strings that start with `A` but exclude the ones the start with `B`, which feels counterintuitive.)
- As for the sequences of `TimePair`, see https://github.com/cmu-delphi/delphi-epidata/issues/999#issuecomment-1302021149. tl;dr: It's not too hard to implement, but some tests suggest that `List[TimePair]` should be supported.

src/server/_query.py
```diff
@@ -174,29 +174,25 @@ def filter_pair(pair: SourceSignalPair, i) -> str:
 def filter_time_pairs(
```
dshemetov, line 174:
```suggestion
def filter_time_pair(
```

and then propagate this renaming out to `where_time_pairs` -> `where_time_pair` and update wherever that's called?

src/server/_query.py
```diff
@@ -174,29 +174,25 @@ def filter_pair(pair: SourceSignalPair, i) -> str:
...
+    pair: Optional[TimePair],
     param_key: str,
     params: Dict[str, Any],
 ) -> str:
     """
-    returns the SQL sub query to filter by the given time pairs
+    returns the SQL sub query to filter by the given time pair
     """
+    if not pair:
+        return "FALSE"
```
dshemetov, line 185: Question: is this a good code path to have? AFAICT there's an exception raised in `covidcast.py:parse_time_pairs` if you don't have either: a) `time=...` or b) `time_type=...` and `time_values=...`. So this will technically never happen in the API. So:
- Pro: function safety, can be reused in the future in other places that didn't have the same validation
- Con: confusing now, because it makes you think this is a possible code path

not super blocking, i can go either way, but curious what everyone thinks here, cc @melange396 @krivard

tests/server/test_params.py
```diff
@@ -267,35 +267,38 @@ def test_parse_day_value(self):
...
+            with app.test_request_context("/?time=week:*;week:202012"):
                 self.assertEqual(
                     parse_time_arg(),
-                    [
-                        TimePair("day", True),
-                        TimePair("day", [20202012]),
-                        TimePair("week", [(202101, 202104)]),
-                    ],
+                    TimePair("week", True)
+                )
```
dshemetov, line 292: Try testing the mixed time type exception here:

```py
        with self.subTest("hybrid"):
            with app.test_request_context("/?time=day:*;day:20202012;week:202101-202104"):
                with pytest.raises(ValidationFailedException(match=r"*mixes \"day\" and \"week\" time types.")):
                    parse_time_arg()
```

dshemetov (changes requested):
looking good, modulo a few small details

src/server/_query.py
rzats: Like you've mentioned, I'd rather have the extra failsafe for function safety so I kept this for now (but it should be trivial to remove if needed)

tests/server/test_params.py
rzats: (Did this with `assertRaisesRegex` which seems identical!)

rzats:
@dshemetov I've addressed your comments! (except for the `if not pair, return false` code path for now)

dshemetov (dismissed):
lgtm

melange396 (dismissed):
This is really great work, thank you for doing it.  Do you feel like youve reduced the total complexity of the api server code?  I feel like you did, and it makes me cringe at the api code less than before.  :)

This all looks quite good to me, but i have a couple optional things you could look at; if you are getting burned out on this PR and wanna move on to something else, i can just turn them into new issues:
  - We've made a bunch of changes to the type hinting stuff -- is all that still happy?  Is your IDE set up to show mismatches and whatnot?  I tried to run a type checker from the command line (with some limited success) and among other things i saw, it suggests it might help to change the type alias for TimePair to use List[...] instead of Sequence[...]
  - Move the `*_to_ranges()` methods (and similar) to TimePair instance methods, so you can have callers be agnostic to week/day/whatever, then (for example) in `_query.py:filter_time_pair()`, instead of `ranges = weeks_to_ranges(pair.time_values) if pair.is_week else days_to_ranges(pair.time_values)` you can just do `ranges = pair.to_ranges()` and `to_ranges()` does the appropriate thing for the time type of the TimePair.
  - Rename TimePair (and also various instances of time_pair variable names or similar) to something more appropriate.  Youve done the hard work to make it better, you should get to choose a new class name.

Im gonna mark this as approved, but you should also add the comment in `_params.py`, on the point Dmitry brought up.

melange396 (approved):

PR #1014, opened 2022-10-31
[JIT Variant] Non-streaming Pandas approach

dshemetov:
A version of #646 without streaming.

**Prerequisites**:

- [x] Build is successful

### Summary

The Python iterator-based approach to JIT has proven to be 3-5x slower than reading those signals from the database. It is well-known that: a) Python for-loops are slow and b) nested iterators incur a stack overhead, which slows computations down. This PR attempts to get around these by using Pandas to do time series computations on a per-signal basis:
- Pandas dataframe computations call fast C code instead of Python for-loops
- loading the data into a dataframe incurs a memory cost, but removes the nested iterator stack overhead


dshemetov:
Docker image tags can't have forward slashes in them, so use the tag name `jit-pandas` for this one.

integrations/server/test_covidcast.py
```diff
@@ -138,7 +138,7 @@ def test_csv_format(self):
       **{'format':'csv'}
     )
 
-    # TODO: This is a mess because of api.php.
+    # TODO: This is a mess because of api.php. Or maybe it's just a mess.
```
melange396, line 141: XD

dshemetov: 🤣 courtesy of Github Copilot

Issue #1007, opened 2022-10-20
Clean covid_hosp_facility hospital_subtype data


krivard:
I was looking for a complete list of all hospital subtypes for the documentation update, and found some alarming miscoded data:

```
mysql>  select distinct hospital_subtype from covid_hosp_facility;
[...]
| 34011F                                                           |
| 457736                                                           |
| 05T782                                                           |
| 405025                                                           |
| 676622                                                           |
| 8dcb4c99-fb3e-4c99-95d7-bdcf1253268d                             |
| 230118                                                           |
| 9f7c7842-12dc-4d6e-b4f6-a864ed471d99                             |
| 02fcc4218077f123aea96b1ccee7688249ed732178bd06ef73d8b227533bc19a |
| 140250                                                           |
| 17A020                                                           |
| 320016                                                           |
| 362035                                                           |
| 36T059                                                           |
| 555143                                                           |
| 677297                                                           |
| dead8933dab2783b14e9cb678ac9c55a70cb9ff5d69cf3dc4cce866ab52e7d8a |
| 040155                                                           |
| 140307                                                           |
| 1be1e598889c91e1d8798795da6e7883cf2b2e8acb4311563c474c54c7322bb0 |
| 333562                                                           |
| 390130                                                           |
| 39T142                                                           |
| 443459                                                           |
| 490001                                                           |
| 8148f640150c54b1ee6c0d275e2f907d61782176d6ee21fbce88fd076477d211 |
| e9947863f76de4aa0145ae11b279480dae1a31a303177d19b6d7bcd2c0ebac18 |
| b2937a1082df4731d7489c512c9155927bb390a498502997939ac4dd8308ffc0 |
| 110252                                                           |
| 370243                                                           |
| e1a5f1ded22a21b1cf6d9d5ff579edbecd8492f810dfbb6108a545f59e758c6c |
+------------------------------------------------------------------+
5109 rows in set (5 min 20.12 sec)
```

This should have been something like Childrens Hospitals, Critical Access Hospitals, Long Term, Psychiatric, Rehabilitation, Short Term; the actual values look more like `hospital_pk`s (CCNs where the hospital has one, a giant hash if they don't).

We should
- [ ] figure out how many rows are affected
- [ ] figure out if our current code is doing this
- [ ] fix the code
- [ ] fix the data

krivard:
Starting id: 12004206
Ending id: 19561068
Affected rows: 146867
Currently active: No

[...]

PR #1016, opened 2022-10-31
covid_hosp_facility lookup speedup

melange396:
still todo:
  - test db migration routine(s) a bit more, and plan for execution on prod (aka plan rollout)
    - choice of migration doesnt matter too much because values will be updated within a week (during the next data acquisition)
  - add test that verifies any updates to facility metadata (like a change in street address) will be persisted and replace previous metadata


src/ddl/migrations/covid_hosp_facility_v0.3-v0.3.1.sql
```diff
@@ -0,0 +1,22 @@
+INSERT INTO `covid_hosp_facility_key`
```
krivard, line 1: This line is likely to fail, since when this file is run, the `covid_hosp_facility_key` table won't exist yet

src/ddl/migrations/covid_hosp_facility_v0.3-v0.3.1.sql
```diff
@@ -0,0 +1,22 @@
...
+-- FROM `covid_hosp_facility` ORDER BY `id` ASC
+-- ON DUPLICATE KEY UPDATE `address`=`address`,`ccn`=`ccn`,`city`=`city`,`fips_code`=`fips_code`,`geocoded_hospital_address`=`geocoded_hospital_address`,`hhs_ids`=`hhs_ids`,`hospital_name`=`hospital_name`,`hospital_subtype`=`hospital_subtype`,`is_metro_micro`=`is_metro_micro`,`state`=`state`,`zip`=`zip`;
+
+-- -- or a query similar to the previous behavior of the covid_hosp_facility endpoint:
+-- INSERT INTO `covid_hosp_facility_key`
+--   (`hospital_pk`,`address`,`ccn`,`city`,`fips_code`,`geocoded_hospital_address`,`hhs_ids`,`hospital_name`,`hospital_subtype`,`is_metro_micro`,`state`,`zip`)
+-- SELECT
+--    `hospital_pk`,MAX(`address`),MAX(`ccn`),MAX(`city`),MAX(`fips_code`),MAX(`geocoded_hospital_address`),MAX(`hhs_ids`),MAX(`hospital_name`),MAX(`hospital_subtype`),MAX(`is_metro_micro`),MAX(`state`),MAX(`zip`)
+-- FROM `covid_hosp_facility`
+-- GROUP BY `hospital_pk`;
```
krivard, line 22: All of the above methods assume we will fix https://github.com/cmu-delphi/delphi-epidata/issues/1007 before migration -- is that strictly necessary, or could we effect a migration by creating the table and then re-importing the most recent facility update?

src/server/endpoints/covid_hosp_facility_lookup.py
```diff
@@ -17,27 +17,25 @@ def handle():
     fips_code = extract_strings("fips_code")
 
     # build query
-    q = QueryBuilder("covid_hosp_facility", "c")
+    q = QueryBuilder("covid_hosp_facility_key", "c")
     q.fields = ", ".join(
-        [
+        [ # NOTE: fields `geocoded_hospital_address` and `hhs_ids` are available but not being provided by this endpoint.
```
krivard, line 22: 💯

src/acquisition/covid_hosp/common/database.py
```diff
@@ -172,6 +172,33 @@ def insert_dataset(self, publication_date, dataframe):
...
+    if hasattr(self, 'AGGREGATE_KEY_COLS'):
+      ak_cols = self.AGGREGATE_KEY_COLS
+
+      # restrict data to just the key columns and remove duplicate rows
+      ak_data = dataframe[ak_cols].drop_duplicates()
+      # cast types (bools and NULLables)
+      for col in ak_cols:
+        if col.startswith('is_'): # TODO: this is hacky af
+          ak_data[col] = (ak_data[col] == 'true')
+      ak_data = ak_data.to_numpy(na_value=None).tolist()
```
krivard, line 184: maybe less hacky -- we already know the python dtype of each column, including the aggregate key columns
```suggestion
      dataframe_typemap = {
        name: dtype
        for name, _, dtype in dataframe_columns_and_types
      }
      # cast types (bools and NULLables)
      for col in ak_cols:
        ak_data[col] = dataframe_typemap[col](ak_data[col])
```

src/acquisition/covid_hosp/common/database.py
```diff
@@ -172,6 +172,33 @@ def insert_dataset(self, publication_date, dataframe):
...
+
+      # create string of tick-quoted and comma-seperated column list
+      ak_cols_str = ','.join([ '`'+col+'`' for col in ak_cols ])
+      # ...and ticked and comma-sep'd "column=column" list for ON UPDATE (to keep only the most recent values for each pk)
+      ak_updates_str = ','.join([ '`'+col+'`=`'+col+'`' for col in ak_cols ])
```
krivard, line 189: tracking lots of nested delimiters is hard to read -- maybe this instead?

```suggestion
      ak_cols_str = ','.join(f'`{col}`' for col in ak_cols)
      # ...and ticked and comma-sep'd "column=column" list for ON UPDATE (to keep only the most recent values for each pk)
      ak_updates_str = ','.join(f'`{col}`=`{col}`' for col in ak_cols)
```

src/ddl/migrations/covid_hosp_facility_v0.3-v0.3.1.sql
melange396: good catch, ill add the new table definition to this

src/ddl/migrations/covid_hosp_facility_v0.3-v0.3.1.sql
melange396: we can fix before migration, or we can TRUNCATE and re-run this INSERT/SELECT as part of the fix (after the other table has been rectified).  we could also remove bad rows from this table as part of the fix, but that requires identifying them which may or may not be easy depending on how the fix is implemented.

src/ddl/migrations/covid_hosp_facility_v0.3-v0.3.1.sql
melange396: done

src/acquisition/covid_hosp/common/database.py
```diff
@@ -172,6 +172,33 @@ def insert_dataset(self, publication_date, dataframe):
...
+      # ...and string of VALUES placeholders
+      values_str = ','.join( ['%s'] * len(ak_cols) )
+      # use aggregate key table alias
+      ak_table = self.table_name + '_key'
+      # assemble full SQL statement
+      ak_insert_sql = f'INSERT INTO `{ak_table}` ({ak_cols_str}) VALUES ({values_str}) ON DUPLICATE KEY UPDATE {ak_updates_str}'
+
+      # commit the data
+      with self.new_cursor() as cur:
+        cur.executemany(ak_insert_sql, ak_data)
```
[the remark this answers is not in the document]
melange396: it could go either way...  its not specific to the facility data, but it is true that the facility schema is the only one we currently have that needs it.  if we leave it here, some future data set from covid_hosp that has composite keys can use it without modification.

src/acquisition/covid_hosp/common/database.py
melange396: lol, maybe i should have left my "hacky" comment out of this...  i did it that way because i didnt want to have to resort to using the `dataframe_columns_and_types` when the boolean column was the only one that needed massaging.

your suggestion is pretty good, but it needs an `apply()` and will probably have a few other hoops to jump through for NaN handling.  ill do that and add another commit.

melange396: less hacky alternative has been pushed

src/acquisition/covid_hosp/common/database.py
```diff
@@ -172,21 +172,31 @@ def insert_dataset(self, publication_date, dataframe):
...
+        for name, _, dtype in dataframe_columns_and_types
+      }
       for col in ak_cols:
-        if col.startswith('is_'): # TODO: this is hacky af
-          ak_data[col] = (ak_data[col] == 'true')
+          def cast_but_sidestep_nans(i):
+            # not the prettiest, but it works to avoid the NaN values that show up in many columns
+            if isinstance(i, float) and math.isnan(i):
+              return None
+            return dataframe_typemap[col](i)
```
krivard, line 191: We also do that here:
https://github.com/cmu-delphi/delphi-epidata/blob/50e40ba964d19f50563f58ddc54e758af768a068/src/acquisition/covid_hosp/common/database.py#L166-L169

should this code get pulled into a function that can be used in both places?

krivard: also/alternatively: this is a touch on the YAGNI side, since I can't imagine a scenario where a float would become part of an aggregate key. maybe the check is unnecessary?

krivard (approved):
Looks good; optional refactor

src/acquisition/covid_hosp/common/database.py
melange396: I considered treading down the slippery slope of refactoring/repairing this (and other) code in this method and module, and ultimately decided it was better to keep my changes separate and distinct.  I can do something to remove the duplication you pointed out, but it felt like touching any of that code would be like opening a can of worms...  My previous "hacky af" handling that only really called out the boolean column(s) was an attempt to produce a smaller and easily understandable diff.

Unfortunately, if we are going to apply the types as defined [here](https://github.com/cmu-delphi/delphi-epidata/blob/ccad94a1ba1b16fb246c4a16e03e5c8c8dac3d35/src/acquisition/covid_hosp/facility/database.py#L18), this check is needed in some form or another.  The dataframe passed as an argument to this method has a NaN wherever the input CSV has an empty field, no matter which column that field is from.  If we apply `str()` (for instance) from `dataframe_typemap` to such a NaN, we get the string `'nan'` when we should really have `None` or even just an empty string.  I almost did something like:
```
for col in ak_cols:
    ak_data[col] = list(map(dataframe_typemap[col], ak_data[col].to_numpy(na_value=None)))
```
but that still needs a guard to prevent (for instance) the string `'None'` from appearing in the database where we now have a `NULL`.  Instead of inventing a new and somewhat overly complicated process, i cribbed from what already existed.

I dont love where this ended up, but there were downsides to each of the options: like the "hacky af" thing that i had before, what we have here in the most recent commit, the incomplete thing i just described in the snippet above, or (as you suggested) altering the existing code to factor out the duplicated functionality.  Maybe i should have just done the latter, as it wouldve taken me a lot less time than typing all this up (lolz).

melange396:
when do we want to deploy this?  running the migration by hand is easy and will not break anything if done before deployment, but if another data acquisition is performed between migration and deployment, the new secondary table may* end up out-of-sync.  depending on the content of the data in such an acquisition, it may not be out-of-sync by much, and in fact, i think it is likely that it will not be out-of-sync at all.  (related to [this comment](https://github.com/cmu-delphi/delphi-epidata/pull/1016/files#r1010539857))

src/acquisition/covid_hosp/common/database.py
krivard: so a string column with one blank row in the csv gets converted into a mixed-type column in the pandas df where some rows are strings and some rows are floats?

melange396: sorta, except theyre all mixed-type columns (more precisely, `dtype: object`) between when the csv is read and here 😬

krivard (approved):
👍

PR #1021, opened 2022-11-02
Convert covidcast signal OR clauses to UNION

LeonLu2:
closes #763

**Prerequisites**:

- [X] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [X] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [X] Build is successful
- [X] Code is cleaned up and formatted

### Summary

1. Refactored the QueryBuilder to convert covidcast signal OR clauses to UNION
2. Edited test_query.py accordingly to test filters changed in server code

PR #1026, opened 2022-11-09
[JIT Variant] Split multi-signal queries into many single-signal queries

dshemetov:
A version of #646 

**Prerequisites**:

- [x] Build is successful

### Summary

Multi-signal queries complicate the JIT work by introducing a lot of iterator overhead. This avoids that by splitting them into many single-signal queries. A few notes:
- this branch makes non-derived signals completely avoid JIT code path, so its speed on non-derived signals should be identical to `dev`
- derived signals that rely on the same base signal make the same SQL query twice, so there's room for optimization here; I'm hoping that database caching makes that extra complexity not worth it
- just like #1014, the Docker image name is `jit-multi-sql` instead of the branch name because `/` is not allowed
- has some conflicts with #1021 that we'll need to address later


dshemetov:
Going with the much cleaner approach in #1049.

PR #1029, opened 2022-11-16
Add live debugging targets to local dev Makefile

krivard:
**Prerequisites**:

- [x] Build is successful

### Summary

Adds two targets useful for debugging:
* `mysql` - a live mysql prompt to the dev database
* `bash` - a live bash prompt in the python environment


melange396 (approved):
this is great!

i ~never use the `docker run` command with "bash" but often with "python3" in its place instead...  but "bash" is good because lets you poke around the file system of the container, and you can easily enter the same interactive python interpreter by just running `python3` from within that bash shell session.

Issue #962, opened 2022-08-26
Design and enforce linting standards


krivard:
https://github.com/cmu-delphi/covidcast-indicators has nice linting requirements, and we deserve nice things over here too. 

To do it the same way that `indicators` does, we'd need to either figure out how to lint a mixed tree of python files or convert `acquisition` and `server` to actual legit python modules. Then we just add pylint commands to CI.

We don't have to do it the same way that `indicators` does though. Alternatives?

dshemetov:
Looking at the [`pylint` docs](https://pylint.pycqa.org/en/latest/user_guide/usage/run.html), it looks like they really leave on your own when your files aren't packaged. [This somewhat hacky](https://stackoverflow.com/a/63044665/4784655) solution might just get the job done though. Can't really think of serious downsides.

dshemetov:
I'll also throw out that I'm pretty partial to [`black`](https://black.readthedocs.io/en/stable/index.html). 

The biggest selling point for me is the time-saving: often the solution to `pylint`'s complaints are deterministic (delete trailing white spaces, break long lines, etc.) and `black` can just do that automatically.

To be fair, sometimes it makes things look worse. You can see instances of it in [Katie and I's conversations here](https://github.com/cmu-delphi/covidcast-indicators/pull/1577). But I think many of those can be mitigated by a judicious use of the [`fmt` flag](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=fmt#code-style).

melange396:
for posterity, the above-referenced mention is here: https://github.com/cmu-delphi/delphi-epidata/pull/1041#issuecomment-1335372234

dshemetov:
Not linting, but formatting work done here #1189

PR #1041, opened 2022-12-02
API server code health improvements

### LINKED ISSUE #998: API server code health improvements
The covidcast API server code works well, but it can be improved with an eye toward extensibility and other future changes.  A number of points of potential refinement exist, including:

- Long and/or complex ternaries and comprehensions are pervasive and should be reduced and/or simplified -- code golf is not good for readability/maintainability
- Many class, method, and variable names are not descriptive or are misleading, and should be clarified
- Documentation in the code for explaining algorithms and implementation choices is very thin and should be expanded.  Undocumented code understandability is further exacerbated by naming choices.
- Type annotation/hinting is clunky and verbose, and often not necessary.

rzats:
Addresses #998.

**Prerequisites**:

- [x] Build is successful

### Summary

General linting pass for the API & acquisition code: removes unused imports, fixes some whitespace issues, fixing exception try/except blocks etc


src/acquisition/nidss/taiwan_nidss.py
```diff
@@ -204,11 +203,6 @@ def get_dengue_data(first_week, last_week):
       fields = line.split(',')
       location_b64 = base64.b64encode(fields[3].encode('utf-8'))
       location = NIDSS._TRANSLATED[location_b64]
-      region = NIDSS.LOCATION_TO_REGION[location]
-      imported_b64 = base64.b64encode(fields[6].encode('utf-8'))
-      imported = imported_b64 == b'5piv'
-      sex = fields[5]
-      age = fields[7]
```
melange396, line 211: maybe just comment out the `region`, `sex`, and `age` lines and note that theyre currently unused?  leaving them in could help someone debug later (at least in terms of parsing the input file), and they might even be data points we will add in the future...

src/server/endpoints/covidcast_utils/model.py
```diff
@@ -272,7 +272,7 @@ def create_source_signal_alias_mapper(source_signals: List[SourceSignalPair]) ->
             continue
         # uses an alias
         alias_to_data_sources.setdefault(source.db_source, []).append(source)
-        if pair.signal == True:
+        if pair.signal:
```
melange396, line 275:
```suggestion
        if pair.signal is True:
```

im pretty sure you wanna have this test for `True`, because that is the `signal=*` (wildcard) indicator.  the alternative for `pair.signal` if it is not `True` would be a sequence of strings (a list of signals), which is "truthy" (as long as its not empty) and so it would fall into this block of the conditional.  the string sequence should go into the `else` block to be used as-is, instead of enumerating all the signals for that source.

melange396 (changes requested):
oof, some of the code here was really gnarly (especially the first file, `src/acquisition/afhsb/afhsb_csv.py`)!  i looooove seeing all the unused imports and variables getting removed!  the whitespace fixes also soothe my ocd.  this is fabulous!!!

what programs/packages/commands/parameters did you use to do this?  its probably something that should be incorporated into #962...

you should change "Closes #998" to "Addresses #998" and we can get this PR merged by itself.  itll be good to separate this set of problematic code changes that a linter found automatically (syntactic points) from the other semantic or logical changes that youll make later in further pursuit of issue #998.

src/server/endpoints/covidcast_utils/model.py
rzats: Good catch, applied!

melange396 (approved):
looks great to me!

rzats:
@melange396 I applied the changes you've requested!

> what programs/packages/commands/parameters did you use to do this? its probably something that should be incorporated into https://github.com/cmu-delphi/delphi-epidata/issues/962...

This PR fixes issues detected by [Prospector](https://prospector.landscape.io/en/master/), which uses [pylint](https://pylint.pycqa.org/en/latest/index.html) internally.

There are tools that can autoformat Python repos, which we can use now that we have a working editable package. [black](https://github.com/psf/black) is one that comes to mind: it can be incorporated as a CI step (yells at you if your code isn't formatted), post-commit hook (formats your code automatically), etc.

**Caveat:** these tools are non-configurable and "uncompromising" by design, so on an initial run, they will reformat the majority of the codebase to the indentation style they like, affecting most lines of code. This will
- break `git blame` for a while
- cause a massive merge conflict for pending PRs (like https://github.com/cmu-delphi/delphi-epidata/pull/1011 or https://github.com/cmu-delphi/delphi-epidata/pull/995)

so we need to think carefully about when to apply them.

Issue #965, opened 2022-08-30
Dependence on `operations` repo


dshemetov:
On the heels of some recent discussions about the tight dependency between this repo and the `operations` repo, I did some searching to see if the base Docker images, whose Dockerfiles live in the `operations` repo, are used anywhere else.

Here are the repos that use them (this is an [organization-wide codebase search query](https://github.com/search?p=1&q=org%3Acmu-delphi+Dockerfile&type=Code)):
- [`delphi-epidata-private`](https://github.com/cmu-delphi/delphi-epidata-private/blob/main/docs/build-docker-images.sh)
- `delphi-epidata`
- [`www-epicast`](https://github.com/cmu-delphi/www-epicast/blob/main/docs/epicast_development.md)
- [`flu-contest`](https://github.com/cmu-delphi/flu-contest/blob/main/src/covid/README.md)

I'm not familiar with those repos, so I'm not sure if they're still used.
- If they are used, then there's still the possibility of just merging the Dockerfiles together in this repo, and keeping operations as a dependency for those.
- If they're not, then it makes even more sense to just merge these Dockerfiles together.

Removing dependency on the `operations` repo would:
- simplify dev setup for this repo
- make dependency updates easier (see for example, the need to do simultaneous PRs in both repos cmu-delphi/operations#70 and #935, in order to update dependencies)

Since this repo already has [devtools](https://github.com/cmu-delphi/delphi-epidata/tree/dev/dev/local), which makes dev setup easier, I would say this isn't a very high priority refactor, but something of a possibility.

melange396:
i looked at a bunch of the `undefx` dependencies while i was doing #1000.  i think you can find virtually all of it with this command: `grep -r "undef-analysis\|undefx\|py3test" .`

- in addition to the removal from `ci.yaml` in #935, you can also potentially remove references to the undefx repos or their location on the local filessystem from these files in this repo:
  - `dev/local/install.sh`
  - `docs/epidata_development.md`
  - `docs/new_endpoint_tutorial.md`

- then in `deploy.json`, remove the line that matches this as well as the comment line above it:
  - `{"type": "py3test"}`

- and finally under `tests/` and `integrations/` , remove lines from the many files that match these two:
  - `# py3tester coverage target`
  - `__test_target__`

- a number of our other repositories also have traces of this stuff:
[...]

PR #1043, opened 2022-12-05
Merge `delphi_web_python` Docker image from `operations` repo

dshemetov:
This is part of #965.

Having our Dockerfiles be partially compiled from the `operations` repo and partially here introduces unnecessary complexity, with no clear benefits (that I can see).

In the JIT work, I need to bump the pandas version, so moving the Python `requirements.txt` file from the `operations` repo here lets me do that in just one place.

I believe this is the last of our three Docker images to depend on the operations repo:
- `delphi_database_epidata` is already fully built from `dev/docker/database/epidata/Dockerfile` in this repo
- `delphi_web_epidata` is also fully built from `devops/Dockerfile` in this repo

This doesn't fully decouple us from `operations` as the `setup.sh` script still copies many old repos to the `delphi_web_python` image. We can consider excising those in a different PR.

**Prerequisites**:

- [x] Build is successful

### Summary

This PR:
- merged the [Dockerfile for the `delphi_python` image](https://github.com/cmu-delphi/operations/tree/main/dev/docker/python) into the Dockerfile for the `delphi_web_python` image in this repo
- copied the Python `requirements.txt` file from [that repo](https://github.com/cmu-delphi/operations/tree/main/dev/docker/python/assets) and merged it in with `requirements.dev.txt` 
- copied `setup.sh` file from [that repo](https://github.com/cmu-delphi/operations/tree/main/dev/docker/python/assets) to `dev/docker/python` in this repo
- renamed `requirements.txt` to `requirements.api.txt` and sorted dependencies
- deduplicated packages between `requirements.api.txt` and `requirements.dev.txt` and removed some unused ones (like `scikit-learn` from `requirements.dev.txt`)
- pinned package versions in `requirements.dev.txt` with the latest working versions


melange396:
nice!  you can then remove these lines as well: https://github.com/cmu-delphi/delphi-epidata/blob/3d37c593ce460ec0bce49f28eafa2284935d0c94/dev/local/Makefile#L124-L125

melange396:
and looks like we couldve/shouldve removed these lines already too:
https://github.com/cmu-delphi/delphi-epidata/blob/3d37c593ce460ec0bce49f28eafa2284935d0c94/dev/local/Makefile#L101-102

dev/docker/python/Dockerfile
```diff
@@ -1,4 +1,22 @@
 # start with the `delphi_python` image
```
melange396, line 1:
```suggestion
```

melange396:
with this PR, we have 3 different "requirements.txt" files in this repo: (ignoring `./src/server/covidcast_issues_migration/requirements.txt` from an old one-off thing)
- `./dev/docker/python/requirements.txt`  (which you just added/copied here)
- `./requirements.txt`
- `./requirements.dev.txt`

all 3 are used in `dev/docker/python/Dockerfile` to create `delphi_web_python`, which is only really used for tests.

only `./requirements.txt` (in this repo's root dir) is used to create `delphi_web_epidata` which runs the api server.

i have a PR (#1046) against this branch.  my changes sort the 3 requirements files, and remove duplicated entries from the new one.  we should prune down that file some more if we can, maybe even remove it entirely by moving its entries to the other two files.

a caveat is that `.github/workflows/update_gdocs_data.yml` also uses `requirements.dev.txt`, and im not sure what it does -- the meat of it seems to be in its line 26 "run: inv update-gdoc" but i cant figure out what that does

dshemetov:
`update_gdocs_data.yml` updates the local `db_sources.csv` and `db_signals.csv` files (that the server uses to understand what signals are available) from [this Google sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vRfXo-qePhrYGAoZqewVnS1kt9tfnUTLgtkV7a-1q7yg4FoZk0NNGuB1H6k10ah1Xz5B8l1S1RB17N6/pub?gid=329338228) (and another one elsewhere).

dshemetov:
Merged in George's mods from #1046.

melange396:
looks pretty good to me:
- removes a cross-repo dependency
- should speed up our build process (including ci)
- generally reduces complexity

my concerns that remain:
- looks like `xlrd` and `matplotlib` can come out of the requirements file(s)?
- do we wanna add `scipy` back into the one requirements file?  our one reference to it is probably being taken care of by a chained dependency, so its not a huge deal
- im not sure we need or want to pin so many of those version numbers, but that can stand for the moment as i am soon going to create a github issue about version pinning
- we may want to "mark for deletion" a number of the things in the `operations` repo that we dont need anymore (`Dockerfile`, `setup.sh`, `requirements.txt`, etc...)

melange396 (dismissed):

melange396 (approved):

PR #1058, opened 2022-12-19
[WIP] API server code health pass

### LINKED ISSUE #998: API server code health improvements

rzats:
Closes #998.

**Prerequisites**:

- [x] Build is successful

### Summary

Fixes a set of issues in the server code as seen [here](https://github.com/cmu-delphi/delphi-epidata/issues/998#issuecomment-1348732544).

- [X] fix up `_handle_lag_issues_as_of()`
- [x] rework `set_order()`
- [X] remove `get_related_signals()`
- [x] remove `_db.py`, move configuration settings which are actually used elsewhere
- [x] rename "pair" (time, geo, sourcesignal) classes
- [x] move `parse` & `extract` to proper location
- [ ] remove global `flask.request` references
- [ ] reword methods like `parse` & `extract` to match their actual purpose
- [ ] integrate a number of class definitions


src/server/endpoints/covidcast.py
```diff
@@ -132,22 +92,29 @@ def handle():
...
 
-    _handle_lag_issues_as_of(q, issues, lag, as_of)
+    # Handle history table in this priority order: issues > lag > as_of
+    # if all are None, we are using the (standard/default) `latest_table`, to fetch the most recent issue quickly
+    if issues:
+        _apply_issues_filter(q, issues)
+    elif lag is not None:
+        _apply_lag_filter(q, lag)
+    elif as_of is not None:
+        _apply_as_of_filter(q, as_of)
```
melange396, line 117: it seems silly to check `if field` / `if field is not None` here when each of these methods do the same test internally.  it also seems silly to silently leave out a filtering condition for `lag` when `issue` is also specified, or to leave out `as_of` when `lag` or `issue` is specified...  i claim that either (1) these 3 conditions should all be applied to the resulting SQL if they are specified in any combination, or (2) an error should be produced if they are truly intended to be mutually exclusive.  i think it is more correct to do the former; that is, leave out the `if/elif/elif` lines and just call the 3 methods here.  thoughts?

melange396 (changes requested):
Instead of letting this snowball into one huge megalithic PR (encompassing many different types of changes) that auto-closes the referenced issue, it might be better to create smaller targeted PRs (that are more easily digested individually) and then ultimately close the PR by hand.  It'll simplify the overall review process by separating concerns.

I don't think "XyzFilter" (TimeFilter, GeoFilter, etc) is a great option for renaming...  Its not really a "filter" until it is applied, and there isnt anything about these objects (that specify sets of times/locations/signalnames) that is intrinsically filter-like.  How about "TimeSpec" instead (or i could even be convinced to use "TimeSet")?  I apologize, i should have been more explicit in my issue comment on the "TimePair" name being bad, and suggested these options then.

rzats:
@melange396 thanks for the comments! I'll close this off for now and split it into a few separate PRs.

src/server/endpoints/covidcast.py
rzats: Applying all three makes sense if it won't break any existing workflows!

PR #1062, opened 2022-12-30
API server code health pass - rework parse* & extract* methods

rzats:
Partially addresses #998.

**Prerequisites**:

- [X] Build is successful

### Summary

Fixes a set of issues in the server code as described [here](https://github.com/cmu-delphi/delphi-epidata/issues/998#issuecomment-1348732544).

- [X] Moves `parse_*` and `extract_*` methods to `params.py` instead of `validate.py`.
- [ ] Also does the same with several methods from `covidcast.py`.
- [ ] Also renames the methods to more accurately represent their purpose (pending some discussion with George)


melange396 (commented):
This looks pretty good, but out of curiousity: Did you do a straight copy of the 11 methods moved to `_params.py` from `_validate.py` and `covidcast.py`, or did you make changes to them?  Similarly, any changes in the test methods moved from `test_validate.py` to `test_params.py`?  I presume there are no such changes, but i cant easily tell from the diffs shown by github.

...

### also, optionally:
_(if you dont want to do the following stuff now, its not a problem -- i will just put these comments into #998 for a future pass)_

It would be nice if [for now] the only places `request` is used as an import is in `_params.py` (which is specifically used to get "parameters" from a "request"), in endpoints (which each legitimately process a "request"), in `main.py` (which is effectively an endpoint itself), and in `_common.py` (which uses it in the context of pre- and post-request app decorators)...  To that end, can we:
  - make the methods left in `_validate.py` take `request` as a function argument instead of as an import?
  - change `print_non_standard()` and `create_printer()` in `_printer.py` to take `format` as an argument instead of pulling it from the imported `request`?

...which leaves `_exceptions.py`, where the constructor for `EpiDataException` uses `request` in a hacky special-cased compatibility workaround.  The subclasses of the exception make use of it too.  We can leave that one alone [for now].

melange396 (approved):
looks good, we will just need to remember to refer to my comment above regarding `request` as an argument (in the subsequent PR)

PR #1065, opened 2023-01-17
Leonlu2/time values and issues

### LINKED ISSUE #672: covidcast: support issues=*, time_values=* queries
> Maybe it'd help to have a "*" option for time_values and issues as well, to avoid this epirange hack.

LeonLu2:
closes #672

**Prerequisites**:

- [X] Build is successful

### Summary

1. add the placeholder option for time_values and issues in queries
2. add unittests of time_values=* and issues=*


src/server/_params.py
```diff
@@ -477,6 +480,8 @@ def parse_time_set() -> TimeSet:
         # old version
         require_all("time_type", "time_values")
         time_values = extract_dates("time_values")
+        if len(time_values) == 1 and time_values[0] == "*":
```
krivard, line 483: is this something you tried?
```suggestion
        if time_values == ["*"]:
```

demo:
```
>>> ["*"] == ["*"]
True
>>> ["*"] == "*"
False
>>> ["abc"] == ["*"]
False
>>> ["*","abc"] == ["*"]
False
>>> [] == ["*"]
False
```

integrations/server/test_covidcast.py
```diff
@@ -73,6 +73,18 @@ def _insert_placeholder_set_four(self):
...
     ]
     self._insert_rows(rows)
+    return rows
+
+  def _insert_placeholder_set_five(self):
+    rows = [
+      self._make_placeholder_row(time_value=2000_01_01, value=i*1., stderr=i*10., sample_size=i*100., issue=2000_01_03+i)[0]
+      for i in [1, 2, 3]
+    ] + [
+      # issue intended to overlap with the issue above
```
krivard, line 83:
```suggestion
      # different time_values, same issues
```

krivard (dismissed):
Looks good! Small adjustments.

src/server/_params.py
```diff
@@ -381,7 +381,10 @@ def _parse_range(part: str):
 def parse_date(s: str) -> int:
     # parses a given string in format YYYYMMDD or YYYY-MM-DD to a number in the form YYYYMMDD
     try:
-        return int(s.replace("-", ""))
+        if s == "*":
+            return s
+        else:
+            return int(s.replace("-", ""))
```
melange396, line 387:
```suggestion
    if s == "*":
        return s
    try:
        return int(s.replace("-", ""))
```
the `if "*"` doesnt need to be in the `try` block (which is there to make sure `s` is made up of numbers (and dashes)), and the `else` is superfluous since the `return` happens right above it.

src/server/_query.py
```diff
@@ -474,8 +474,11 @@ def apply_lag_filter(self, history_table: str, lag: Optional[int]) -> "QueryBuil
 
     def apply_issues_filter(self, history_table: str, issues: Optional[TimeValues]) -> "QueryBuilder":
         if issues:
-            self.retable(history_table)
-            self.where_integers("issue", issues)
+            if issues == ["*"]:
+                self.retable(history_table)
+            else:
+                self.retable(history_table)
+                self.where_integers("issue", issues)
```
melange396, line 481:
```suggestion
            self.retable(history_table)
            if issues != ["*"]:
                self.where_integers("issue", issues)
```
you dont have to repeat the `retable()` line, and the `where_integers()` call is better done under the `if` instead of the `else` (by reversing the comparison).

melange396 (commented):
its a little late now since this was already merged, but i had a couple style points...

PR #1069, opened 2023-01-18
API server code health pass - reduce use of request global

rzats:
Partially addresses #998.

**Prerequisites**:

- [X] Build is successful

### Summary

Fixes a set of issues in the server code as described [here](https://github.com/cmu-delphi/delphi-epidata/issues/998#issuecomment-1348732544).

- [X] Makes the methods left in `_validate.py` take `request` as a function argument instead of as an import
- [X] Changes `print_non_standard()` and `create_printer()` in _printer.py to take `format` (retrieved from `request`) as a function argument

### DIFF
--- a/src/server/endpoints/covidcast_nowcast.py
+++ b/src/server/endpoints/covidcast_nowcast.py
@@ -19,8 +19,8 @@ alias = None
 
 @bp.route("/", methods=("GET", "POST"))
 def handle():
-    require_all(
-        request, "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
+    require_all(request,
+        "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
     )
     require_any(request, "geo_value", "geo_values", empty=True)
 
--- a/src/server/endpoints/covidcast_nowcast.py
+++ b/src/server/endpoints/covidcast_nowcast.py
@@ -20,9 +20,9 @@ alias = None
 @bp.route("/", methods=("GET", "POST"))
 def handle():
     require_all(
-        "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
+        request, "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
     )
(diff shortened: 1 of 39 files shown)


src/server/endpoints/covidcast_nowcast.py
```diff
@@ -20,9 +20,9 @@
 @bp.route("/", methods=("GET", "POST"))
 def handle():
     require_all(
-        "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
+        request, "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
```
melange396, line 23: how about moving `request` to the line above to separate it from the long list of keys that follow, which all logically group together?

itd look like:
```
    require_all(request,
         "data_source", "time_type", "geo_type", "time_values", "signals", "sensor_names"
```

src/server/endpoints/sensors.py
```diff
@@ -27,8 +27,8 @@
 PHP_INT_MAX = 2147483647
 
 
-def _authenticate(names: List[str]):
-    auth_tokens_presented = (resolve_auth_token() or "").split(",")
+def _authenticate(r: Request, names: List[str]):
+    auth_tokens_presented = (resolve_auth_token(r) or "").split(",")
```
melange396, line 31:
```suggestion
def _authenticate(req: Request, names: List[str]):
    auth_tokens_presented = (resolve_auth_token(req) or "").split(",")
```
how about `req` instead of `r`?  single-character variable names can be problematic outside of a few special cases

melange396 (changes requested):
This is all pretty good, except i dont love that `print_non_standard()` and `create_printer()` (from `src/server/_printer.py`) take `request` just to pull one value from it.  They could instead take `format` and can be called like `create_printer(request.values.get("format"))` (note that `"classic"` is never actually evaluated against, but _printer.py:257 will likely need to be changed to `if format and format.startswith("tree-"):` or similar)

rzats:
@melange396 fixes applied!

melange396 (approved):
looks good!

PR #949, opened 2022-08-02
[Refactor] Clarify a few `csv_importer` types and names

dshemetov:
A few small refactors for clarity that I couldn't help myself in making while working on #947.

Most of the changes are purely for better readability. No functionality is altered, except for the interface of `load_csv`, but this should not affect actual usage.

## Changes:
* rename RowValues to CsvRowValue
* make CsvRowValue into a dataclass
* remove `pandas` optional argument from `load_csv` that was only used for test mocks
* use `unittest.patch` to mock test instead
* update other tests
* update `csv_to_database` with a few relative imports

## Further details:
I changed 
```py
# From
def load_csv(filepath, geo_type, pandas=pandas):

# To
def load_csv(filepath, geo_type):
```
where `pandas` was just an abstract argument to the Pandas library meant for mocking. Instead of complicating this function's interface, we can just use `@patch("pandas.read_csv")` in the tests, which automatically mocks any calls to that function with our desired mock object.

**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

PR #1072, opened 2023-01-21
[Refactor] Further cleans of `csv_importer` 

dshemetov:
**Prerequisites**:

- [x] Build is successful

### Summary

A follow up to #949. There were just some pieces of this code that really bothered me cause they were unnecessarily confusing to read every time I went there.

- Remove a bunch of unnecessary function arguments by correctly patching in tests
- Add type hints to many functions
- Simplify the logic that returns a list of CovidcastRows for database import
- Convert the path details tuple to a NamedTuples, so it's less of a mystery

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -6,6 +6,7 @@ import re
 from dataclasses import dataclass
 from datetime import date
 from glob import glob
+from typing import Iterator, NamedTuple, Optional, Tuple
 
 # third party
 import epiweeks as epi
@@ -14,6 +15,10 @@ import pandas as pd
 # first party
 from delphi_utils import Nans
 from delphi.utils.epiweek import delta_epiweeks
+from delphi.epidata.acquisition.covidcast.database import CovidcastRow
+from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
+
+DFRow = NamedTuple('DFRow', [('geo_id', str), ('value', float), ('stderr', float), ('sample_size', float), ('missing_value', int), ('missing_stderr', int), ('missing_sample_size', int)])
 PathDetails = NamedTuple('PathDetails', [('source', str), ("signal", str), ('time_type', str), ('geo_type', str), ('time_value', int), ('issue', int), ('lag', int)])
 
 
@@ -180,7 +185,8 @@ class CsvImporter:
         yield (path, None)
         continue
 
-      yield (path, (source, signal, time_type, geo_type, time_value, issue_value, lag_value))
+      yield (path, PathDetails(source, signal, time_type, geo_type, time_value, issue_value, lag_value))
+
 
   @staticmethod
   def is_header_valid(columns):
@@ -344,7 +350,7 @@ class CsvImporter:
 
 
   @staticmethod
-  def load_csv(filepath, geo_type):
+  def load_csv(filepath: str, details: PathDetails) -> Iterator[Optional[CovidcastRow]]:
     """Load, validate, and yield data as `RowValues` from a CSV file.
 
     filepath: the CSV file to be loaded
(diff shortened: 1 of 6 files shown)


src/acquisition/covidcast/csv_to_database.py
```diff
@@ -89,20 +95,16 @@ def upload_archive(
...
       continue
 
-    (source, signal, time_type, geo_type, time_value, issue, lag) = details
-
-    csv_rows = csv_importer_impl.load_csv(path, geo_type)
-
-    cc_rows = CovidcastRow.fromCsvRows(csv_rows, source, signal, time_type, geo_type, time_value, issue, lag)
-    rows_list = list(cc_rows)
+    csv_rows = CsvImporter.load_csv(path, details)
+    rows_list = list(csv_rows)
```
krivard, line 107: praise: this is an enormous improvement

way more readable and less dependent on catching small errors in long tuple definitions. I have no idea why load_csv wasn't returning a CovidcastRow iterator to begin with.

tests/acquisition/covidcast/test_csv_importer.py
```diff
@@ -42,37 +42,38 @@ def test_is_sane_week(self):
     self.assertFalse(CsvImporter.is_sane_week(202054))
     self.assertFalse(CsvImporter.is_sane_week(20200418))
 
+
+  @patch("delphi.epidata.acquisition.covidcast.csv_importer.glob")
```
krivard, line 46: question: why does this patch location need to be `delphi.epidata.acquisition.covidcast.csv_importer` when the patch for `isdir` on the line below can be in `os.path`?

tests/acquisition/covidcast/test_csv_to_database.py
```diff
@@ -14,55 +15,62 @@
...
     """Scan the data directory."""
 
-    mock_csv_importer = MagicMock()
-    mock_csv_importer.find_csv_files.return_value = self._path_details()
-    collect_files(
-      "fake_data_dir",
-      False, # no specific issue
-      csv_importer_impl=mock_csv_importer)
+    mock_csv_importer.find_csv_files.return_value = self._path_details
+    collect_files("fake_data_dir",False) # no specific issue
```
krivard, line 38: nit: whitespace

```suggestion
    collect_files("fake_data_dir", False) # no specific issue
```

krivard (approved):
🏆 refactoring from `*_impl` to `@patch` has been on my irklist for ages

some questions and a suggestion but i wouldn't be mad if this merged as-is

experimentally using the Conventional Comments approach

src/acquisition/covidcast/csv_importer.py
```diff
@@ -1,20 +1,26 @@
...
 # standard library
-from dataclasses import dataclass
-from datetime import date
-import glob
 import os
 import re
+from dataclasses import dataclass
+from datetime import date
+from glob import glob
+from typing import Iterator, NamedTuple, Optional, Tuple
```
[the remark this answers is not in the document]
dshemetov: So I'm using a Python import sort extension in my IDE and it likes to place bare `import X` statements above `from X import Y` statements, but then sorts each of those, so that's convention I followed here.

tests/acquisition/covidcast/test_csv_importer.py
dshemetov: So the TLDR of this is due to Python imports
- `csv_importer` uses `import os`, so we can patch `os.path`, because `csv_importer` points to `os.path`
- `csv_importer` uses `from glob import glob`, which makes a new variable `glob` in the `csv_importer` namespace with the same value as `glob.glob`, but if we patch `glob.glob`, which updates that value to a Mock, `csv_importer.glob` retains its own value, so it's effectively not patched

So the end result is that you generally want to patch where a name is used, not a module itself.

[Here's a more complete blog post on this](https://nedbatchelder.com/blog/201908/why_your_mock_doesnt_work.html).

src/acquisition/covidcast/csv_importer.py
```diff
@@ -1,20 +1,26 @@
...
 
 # first party
 from delphi_utils import Nans
 from delphi.utils.epiweek import delta_epiweeks
-from .logger import get_structured_logger
+from delphi.epidata.acquisition.covidcast.database import CovidcastRow
+from delphi.epidata.acquisition.covidcast.logger import get_structured_logger
+
+DFRow = NamedTuple('DFRow', [('geo_id', str), ('value', float), ('stderr', float), ('sample_size', float), ('missing_value', int), ('missing_stderr', int), ('missing_sample_size', int)])
+PathDetails = NamedTuple('PathDetails', [('source', str), ("signal", str), ('time_type', str), ('geo_type', str), ('time_value', int), ('issue', int), ('lag', int)])
```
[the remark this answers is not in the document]
dshemetov: Yea, it's just a legacy ordering. I can change it to path ordering for now, though it may be something worth revisiting later.

src/acquisition/covidcast/csv_importer.py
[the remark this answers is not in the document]
dshemetov: I feel like this is an open ended question that we can address in a more comprehensive push for broader linting/formatting goals. Gonna keep as is for now.

dshemetov:
Thanks @krivard ! Just going to merge this one in, we can add any further feedback to another issue.

dshemetov:
Ok guess not, whenever @melange396 gets to it.

krivard:
@dshemetov yeah I have it set to dismiss reviews when new commits are added; we had an intern that merged some unreviewed code a while back and it seemed prudent at the time. it's kindof a pain if we don't need it though. would you prefer to wait for George or want me to override and merge now?

dshemetov:
I'd prefer to just merge it in, since George won't be back until Wednesday and I don't think this is that complicated of a change

PR #1080, opened 2023-02-09
Fix packages incompatibility with Apple M1 by defaulting to linux/amd64 containers

minhkhul:
### Summary

Due to [package incompatibility](https://github.com/Toblerity/Fiona/discussions/1071), this PR adjusts the Makefile to build and run affected containers using linux/amd64 by default.

### Testing
Local make all and make tests has no error and takes around 50s on my M1 macbook.
Screenshot 2023-02-09 at 12 05 41 PM

### DIFF
--- a/dev/local/Makefile
+++ b/dev/local/Makefile
@@ -37,15 +37,11 @@
 #   test=         Only runs tests in the directories provided here, e.g.
 #                 repos/delphi/delphi-epidata/tests/acquisition/covidcast
 #   sql=          Overrides the default SQL connection string.
-#   m1=1          Mandatory if your local system uses Apple M1 chip
 
 # Set optional argument defaults
-ifdef m1
-	override m1 =--platform linux/amd64
-endif
 
 ifdef pdb
-	override =--pdb
+	override pdb =--pdb
 else
 	pdb=
 endif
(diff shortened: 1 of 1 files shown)


krivard (approved):
Looks good, and thankyou for keeping the line lengths reasonable 🙏 

Verified running on my machine with these changes does not affect running time for `make all test`.

melange396 (changes requested):
looks pretty good!  two things though:

- (**_nit_**) it helps readability and is a good convention to put a space before each line-continuation backslash.

- (_**todo**_) you should make the `--platform linux/amd64` optional so that non-m1 environments arent forced to use it by default -- platform auto-selection seems to work properly for all but the m1 mac environment.  you could make an `m1` argument to force this behavior (see the `pdb` argument for reference), or even use the output of the command `uname -smp` to make that determination for you.

dev/local/Makefile
```diff
@@ -37,11 +37,15 @@
...
+#   m1=1          Mandatory if your local system uses Apple M1 chip
 
 # Set optional argument defaults
+ifdef m1
+	override m1 =--platform linux/amd64
+endif
+
 ifdef pdb
-	override pdb=--pdb
+	override =--pdb
```
krivard, line 48: issue: variable name removed from assignment, likely by accident

```suggestion
	override pdb=--pdb
```

krivard (changes requested):
👍 fulfills request for command-line switch

probably the change to pdb handling was unintentional and can be fixed directly from the suggestion; if it was on purpose can you say more about what you need it to do?

melange396 (dismissed):
fix the `pdb` thing, and it looks good!

minhkhul:
Improvement with uname use. No longer need optional param.

melange396 (approved):
automatic!  awesome!!!
i would put an `@echo` line inside your `ifeq` block that prints something like "Apple M1 environment detected, using 'linux/amd64' platform for compatibility", so that its more obvious that the alternative platform is being used.  but other than that, this is great!

minhkhul:
Added some code to notify user when M1 system is detected.
Screenshot 2023-02-13 at 2 20 10 PM

melange396 (approved):
excellent work!

melange396 (approved):

krivard (approved):
👍 great adjustments!

PR #1084, opened 2023-02-13
Linters

BrainIsDead:


dshemetov:
Can we wait to merge this until after #646? It would make it difficult to resolve conflicts with `dev` there.

dshemetov:
Should be fine to go ahead, since JIT is shelved ATM

melange396:
This is out of date now and it is HUGE so fixing conflicts will not be fun...  I think these changes were mostly automated though, so re-applying it to the current codebase should be straightforward if we choose to go that route.  Closing.

PR #1091, opened 2023-02-21
Use epidata client in covidcast integration tests + fix csv format bug + fix fields param bug

### LINKED ISSUE #67: use epidata client in covidcast integration tests
the covidcast integration tests should use the epidata client now that it supports the covidcast endpoints

minhkhul:
- Use epidata client in covidcast integration tests 
- fix format = csv bug 
- fix fields param bug

fix #67

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -11,6 +11,7 @@ import requests
 # first party
 from delphi_utils import Nans
 from delphi.epidata.acquisition.covidcast.test_utils import CovidcastBase, CovidcastTestRow
+from delphi.epidata.client.delphi_epidata import Epidata
 
 # use the local instance of the Epidata API
 BASE_URL = 'http://delphi_web_epidata/epidata/api.php'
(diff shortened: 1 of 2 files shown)


integrations/server/test_covidcast.py
```diff
@@ -194,7 +193,7 @@ def test_fields(self):
     row = self._insert_placeholder_set_one()
 
     # limit fields
-    response = self.request_based_on_row(row, fields='time_value,geo_value')
+    response = self.request_based_on_row(row, **{"fields":"time_value,geo_value"})
```
krivard, line 196: question: did you ever figure out why the **{} is needed?

it's fine this way, just curious

krivard (approved):
👍 Solid choices!

integrations/server/test_covidcast.py
minhkhul: Both ways work now. I choose to keep it like this to be consistent with how extra arguments are added in other test functions, but having it as fields='time_value,geo_value' also works.

PR #1092, opened 2023-02-22
deleted *.php from /server and docs, deleted Dockerfile

BrainIsDead:
closes 1001

### Summary

Deleted `*.php` files from `/server` dir and `docs`, deleted `Dockerfile` related to php files


BrainIsDead:
@melange396  @rzats  there are `cdc_upload.php`, `dasboard.php` and `master.php` in `./src/acquisition/*/*`:
*  `cdc_upload.php` meets once in `/delphi-epidata/deploy.json`
* `dashboard.php` should be available by the link `https://delphi.cmu.edu/~automation/public/wiki/` but it returns 404 for me, also meets in `/delphi-epidata/deploy.json` 
* `master.php` is for rearing `wiki_raw` table independently and in parallel to `wiki_download.py`,  and looks loke exists in `wiki_download` as `MASTER_URL = 'https://delphi.cmu.edu/~automation/public/wiki/master.php'` (but link returns 404 for me). Also meets in `/delphi-epidata/deploy.json` 

Not sure what I should to do with these files

melange396 (changes requested):
Regarding `./src/acquisition/*/*.php`, those should be deleted as well (and the references in `deploy.json`, of course).  We havent had anything running them for a long time, so they will not be missed.

Could you also update `docs/epidata_development.md` to remove references to the `delphi_web` image, and update the `delphi_web_epidata` image build statement from the `Dockerfile` being deleted in this PR to the current proper `Dockerfile` location (as in the dev `Makefile`)?  There are other anachronisms in that file, but i would say we can leave them alone for now.

melange396 (approved):
awesome!  i am very glad to see this stuff disappear!!!

PR #1093, opened 2023-02-23
[JIT] Implement non-streaming JIT with `polars`

dshemetov:
A variant of #1014 using [Polars](https://www.pola.rs/) instead of Pandas.

**Prerequisites**:

- [x] Build is successful

### Summary

A reimplementation of the internals of the JIT system using Pandas. My local machine benchmarking showed that the Polars implementation might be 2x as fast as the Pandas. We will see how it works in Locust.

In general, our JIT approaches are bottlenecked by two main components: data loading and data dumping. The data loading refers to loading data from the database into a fast memory format (like Numpy arrays or Pandas DataFrame or Polars DataFrame). This is slow because the the database result is first handled by Python, which then traverses through each row in the result (using a slow Python loop), and loads them into the new fast format. The data dumping part requires us convert back from the fast format into something we can send to the client. This is extremely slow if we need to convert back to a list of Python rows, which is then made even slower by the overhead in our API server's APrinter that handles writing the results to json or csv. 

Fortunately, both Pandas and Polars have a way to go directly to JSON, using a fast internal method (e.g. `pd.DataFrame.to_json(orient="records")` and `pl.DataFrame.write_json(row_oriented=True)`). This has made #1014 competitive performance-wise. From my local benchmarking, it appears that most of the gains in this branch over the Pandas branch comes from improvements to the speed of this step.

The drawback however, is that because I bypassed APrinter, I rewrote a lot of the logic it handles. This part of the code will need to be addressed and double-checked again before release into production. 

TODO: benchmark a separate branch of `dev` that does not include JIT, but loads the query results to Polars and uses write_json to bypass APrinter. I suspect that this might be a substantial performance improvement.


dshemetov:
Update on the derived signals benchmark:
- `dev` runs in ~2 minutes
- `jit-pandas` runs in ~3:30 minutes
- `jit-polars` runs in ~3 minutes

krivard:
issue: branch not up to date with `dev`

this PR is still marked as a draft, but FYI several of the commits in it seem to be rebased versions of commits dev already has. that makes it difficult to find the new functionality in the diff so that folks can provide feedback.

e.g.
* This PR has https://github.com/cmu-delphi/delphi-epidata/commit/869ff216272fe6f422d6f0244d0dba0eb536b936
* Dev already has these changes as part of https://github.com/cmu-delphi/delphi-epidata/commit/c5c1989a127fb57f332c49e5a72f4fe2c4ae35ef

Could you build a link to the files pane that highlights the commits you need feedback on most urgently? I made an attempt, but [this one was too broad](https://github.com/cmu-delphi/delphi-epidata/pull/1093/files/aff6036d15d53e6a75ac9f1f9279fa1393f41b95..e0e0d038e263dadde9d5ae705b2396fd20144a2e) and [this one was too narrow](https://github.com/cmu-delphi/delphi-epidata/pull/1093/files/72988622ec5dd8deb21c82f39697428de9edf647..e0e0d038e263dadde9d5ae705b2396fd20144a2e).

dshemetov:
Sorry yea, this is definitely not ready for viewing rn. I'm going to rebase soon.

PR #1096, opened 2023-02-24
Update performance scaling test rig

rzats:
**Prerequisites**:

- [X] Build is successful

### Summary

Adds several updates to the performance scaling test rig:
- Now clones `cmu-delphi/delphi-admin` repo, so files do not need to be duplicated to this repo.
  - The test files added to this repo have been deleted accordingly.
- Changes the approach to selecting queries performance testing - instead of running queries from a small sample for a fixed amount of time, runs every query from a larger sample once; this is more consistent across multiple runs.
- Writes Locust output to an artifact in the GitHub Actions run.


.github/workflows/performance-tests.yml
```diff
@@ -52,11 +52,29 @@ jobs:
...
           cd ../driver
           sudo make web sql="${{ secrets.DB_CONN_STRING }}"
+      - name: Check out delphi-admin
+        uses: actions/checkout@v3
+        with:
+          repository: cmu-delphi/delphi-admin
+          token: ${{ secrets.GH_PERSONAL_TOKEN }}
+          path: delphi-admin
       - name: Build & run Locust
+        continue-on-error: true # sometimes ~2-5 queries fail, we shouldn't end the run if that's the case
```
melange396, line 62: what are the failures?  they only happen transiently?

melange396 (approved):
looks pretty good!

on the wishlist: it might be nice to have summary info about the results of the run included in the follow-up comment, like the running time of the `docker run locust` command or a count/percent of the failed queries.  [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs) looks like a good way to get information passed between the jobs.

.github/workflows/performance-tests.yml
rzats: There are two queries from the "v4 - small" set that consistently fail.

.github/workflows/performance-tests.yml
```diff
@@ -52,11 +52,29 @@ jobs:
         run: |
           cd ../driver
           sudo make web sql="${{ secrets.DB_CONN_STRING }}"
+      - name: Check out delphi-admin
+        uses: actions/checkout@v3
+        with:
+          repository: cmu-delphi/delphi-admin
+          token: ${{ secrets.GH_PERSONAL_TOKEN }}
```
korlaxxalrok, line 59:
```suggestion
          token: ${{ secrets.CMU_DELPHI_DEPLOY_MACHINE_PAT }}
```

korlaxxalrok (approved):
LGTM 👍

.github/workflows/performance-tests.yml
```diff
@@ -52,11 +52,29 @@ jobs:
...
+          token: ${{ secrets.CMU_DELPHI_DEPLOY_MACHINE_PAT }}
+          path: delphi-admin
       - name: Build & run Locust
+        continue-on-error: true # sometimes ~2-5 queries fail, we shouldn't end the run if that's the case
         run: |
-          cd ../driver/repos/delphi/delphi-epidata/tests/performance
+          cd delphi-admin/load-testing/locust
           docker build -t locust .
-          docker run --net=host -v $PWD:/mnt/locust -e CSV=/mnt/locust/v4-requests-smaller.csv locust -f /mnt/locust/v4.py --host http://127.0.0.1:10080/ --users 10 --spawn-rate 1 --headless -t 15m
+          export CSV=v4-requests-small.csv
```
krivard, line 66: question: why move the CSV filename into a variable?

(is it a stub for being able to override it in the future?)

krivard (approved):
👍 happy to have this merged as-is; question is largely out of curiosity

.github/workflows/performance-tests.yml
rzats: It's just to avoid repeating the same filename twice, since it's now used in the `wc -l` bit as well.

PR #1108, opened 2023-03-09
add indicator_name

minhkhul:


korlaxxalrok (dismissed):
Looks good! I will give my 👍 , but I think that @krivard and/or @melange396 should put eyes on it as well.

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -28,16 +28,21 @@ def get_argument_parser():
   parser.add_argument(
     '--log_file',
     help="filename for log output (defaults to stdout)")
+  parser.add_argument(
+    '--indicator_name',
+    nargs='?',
```
melange396, line 33: whats the purpose of this option?  do we need it?

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -28,16 +28,21 @@ def get_argument_parser():
...
+    default='*',
+    help='Name of one indicator directory to run acquisition on')
   return parser
 
 
-def collect_files(data_dir: str, specific_issue_date: bool):
+def collect_files(data_dir: str, specific_issue_date: bool, indicator_name = "*"):
```
melange396, line 39:
```suggestion
def collect_files(data_dir: str, specific_issue_date: bool, indicator_name="*"):
```
nit: when specifying keyword arguments, the convention is to omit spaces around the equals

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -28,16 +28,21 @@ def get_argument_parser():
...
   """Fetch path and data profile details for each file to upload."""
   logger= get_structured_logger('collect_files')
   if specific_issue_date:
     results = list(CsvImporter.find_issue_specific_csv_files(data_dir))
   else:
-    results = list(CsvImporter.find_csv_files(os.path.join(data_dir, 'receiving')))
+    results = list(CsvImporter.find_csv_files(os.path.join(data_dir, 'receiving'), indicator_dir = indicator_name))
```
melange396, line 45:
```suggestion
    results = list(CsvImporter.find_csv_files(os.path.join(data_dir, 'receiving'), indicator_dir=indicator_name))
```
nit: when specifying keyword arguments, the convention is to omit spaces around the equals

src/acquisition/covidcast/csv_to_database.py
```diff
@@ -146,9 +151,8 @@ def main(args):
 
   logger = get_structured_logger("csv_ingestion", filename=args.log_file)
   start_time = time.time()
-
   # shortcut escape without hitting db if nothing to do
-  path_details = collect_files(args.data_dir, args.specific_issue_date)
+  path_details = collect_files(args.data_dir, args.specific_issue_date, indicator_name = args.indicator_name)
```
melange396, line 155:
```suggestion
  path_details=collect_files(args.data_dir, args.specific_issue_date, indicator_name = args.indicator_name)
```
nit: when specifying keyword arguments, the convention is to omit spaces around the equals

melange396 (changes requested):
nice job!  just a couple little comments...

src/acquisition/covidcast/csv_to_database.py
minhkhul: It's to run acquisition on one specific indicator. We do need it.

A bit of context: the final goal is to run acquisition immediately after each specific indicator run on Cronicle, so the database gets csv files immediately when new data is available. 

Right now on Automation, acquisition runs at a specific time, on any indicators with new csv available at the time. Since acquisition process checks all indicator directories under data_dir/receiving, there's been some issue with acquisition job trying to access a indicator dir that's being written to by an indicator run at the same time on Cronicle, causing failure to both processes. This change allows the option to limit such checks to one specific indicator_dir under data_dir/receiving only.  

This fix hopefully will help separate these processes as much as possible so they don't step on each other's foot.  

I have tested both the [optional arg](https://github.com/cmu-delphi/delphi-epidata/pull/1108/commits/ccad57f167607bbe4fbe760862ca1edee03eb61a) and [lacking thereof](https://github.com/cmu-delphi/delphi-epidata/pull/1108/commits/ab913234dd1370797eb27acec538ba01b213db35). test_csv_uploading.py pass in both cases, meaning if the argument is there, csv_uploading works, and if the argument is not there, csv_uploading also works.

melange396: whoops!  that comment was supposed to be attached to line 33 -- i was asking about the `nargs` argument to the `add_argument()` call.  🤦

melange396: specifying `default` does that by itself, making `nargs` unnecessary.  i think you should take it out to avoid confusion but it doesnt break anything to leave it there.  ¯\\\_(ツ)_/¯

melange396 (approved):
👍

PR #1109, opened 2023-03-15
change github perftest workflow to use as_of query set

melange396:
this changes the github action/workflow for performance testing to use the ["as_of" set of sample queries](https://github.com/cmu-delphi/delphi-admin/blob/main/load-testing/locust/v4-requests-as_of.csv).

question: will this need to be merged into the `main` branch before the action/workflow uses the new query set, or will it work after being merged into `dev`?


rzats (approved):

rzats:
@melange396 during testing it worked after being merged into `dev`. (However, the changes won't be reflected in the run made by the comment you left just now)

PR #1110, opened 2023-03-15
(potential) as_of JOIN improvement

melange396:
[As pointed out by @lcbrooks](https://delphi-org.slack.com/archives/C013AH5N01E/p1678867932010249), our `as_of` queries may potentially benefit from a small v4-schema-aware change to the JOIN condition on the sub-query...  This is an exploration on that.

### DIFF
--- a/src/server/_query.py
+++ b/src/server/_query.py
@@ -491,7 +491,7 @@ class QueryBuilder:
             sub_fields = "max(issue) max_issue, time_type, time_value, `source`, `signal`, geo_type, geo_value"
             sub_group = "time_type, time_value, `source`, `signal`, geo_type, geo_value"
             alias = self.alias
-            sub_condition = f"x.max_issue = {alias}.issue AND x.time_type = {alias}.time_type AND x.time_value = {alias}.time_value AND x.source = {alias}.source AND x.signal = {alias}.signal AND x.geo_type = {alias}.geo_type AND x.geo_value = {alias}.geo_value"
+            sub_condition = f"x.max_issue = {alias}.issue AND x.time_type = {alias}.time_type AND x.time_value = {alias}.time_value AND x.source = {alias}.source AND x.signal = {alias}.signal AND x.geo_type = {alias}.geo_type AND x.geo_value = {alias}.geo_value" # placeholder
             self.subquery = f"JOIN (SELECT {sub_fields} FROM {self.table} WHERE {self.conditions_clause} AND {sub_condition_asof} GROUP BY {sub_group}) x ON {sub_condition}"
         return self
 


melange396:
i fixed the sample query file to remove the 40 bad queries (!) and repeated the two good ones 5x each...  trying again!
/run performance test

melange396:
updated JOIN condition, getting running time again:
/run performance test

melange396:
hmmm, thats actually worse performance!  trying again to see if maybe there was a transient network event or something between the github actions platform and our database...
/run performance test

melange396:
is this repeatable?  trying again...
/run performance test

melange396:
it seems performance tests cannot be queued up. ¯\\\_(ツ)_/¯

there is a lot of variance in the timing of these results... ill try to come up with a bigger query set to see if we can reduce that.

melange396:
for good measure:
/run performance test

brookslogan:
You might be able to remove
```
`source`, `signal`, geo_type, geo_value
```
from the `sub_fields` in case they are tricking the SQL engine to actually perform the join to get those columns in the subquery as well as the main query.

(Or maybe `PLAN` on the original might reveal whether or not the optimizer/planner is smart enough to automatically do some transformation of the subquery join already?)

melange396:
@lcbrooks we can drop those from `sub_fields`, but some of them will still be used when filtered on in the WHERE clause of the subquery, which is inherited from the WHERE of the encapsulating query.

i think the planner and optimizer tend to be pretty smart but they are obviously not perfect.  as far as evaluating them, though, you cant just EXPLAIN one query and get a universal plan for all subsequent queries of a similar form -- the results depend on and will change with various constraints that are included in the query (in particular for the `as_of` queries in question, that would be different dates/ranges and geos).  the EXPLAIN plan will sometimes also show some really dumb things, like if you join a table to another but only SELECT values from the first, [the plan](https://gist.github.com/melange396/2ad242bf57ad6fa6f4abb1aef58746b3) still shows it including the second table even though it has no bearing on the results and can be effectively ignored.  that manifests for us in that, since we use VIEWs to mimic v3 tables, there are inherent JOINs to the dimension tables that the planner/optimizer seems to [always include](https://gist.github.com/melange396/d02e2e8656097755a2c602fab49ce53b).

ultimately, i think it may be more appropriate for us to come up with a large representative set of `as_of` queries and use those to measure performance instead of relying on analyzing cryptic EXPLAIN plans.  i just spent too long investigating alternative options for our `as_of` query format ("too long" because i have other pressing work things that i couldve spent that time on), though i do have two new variations (_which this margin is too narrow to contain_) that could carry performance improvements.  i will return to this in the near future to come up with a better set of sample "as_of" API queries and then evaluate the runtime of those queries against the aforementioned `as_of` query constructions.

melange396:
i havent updated the query set yet but, for posterity, im gonna try another concurrent test
/run performance test

brookslogan:
The `PLAN`s you posted seem pretty important.  Actually, I think the engine is right to think it *has to* perform the join: because we are using an `(INNER) JOIN` without `FOREIGN KEY` constraints, the first `epimetric_id` in `epimetric_full` might actually map to 0 rows in `epimetric_full_v` and it has to join to figure this out.  Either using a `LEFT JOIN` or adding `FOREIGN KEY` constraints might address this issue.

PR #1103, opened 2023-03-01
 1078 - Refactor csv_importer.py and csv_to_database.py

BrainIsDead:
closes [1078](https://github.com/cmu-delphi/delphi-epidata/issues/1078)

* `csv_to_database ` functionality moved to `csv_importer ` (csv_to_database  deleted)
* fixed imports
* changed mocks
* added  pandas errors exceptions to `pd.read_csv` in the `load_csv` method
* changed `is_header_valid` method with `issubset` method

PR #1116, opened 2023-03-31
Refactor csv_importer.py with Pandas

dshemetov:
Updates #1103 

**Prerequisites**:

- [x] Build is successful

### Summary

I meant to make only a few specific changes, but I ended up making a lot of changes.
* `pandas.DataFrame.apply` is just a Python for-loop under the hood, which is slow, so I rewrote these methods to use the built-ins, which rely on fast numpy functions
* we don't need to fall back to `str` dtypes in the Pandas dataframe: it complicates code downstream, makes the code slower (because of redundant dtype casting), and, based on my log-digging in [this thread](https://delphi-org.slack.com/archives/C013AH5N01E/p1680210412446129), the code path hasn't occurred in over a year (and only then occurred once due to a bug which was fixed with a data type change)
* fixed/updated a few exception handlings: 
  * ValueError in the first exception in load_csv made the second exception statement unreachable, so removed it
  * added GeoTypeSanityCheckException, ValueSanityCheckException to handle a couple other exceptions
* removed the old staticmethods that are unused after these changes
* changed tests to conform to new data type assumptions and overall fixed tests to pass

### DIFF
--- a/src/acquisition/covidcast/csv_importer.py
+++ b/src/acquisition/covidcast/csv_importer.py
@@ -12,9 +12,10 @@ from dataclasses import dataclass
 from datetime import date
 from glob import glob
 from logging import Logger
-from typing import Callable, Iterable, Iterator, NamedTuple, Optional, Tuple
+from typing import Callable, Iterable, List, NamedTuple, Optional, Tuple
 
 import epiweeks as epi
+import numpy as np
 import pandas as pd
 from delphi.epidata.acquisition.covidcast.covidcast_row import CovidcastRow
 from delphi.epidata.acquisition.covidcast.database import Database, DBLoadStateException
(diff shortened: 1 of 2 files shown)


BrainIsDead:
Wow!!!!!! Looks perfect you are definitely better in pandas than me
image

dshemetov:
Haha! I just spent the past 6 months trying to make JIT fast and that involved figuring out what's fast and what's not in Pandas, so I have a much better sense now than before. Most of that involved speed profiling, but here are a few good articles:
- It also helps to understand that Pandas stores its data as Numpy arrays and then applies [fast vectorized Numpy methods](https://pythonspeed.com/articles/pandas-vectorization/) to do its computations
- Here is a [decent SO thread about `apply`](https://stackoverflow.com/questions/54432583/when-should-i-not-want-to-use-pandas-apply-in-my-code)

BrainIsDead:
> Haha! I just spent the past 6 months trying to make JIT fast and that involved figuring out what's fast and what's not in Pandas. Most of that involved speed profiling, but here are a few good articles:
> 
> * It helps to understand that Pandas stores its data as Numpy arrays and the built-ins use [fast vectorized Numpy methods that use C](https://pythonspeed.com/articles/pandas-vectorization/)
> * [`apply` on the other hand, uses Python for-loops and function calls](https://stackoverflow.com/questions/54432583/when-should-i-not-want-to-use-pandas-apply-in-my-code)

Thanks

BrainIsDead (approved):

PR #1121, opened 2023-04-05
get_geo_values() doesnt know about a plain "state" type

melange396:
but it does know about "state_id", "state_code", and "state_name" (see [delphi_utils/geomap.py](https://github.com/cmu-delphi/covidcast-indicators/blob/2f92ca7ecf88c715bf152bd23e256e22347c50ae/_delphi_utils_python/delphi_utils/geomap.py#L545-L546)).  "state_id" appears to be the two-digit state abbreviation, if [delphi_utils/data/2020/state_pop.csv](https://github.com/cmu-delphi/covidcast-indicators/blob/main/_delphi_utils_python/delphi_utils/data/2020/state_pop.csv) is to be believed.

this is untested!  i whipped it up fast to address the issue we saw post-rollout.  its kinda hacky and i dont love the ternary form of it.  ...but it should do the trick, at least in prod for now.

obviously, it should be accompanied by some new additional unit and/or integration tests that involve a "state" geo_type.

### DIFF
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -218,7 +218,7 @@ class DelphiEpidataPythonClientTests(CovidcastBase):
     # insert placeholder data: three counties, three MSAs
     N = 3
     rows = [
-      CovidcastTestRow.make_default_row(geo_type="fips", geo_value=FIPS[i], value=i)
+      CovidcastTestRow.make_default_row(geo_type="county", geo_value=FIPS[i], value=i)
       for i in range(N)
     ] + [
       CovidcastTestRow.make_default_row(geo_type="msa", geo_value=MSA[i], value=i*10)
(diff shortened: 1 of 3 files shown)


melange396:
this feels much less hacky.  added geo_type "county" handling, and it now steps around geo_types that are unknown to GeoMapper (like "dma").  we may want to keep the just-added `geo_type_translator` in sync with `GEOGRAPHIC_RESOLUTIONS` from [csv_importer.py](https://github.com/cmu-delphi/delphi-epidata/blob/4e8b410e50d62f952ca874bef55b7917691c38e4/src/acquisition/covidcast/csv_importer.py#L66)

now this is failing tests, presumably because tests use a geo_type of "fips" when they should specify "county" instead.

krivard:
I'm working on a test now

krivard (dismissed):
I added and fixed tests, lmk if they're friendly and we can merge and release

melange396:
tests look good.  i added a note about syncing geo_type.  i think this should be good to go!

krivard (approved):
🙌 stellar addition

PR #1124, opened 2023-04-06
Leonlu2/inequality issue time value

### LINKED ISSUE #1085: Permit inequalities for `issue`, `time_value`
With #1070 we now support `time_value=*` and `issue=*` in covidcast requests, which prevents users from having to specify ridiculous ranges like 11111111-99999999 in their queries. But if a user wants an open-ended portion of the complete calendar or history, they still have to put in silly values.

Could we support inequalities like the below?

* all issues on or before some date
* all issues on or after some date
* all time_values on or before some date
* all time_values on or after some date

I'm not sure what a good syntax would be. Some ideas are:

* `<YYYYmmdd` `<=YYYYmmdd` `>YYYYmmdd` `>=YYYYmmdd`
* `*-YYYYmmdd` `YYYYmmdd-*`

We'd then have the following mapping to SQL queries:

| meaning | parameter syntax | SQL |
| - | - | - |
| one date | `YYYYmmdd` or `YYYY-mm-dd` | `issue = YYYYmmdd` or `time_value = YYYYmmdd` |
| range | `YYYYmmdd-YYYYmmdd` or `YYYY-mm-dd--YYYY-mm-dd` | `issue BETWEEN YYYYmmdd AND YYYYmmdd` or `time_value BETWEEN YYYYmmdd AND YYYYmmdd` |
| any | `*` | `issue` or `time_value` not included in WHERE clause |
| open-ended | TBD | `issue < YYYYmmdd` or `issue <= YYYYmmdd` or `issue > YYYYmmdd` or `issue >= YYYYmmdd` or similar for time_value |

LeonLu2:
closes #1085

**Prerequisites**:

- [X] Build is successful

### Summary

1. add inequality operators handler for time_values and issues in server
2. add inequality tests in test_covidcast.py accordingly
3. other minor changes

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -81,6 +81,18 @@ class CovidcastTests(CovidcastBase):
     ]
     self._insert_rows(rows)
     return rows 
+  
+  def _insert_placeholder_set_six(self):
+    rows = [
+      CovidcastTestRow.make_default_row(time_value=2000_01_01+i, value=i*1., stderr=i*10., sample_size=i*100., issue=2000_01_03)
+      for i in [1, 2, 3]
+    ] + [
+      # 
+      CovidcastTestRow.make_default_row(time_value=2000_01_01+i, value=i*1., stderr=i*10., sample_size=i*100., issue=2000_01_03)
+      for i in [4, 5, 6]
+    ]
+    self._insert_rows(rows)
+    return rows 
 
   def test_round_trip(self):
     """Make a simple round-trip with some sample data."""
@@ -167,7 +179,6 @@ class CovidcastTests(CovidcastBase):
          .assign(direction = None)
          .to_csv(columns=column_order, index=False)
     )
-
     # assert that the right data came back
     self.assertEqual(response, expected)
 
@@ -231,7 +242,6 @@ class CovidcastTests(CovidcastBase):
     # make the request
     response = self.request_based_on_row(rows[0], geo_value="*")
 
-    self.maxDiff = None
     # assert that the right data came back
     self.assertEqual(response, {
       'result': 1,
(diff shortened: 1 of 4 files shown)


integrations/server/test_covidcast.py
```diff
@@ -267,14 +338,63 @@ def test_issues_wildcard(self):
...
 
-    self.maxDiff = None
     # assert that the right data came back
     self.assertEqual(response, {
       'result': 1,
       'epidata': expected,
       'message': 'success',
     })
 
+  def test_issues_inequality(self):
```
krivard, line 348: suggestion: refactor to remove duplication

The tests for this and `test_time_value_inequality` are super similar. Would it make sense to do something like this instead?

```python
def helper_test_inequality(self, rows, time_field):
    expected = ...
    def fetch(datecode):
        # make the request
        response = self.request_based_on_row(rows[0], **{time_field: datecode})
        return response
    r = fetch(...
    ...

def test_time_value_inequality(self):
    self.helper_test_inequality(self._insert_placeholder_set_six(), "time_value")

def test_issues_inequality(self):
    self.help_test_inequality(self._insert_placeholder_set_seven(), "issue")
```

src/server/_params.py
```diff
@@ -425,6 +426,17 @@ def extract_dates(key: Union[str, Sequence[str]]) -> Optional[TimeValues]:
...
 
+    def handle_inequality(part: str):
+        inequality_operator = None
+        for operator in ['<=', '>=', '<', '>']:
+            if part.startswith(operator):
+                inequality_operator = operator
+                part = part[len(operator):]
+                if not part:
+                    raise ValidationFailedException("missing parameter: date after the inequality operator")
+                return (inequality_operator,), part
```
krivard, line 437: nit: using a length-1 tuple here feels weird

later on we're using the tuple type information to engage the inequalities logic instead of the range logic, so it's clearly needed, but it feels unsatisfying regardless

krivard (approved):
👍 

Could merge as is but I'd like to hear Rostyslav's thoughts

src/server/_params.py
krivard: hey @rzats do you have any ideas for how we can more clearly indicate that a date parameter value is an inequality? the return value from this function is read in the date span processing at _query.py:45, dates.py:125, and dates.py:147, where inequalities need to be handled differently from other kinds of spans.

rzats: Yep, I can see what this PR is going for and there might be a better way to handle the inequalities. Right now this is returning a single-element tuple to fit within the following type structure;

```
IntRange: TypeAlias = Union[Tuple[int, int], int]
TimeValues: TypeAlias = Sequence[IntRange]
```

A clean solution would be to add a new type alias called `IntStrRange` which is a union of `Tuple[int, int]`, `int` and a third type - either `str` if we just want to know what the inequality operator is or `Tuple[str, int]` if we want the date too.

`TimeValues` can then be replaced with a `Sequence[IntStrRange]`, and we can then refactor the date span processing functions where `TimeValues` was previously used.

PR #1126, opened 2023-04-07
[WIP] Split covid_hosp into daily & timeseries tables

rzats:
**Prerequisites**:

- [x] Build is successful

### Summary

Currently, the `covid_hosp_state_timeseries` table stores data from two datasets at once - daily snapshots and timeseries data; the nature of this data is distinguished by the `record_type` column. This PR by:

- Updating the `state_daily` DDL with two tables instead of one
- Creating a migration SQL from one table into two
- [TBD] Modifying the timeseries endpoint to use two tables instead of one


krivard:
Sonarcloud is overzealous

src/ddl/covid_hosp.sql
```diff
@@ -500,15 +489,151 @@ CREATE TABLE `covid_hosp_state_timeseries` (
...
+  KEY `state_by_issue_and_date` (`issue`, `date`, `state`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 
+/*
+`covid_hosp_state_timeseries` stores the versioned "state timeseries" dataset,
+which contains data from the daily snapshot files.
+Schema is equivalent to `covid_hosp_state_timeseries`.
+*/
+CREATE TABLE `covid_hosp_state_daily` (
```
melange396, line 506: since this essentially duplicates `covid_hosp_state_timeseries`, you can use a shortcut to create it...  see [v4_schema.sql](https://github.com/cmu-delphi/delphi-epidata/blob/4e46532388de61e30ed010334c4ad0f0f11f4749/src/ddl/v4_schema.sql#L54-L59) for an example.

src/ddl/migrations/covid_hosp_state_split_tables.sql
```diff
@@ -0,0 +1,148 @@
+-- 1. Add new state_daily table mirroring state_timeseries table
+
+CREATE TABLE `covid_hosp_state_daily` (
```
melange396, line 3: since this essentially duplicates `covid_hosp_state_timeseries`, you can use a shortcut to create it...  see [v4_schema.sql](https://github.com/cmu-delphi/delphi-epidata/blob/4e46532388de61e30ed010334c4ad0f0f11f4749/src/ddl/v4_schema.sql#L54-L59) for an example.

src/ddl/migrations/covid_hosp_state_split_tables.sql
```diff
@@ -0,0 +1,148 @@
...
+WHERE record_type='D';
+
+-- 3. Remove that data from the old table
+
+DELETE FROM `covid_hosp_state_timeseries` WHERE record_type='D';
+
+-- 4. Remove the record_type column from both tables
+
+ALTER TABLE `covid_hosp_state_daily` DROP COLUMN record_type;
+ALTER TABLE `covid_hosp_state_timeseries` DROP COLUMN record_type;
```
melange396, line 148: record_type is included in some indexes in `covid_hosp_state_timeseries`...  does this DROP COLUMN make any complaints related to that?

src/acquisition/covid_hosp/common/database.py
```diff
@@ -45,7 +45,7 @@ def __init__(self,
     self.connection = connection
     self.table_name = table_name
     self.hhs_dataset_id = hhs_dataset_id
-    self.publication_col_name = "issue" if table_name == 'covid_hosp_state_timeseries' else \
+    self.publication_col_name = "issue" if table_name == 'covid_hosp_state_timeseries' or table_name == "covid_hosp_state_daily" else \
```
melange396, line 48: this line is getting a little gnarly, but we will fix it with the other yaml-schema changes so you can leave it for now

integrations/acquisition/covid_hosp/state_timeseries/test_scenarios.py
```diff
@@ -52,9 +53,9 @@ def test_acquire_dataset(self):
         self.test_utils.load_sample_dataset()
 
     # make sure the data does not yet exist
-    with self.subTest(name='no data yet'):
-      response = Epidata.covid_hosp('MA', Epidata.range(20200101, 20210101))
-      self.assertEqual(response['result'], -2)
+    # with self.subTest(name='no data yet'):
+    #  response = Epidata.covid_hosp('MA', Epidata.range(20200101, 20210101))
+    #  self.assertEqual(response['result'], -2)
```
melange396, line 58: is this not passing any more?

integrations/server/test_covid_hosp.py
```diff
@@ -42,10 +48,10 @@ def test_query_by_issue(self):
...
         # inserting out of order to test server-side order by
         # also inserting two for 20201201 to test tiebreaker.
-        self.insert_issue(cur, 20201201, 123, 'T')
-        self.insert_issue(cur, 20201201, 321, 'D')
-        self.insert_issue(cur, 20201203, 789, 'T')
-        self.insert_issue(cur, 20201202, 456, 'T')
+        self.insert_timeseries(cur, 20201201, 123)
+        self.insert_daily(cur, 20201201, 321)
+        self.insert_timeseries(cur, 20201203, 789)
+        self.insert_timeseries(cur, 20201202, 456)
```
melange396, line 54: nice, clean change!  👍

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,20 @@ def handle():
     q.where_integers("date", dates)
     q.where_strings("state", states)
 
+    merge_tables = "(SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` UNION ALL SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries`) c"
```
melange396, line 155: the logic is perfect, but i fear that using this approach will cause a materialization of the entirety of the underlying tables...  can you do some EXPLAINs on the resulting queries to see if thats the case?

melange396 (changes requested):
nicely done!  i think the suggestions i made in the .sql files might make sonarcloud chillax

krivard:
I've swapped the quality gate profile to one with a 20% duplication threshold. It should take effect on the next push to this branch (I attempted a manual re-run but it doesn't want to listen to me)

(the duplicated lines are from database.py constructors so wouldn't be affected by .sql file changes)

integrations/acquisition/covid_hosp/state_timeseries/test_scenarios.py
rzats: Actually this should be passing because of a fix I've added later (lines 40-41) - I'll uncomment this and see how it goes

src/ddl/migrations/covid_hosp_state_split_tables.sql
rzats: Nope, did a quick test and this ran with no warnings:

```
mysql> ALTER TABLE `covid_hosp_state_daily` DROP COLUMN record_type;
Query OK, 0 rows affected (0.03 sec)
Records: 0  Duplicates: 0  Warnings: 0

mysql> ALTER TABLE `covid_hosp_state_timeseries` DROP COLUMN record_type;
Query OK, 0 rows affected (0.03 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

src/server/endpoints/covid_hosp_state_timeseries.py
rzats:
```
mysql> EXPLAIN SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` UNION ALL SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries`;
+----+-------------+-----------------------------+------------+------+---------------+------+---------+------+------+----------+-------+
| id | select_type | table                       | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra |
+----+-------------+-----------------------------+------------+------+---------------+------+---------+------+------+----------+-------+
|  1 | PRIMARY     | covid_hosp_state_daily      | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    3 |   100.00 | NULL  |
|  2 | UNION       | covid_hosp_state_timeseries | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    1 |   100.00 | NULL  |
+----+-------------+-----------------------------+------------+------+---------------+------+---------+------+------+----------+-------+
2 rows in set, 1 warning (0.01 sec)
```
```
mysql> EXPLAIN ANALYZE SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` UNION ALL SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries`;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| -> Append  (actual time=0.086..0.153 rows=4 loops=1)
    -> Stream results  (cost=0.55 rows=3) (actual time=0.086..0.100 rows=3 loops=1)
        -> Table scan on covid_hosp_state_daily  (cost=0.55 rows=3) (actual time=0.071..0.078 rows=3 loops=1)
    -> Stream results  (cost=0.35 rows=1) (actual time=0.049..0.050 rows=1 loops=1)
        -> Table scan on covid_hosp_state_timeseries  (cost=0.35 rows=1) (actual time=0.044..0.045 rows=1 loops=1)
 |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```
Doesn't look like there's a materialization step here.

src/ddl/covid_hosp.sql
rzats: Tried this approach out, it works fine in the sandbox but causes some issues when trying to re-run `acquisition` on the tables - https://github.com/cmu-delphi/delphi-epidata/actions/runs/4681152223/jobs/8293372868?pr=1126#step:16:84. I think it's fine to stick with two table definitions even if repetitive

melange396:
oof, looks like that force-push lost some changes 😢

rzats:
@melange396 that one's intended - I reverted a couple of changes because of https://github.com/cmu-delphi/delphi-epidata/pull/1126#discussion_r1164469299

melange396:
hmmm, [this](https://github.com/cmu-delphi/delphi-epidata/compare/d07299f111c4aab6ebec7c3e979136834461722d..caa796c4b4d32d6cff93a7050f2b3f3e47f1a88e) makes it look like you lost a change to `integrations/server/test_covid_hosp.py` and re-introduced a comment removed from `src/ddl/covid_hosp.sql` ¯\\\_(ツ)_/¯

src/ddl/covid_hosp.sql
melange396: thats concerning.  i think you just forgot a `ALTER TABLE covid_hosp_state_daily MODIFY id INT NOT NULL AUTO_INCREMENT;`

while it is fine to have separate and full table definitions in this file (in fact, it eventually will be desired (if not required!) when this file becomes auto-generated), we should endeavor to use the shorthand in the migration file.  there it serves as insurance against transcription/copypasta errors, while making the intent clearer and the entirety of the process more understandable.

integrations/server/test_covid_hosp.py
```diff
@@ -29,10 +29,16 @@ def setUp(self):
         cur.execute('truncate table covid_hosp_meta')
 
 
-  def insert_issue(self, cur, issue, value, record_type):
+  def insert_timeseries(self, cur, issue, value):
     so_many_nulls = ', '.join(['null'] * 57)
     cur.execute(f'''insert into covid_hosp_state_timeseries values (
-      0, {issue}, 'PA', 20201118, {value}, {so_many_nulls}, '{record_type}', {so_many_nulls}
+      0, {issue}, 'PA', 20201118, {value}, {so_many_nulls}, {so_many_nulls}
```
melange396, line 36: if you change `57` to `114` two lines above, you can make this insert template clearer.  same goes for the near-identical method below.
```suggestion
      0, {issue}, 'PA', 20201118, {value}, {so_many_nulls}
```

src/acquisition/covid_hosp/state_timeseries/database.py
```diff
@@ -225,5 +225,4 @@ def __init__(self, *args, **kwargs):
         table_name=Database.TABLE_NAME,
         hhs_dataset_id=Network.DATASET_ID,
         columns_and_types=Database.ORDERED_CSV_COLUMNS,
-        key_columns=Database.KEY_COLS,
-        additional_fields=[Columndef('T', 'record_type', None)])
```
melange396, line 229: please also remove `additional_fields` from the rest of `src/acquisition/covid_hosp/*`, as nothing else uses it now and it will only serve to add complexity and confusion.

rzats:
@melange396 good point, re-added those two changes! (The tests actually pass just fine without the `truncate table` call, since this is the only test that uses these tables, but it's probably nice to have just in case.)

melange396 (changes requested):
a few more points...

src/server/endpoints/covid_hosp_state_timeseries.py
rzats: Here's a query plan for that:
```
mysql> EXPLAIN ANALYZE WITH c AS (SELECT * , ROW_NUMBER() OVER (PARTITION BY date, state, issue ORDER BY record_type) `row` FROM (SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` UNION ALL SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries`) c) SELECT * FROM c;
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| -> Table scan on c  (cost=2.50..2.50 rows=0) (actual time=0.233..0.238 rows=4 loops=1)
    -> Materialize CTE c  (cost=0.00..0.00 rows=0) (actual time=0.231..0.231 rows=4 loops=1)
        -> Window aggregate: row_number() OVER (PARTITION BY c.`date`,c.state,c.issue ORDER BY c.record_type )   (actual time=0.185..0.199 rows=4 loops=1)
            -> Sort: c.`date`, c.state, c.issue, c.record_type  (cost=5.05..5.05 rows=4) (actual time=0.176..0.179 rows=4 loops=1)
                -> Table scan on c  (cost=1.94..3.85 rows=4) (actual time=0.154..0.159 rows=4 loops=1)
                    -> Union all materialize  (cost=1.30..1.30 rows=4) (actual time=0.152..0.152 rows=4 loops=1)
                        -> Table scan on covid_hosp_state_daily  (cost=0.55 rows=3) (actual time=0.069..0.076 rows=3 loops=1)
                        -> Table scan on covid_hosp_state_timeseries  (cost=0.35 rows=1) (actual time=0.044..0.045 rows=1 loops=1)
 |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```
And an alternate version of that query which puts the `UNION ALL` in the `WITH` clause:
```
mysql> EXPLAIN ANALYZE WITH merge_tables AS (SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` UNION ALL SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries`), c AS (SELECT * , ROW_NUMBER() OVER (PARTITION BY date, state, issue ORDER BY record_type) `row` FROM merge_tables) SELECT * FROM c;
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| -> Table scan on c  (cost=2.50..2.50 rows=0) (actual time=0.232..0.237 rows=4 loops=1)
    -> Materialize CTE c  (cost=0.00..0.00 rows=0) (actual time=0.230..0.230 rows=4 loops=1)
        -> Window aggregate: row_number() OVER (PARTITION BY merge_tables.`date`,merge_tables.state,merge_tables.issue ORDER BY merge_tables.record_type )   (actual time=0.183..0.198 rows=4 loops=1)
            -> Sort: merge_tables.`date`, merge_tables.state, merge_tables.issue, merge_tables.record_type  (cost=5.05..5.05 rows=4) (actual time=0.176..0.178 rows=4 loops=1)
                -> Table scan on merge_tables  (cost=1.94..3.85 rows=4) (actual time=0.155..0.160 rows=4 loops=1)
                    -> Materialize union CTE merge_tables  (cost=1.30..1.30 rows=4) (actual time=0.153..0.153 rows=4 loops=1)
                        -> Table scan on covid_hosp_state_daily  (cost=0.55 rows=3) (actual time=0.069..0.077 rows=3 loops=1)
                        -> Table scan on covid_hosp_state_timeseries  (cost=0.35 rows=1) (actual time=0.044..0.045 rows=1 loops=1)
 |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

rzats:
@melange396 applied the changes from this round as well!

src/ddl/migrations/covid_hosp_state_split_tables.sql
```diff
@@ -130,19 +9,16 @@ CREATE TABLE `covid_hosp_state_daily` (
   KEY `date_by_issue_and_state` (`issue`, `state`, `date`),
   -- for fast lookup of all states for a given date and issue
   KEY `state_by_issue_and_date` (`issue`, `date`, `state`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8
+SELECT * FROM covid_hosp_state_timeseries;
+ALTER TABLE covid_hosp_state_daily MODIFY id INT NOT NULL AUTO_INCREMENT;
```
melange396, line 14: its worth a comment here explaining why we are executing this statement.  same in the definition file.

src/server/endpoints/covid_hosp_state_timeseries.py
melange396: yuck, that looks like its gonna be pretty bad.  plus that EXPLAIN is for just one of three query forms that are used. 
 this is going to need some work, like distributing/factoring conditions across or into subqueries.  honestly, the existing query forms seem like they are more complicated than they need to be, so perhaps rewriting them is the way to go.

melange396 (changes requested):

melange396:
this looks pretty good, though i havent given it a very deep inspection yet.  in the meantime, you should include explain plans and timing for some "mocked" sample queries against real data (where we can use `covid_hosp_state_timeseries` twice, in place of the new table).

rzats:
(rebasing against latest `dev` branch, no other changes)

src/ddl/migrations/covid_hosp_state_split_tables.sql
```diff
@@ -0,0 +1,24 @@
...
+  -- for fast lookup of a time-series for a given state and issue
+  KEY `date_by_issue_and_state` (`issue`, `state`, `date`),
+  -- for fast lookup of all states for a given date and issue
+  KEY `state_by_issue_and_date` (`issue`, `date`, `state`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8
+SELECT * FROM covid_hosp_state_timeseries WHERE record_type='D';
+-- Re-add autoincrement capability
+ALTER TABLE covid_hosp_state_daily MODIFY id INT NOT NULL AUTO_INCREMENT;
+
+-- 2. Remove data with incorrect record_type from both tables (keep D in daily, T in timeseries)
```
krivard, line 17:
```suggestion
-- 2. Remove data with incorrect record_type from timeseries table (D records were moved to daily)
```

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,50 @@ def handle():
...
+                FROM {union_subquery}
+            )
+            SELECT {q.fields_clause} FROM {q.alias} WHERE `row` = 1 ORDER BY {q.order_clause}
+        '''
     else:
-        # final query using most recent issues
-        subquery = f"(SELECT max(`issue`) `max_issue`, `date`, `state` FROM {q.table} WHERE {q.conditions_clause} GROUP BY `date`, `state`) x"
-        condition = f"x.`max_issue` = {q.alias}.`issue` AND x.`date` = {q.alias}.`date` AND x.`state` = {q.alias}.`state`"
-        query = f"WITH c as (SELECT {q.fields_clause}, ROW_NUMBER() OVER (PARTITION BY date, state, issue ORDER BY record_type) `row` FROM {q.table} JOIN {subquery} ON {condition}) select {q.fields_clause} FROM {q.alias} WHERE `row` = 1 ORDER BY {q.order_clause}"
+        if as_of is not None:
```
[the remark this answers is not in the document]
rzats: If you take a close look at the queries, the `issues` branch partition like so, and take the first row from each grouping:
```
PARTITION BY date, state, issue ORDER BY record_type
```
This is a way to prioritize data from the daily table (where `record_type` is `D`) over timeseries (where `record_type` is `T`).

However, issue queries and `as_of` queries partition the data like so:
```
PARTITION BY date, state ORDER BY issue DESC, record_type
```
This means that the latest issues are prioritized first, and only then `record_type` - the only difference between the issue and `as_of` is an extra filter in the `as_of` subquery.

melange396 (changes requested):
this is lookin great!  just a few suggestions for clarity...

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,50 @@ def handle():
     q.where_integers("date", dates)
     q.where_strings("state", states)
 
+    # These queries prioritize the daily value if there is both a time series and daily value for a given issue/date/state.
+    # Further details: https://github.com/cmu-delphi/delphi-epidata/pull/336
     if issues is not None:
```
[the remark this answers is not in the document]
krivard: more specifically -- do we always want to treat falsy values of `issues` as if `issues` was `None`? 0, False, [], "", {}, etc

we don't encourage users to submit 0 as an issue, but that doesn't mean they couldn't do it

rzats:
Requesting re-review after _many_ iterations on the `covid_hosp_state_timeseries` query :)

The testing results of this query compared to the current one in `dev` can be found in [this spreadsheet](https://docs.google.com/spreadsheets/d/1evI52DVPN-830_C-wQPjPjFcKh-9A4tsSDxRbTz1QsE/edit#gid=0) (tl;dr this outperforms `dev` significantly, in a certain branch by **>300%**). The optimizations made were:
1. switching the "get greatest value per group" logic to a more well-performing approach (`PARTITION` -> `INNER JOIN`)
2. incidentally fixing an error which may have caused the query to ignore the [index intended for it](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/ddl/covid_hosp.sql#L504) (that index has the columns ordered as `state -> date -> issue`, but the query was previously grouping by `date -> state -> issue`)

src/ddl/covid_hosp.sql
```diff
@@ -500,15 +488,34 @@ CREATE TABLE `covid_hosp_state_timeseries` (
...
+  PRIMARY KEY (`id`),
+  -- for fast lookup of most recent issue for a given state and date
+  UNIQUE KEY `issue_by_state_and_date` (`state`, `date`, `issue`),
+  -- for fast lookup of a time-series for a given state and issue
+  KEY `date_by_issue_and_state` (`issue`, `state`, `date`),
+  -- for fast lookup of all states for a given date and issue
+  KEY `state_by_issue_and_date` (`issue`, `date`, `state`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8
+SELECT * FROM covid_hosp_state_timeseries;
+-- Re-add autoincrement capability
```
melange396, line 516: to match the migration file:
```suggestion
-- AUTOINCREMENT is not preserved by `CREATE TABLE ... SELECT`; Re-add
```

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,65 @@ def handle():
...
+        if as_of is not None:
+            # ...Filter for most recent issues before a given as_of
+            cond_clause += " AND (issue <= :as_of)"
+            q.params["as_of"] = as_of
+        join_condition = f"{q.alias}.state = x.state AND {q.alias}.date = x.date AND {q.alias}.issue = x.max_issue"
+
+        # Get the rows from the daily & timeseries tables with the highest issue value within each state/date group
+        join_daily = f'''
+            SELECT {q.fields_clause}, 'D' AS record_type FROM `covid_hosp_state_daily` {q.alias}
+            INNER JOIN (
```
melange396, line 188: nit: a plain `JOIN` is identical to `INNER JOIN`, and i think it seems to be a more common usage.
```suggestion
           JOIN (
```

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,65 @@ def handle():
...
-        sub_condition_asof = "(issue <= :as_of)"
-        q.params["as_of"] = as_of
-        query = f"WITH c as (SELECT {q.fields_clause}, ROW_NUMBER() OVER (PARTITION BY date, state ORDER BY issue DESC, record_type) `row` FROM {q.table} WHERE {q.conditions_clause} AND {sub_condition_asof}) SELECT {q.fields_clause} FROM {q.alias} WHERE `row` = 1 ORDER BY {q.order_clause}"
+
+        # Get all issues matching the conditions from daily & timeseries
+        union_subquery = f'''
+        (
+            SELECT *, 'D' as record_type FROM `covid_hosp_state_daily` {q.alias} WHERE {q.conditions_clause}
+            UNION ALL
+            SELECT *, 'T' as record_type FROM `covid_hosp_state_timeseries` {q.alias} WHERE {q.conditions_clause}
```
melange396, line 166: nit: capitalization for consistency
```suggestion
            SELECT *, 'D' AS record_type FROM `covid_hosp_state_daily` {q.alias} WHERE {q.conditions_clause}
            UNION ALL
            SELECT *, 'T' AS record_type FROM `covid_hosp_state_timeseries` {q.alias} WHERE {q.conditions_clause}
```

another nit: `AS` is missing (because it is optional, after all) from many of the row and table alias specifications scattered around this file, but i think it reads much better when it is used.  you did already include it with `max_issue`.

melange396 (changes requested):
his is awesome.  its much easier to understand (IMNSHO), even though we moved away from the one overloaded table and into two (getting rid of the `WITH` expressions is a big help).  plus the performance increases speak for themselves....

all comments marked 'nit:' are elective, you can decide whether you want to do them or not.

src/server/endpoints/covid_hosp_state_timeseries.py
```diff
@@ -152,19 +152,65 @@ def handle():
...
+        ) c'''
```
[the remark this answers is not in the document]
rzats: I've gone for `{q.alias}` just in case the name for it ever gets changed or the `QueryBuilder` is otherwise refactored.

rzats:
@melange396 applied the fixes!

melange396 (approved):
awesome!

melange396:
we probbly want to hold off on merging just yet, and first figure out whether to do this as a migration or as a re-load of all the data from the external sources

PR #1133, opened 2023-04-18
POC  converting acquisition/ code to ORM

BrainIsDead:
closes|addresses  https://cmu-delphi.atlassian.net/browse/EPID-42
### Summary:

FluSurv model and flusurv_update refactor

### Prerequisites:

- [ ] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [ ] Build is successful

### DIFF
--- a/src/acquisition/flusurv/flusurv_update.py
+++ b/src/acquisition/flusurv/flusurv_update.py
@@ -68,122 +68,112 @@ rate_age_7: hospitalization rate for ages 85+
   + initial version
 """
 
-# standard library
 import argparse
 
-# third party
-import mysql.connector
-
-# first party
 from delphi.epidata.acquisition.flusurv import flusurv
-import delphi.operations.secrets as secrets
 from delphi.utils.epidate import EpiDate
 from delphi.utils.epiweek import delta_epiweeks
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+from ...server._config import (
+    SQLALCHEMY_DATABASE_URI,
+    SQLALCHEMY_ENGINE_OPTIONS
+)
+from .models import FluSurv
+
+engine = create_engine(SQLALCHEMY_DATABASE_URI, **SQLALCHEMY_ENGINE_OPTIONS)
 
 
-def get_rows(cur):
-  """Return the number of rows in the `flusurv` table."""
-
-  # count all rows
-  cur.execute('SELECT count(1) `num` FROM `flusurv`')
-  for (num,) in cur:
-    return num
+def get_rows(session):
+    """Return the number of rows in the `flusurv` table."""
+    return session.query(FluSurv.id).count()
 
 
 def update(issue, location_name, test_mode=False):
-  """Fetch and store the currently avialble weekly FluSurv dataset."""
+    """Fetch and store the currently avialble weekly FluSurv dataset."""
 
-  # fetch data
-  location_code = flusurv.location_codes[location_name]
-  print('fetching data for', location_name, location_code)
-  data = flusurv.get_data(location_code)
+    # fetch data
+    location_code = flusurv.location_codes[location_name]
+    print('fetching data for', location_name, location_code)
+    data = flusurv.get_data(location_code)
 
-  # metadata
-  epiweeks = sorted(data.keys())
-  location = location_name
-  release_date = str(EpiDate.today())
+    # metadata
+    epiweeks = sorted(data.keys())
+    location = location_name
+    release_date = str(EpiDate.today())
 
-  # connect to the database
-  u, p = secrets.db.epi
-  cnx = mysql.connector.connect(
-      host=secrets.db.host, user=u, password=p, database='epidata')
-  cur = cnx.cursor()
-  rows1 = get_rows(cur)
-  print('rows before: %d' % rows1)
+    with Session(engine) as session:
+        rows1 = get_rows(session)
+        print('rows before: %d' % rows1)
+        for epiweek in epiweeks:
+            lag = delta_epiweeks(epiweek, issue)
+            if lag > 52:
+                # Ignore values older than one year, as (1) they are assumed not to
+                # change, and (2) it would adversely affect database performance if all
+                # values (including duplicates) were stored on each run.
+                continue
+            args_meta = {
+                "issue": issue,
+                "epiweek": epiweek,
+                "location": location,
+            }
+            args_update = {
+                "release_date": release_date,
+                "lag": lag,
+                "rate_age_0": data[epiweek][0],
+                "rate_age_1": data[epiweek][1],
+                "rate_age_2": data[epiweek][2],
+                "rate_age_3": data[epiweek][3],
+                "rate_age_4": data[epiweek][4],
+                "rate_overall": data[epiweek][5],
+                "rate_age_5": data[epiweek][6],
+                "rate_age_6": data[epiweek][7],
+                "rate_age_7": data[epiweek][8],
+            }
+            existing_flusurv = session.query(FluSurv).filter_by(**args_meta)
+            if existing_flusurv.first() is not None:
+                existing_flusurv.update(args_update)
+            else:
+                args_create = args_meta
+                args_create.update(args_update)
+                session.add(FluSurv(**args_create))
 
-  # SQL for insert/update
-  sql = '''
-  INSERT INTO `flusurv` (
-    `release_date`, `issue`, `epiweek`, `location`, `lag`, `rate_age_0`,
-    `rate_age_1`, `rate_age_2`, `rate_age_3`, `rate_age_4`, `rate_overall`,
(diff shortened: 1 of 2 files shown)


src/acquisition/flusurv/models.py
```diff
@@ -0,0 +1,29 @@
+from sqlalchemy import Column
+from sqlalchemy.ext.declarative import declarative_base
```
krivard, line 2: question: what's the difference between `sqlalchemy.ext.declarative.declarative_base` and `sqlalchemy.orm.declarative_base`?

[The 1.4 documentation](https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html) uses the `.orm` name. Is `.ext` maybe for sqlalchemy 1.3? [Our requirements.txt lists 1.4.40.](https://github.com/cmu-delphi/delphi-epidata/blob/4dc5f129704337a83bbd1ecd3bf2ad573d1e1e5f/requirements.api.txt#L13)

src/acquisition/flusurv/models.py
```diff
@@ -0,0 +1,29 @@
...
+from sqlalchemy.types import Date, Float, Integer, String
+
+Base = declarative_base()
+
+
+class FluSurv(Base):
```
krivard, line 8: question: does this depend on .sql ddl files or can it replace them?

[src/ddl/fluview.sql](https://github.com/cmu-delphi/delphi-epidata/blob/EPID-42/src/ddl/fluview.sql) is used in two scenarios:

1. (common) setting up the database for the local development environment in Docker
2. (uncommon) setting up a fresh production/staging/QA environment if something has gone Horribly Wrong with one of the existing ones

in both scenarios we want the resulting (empty) database to match what's in (full) prod, with the exception of column order.

with this change, we now have parallel table definitions in fluview.sql and this file. can we drop the .sql definition? if so, what's the procedure for spinning up an empty table based on the sqlalchemy definition?

src/acquisition/flusurv/models.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    SQLAlchemy model representing flusurve data.
+    """
+
+    __tablename__ = 'flusurv'
```
krivard, line 13: suggestion: include constraints and indexes under their existing names

```suggestion
    __tablename__ = 'flusurv'
    __table_args__ = (
        UniqueConstraint("issue", "epiweek", "location", name="issue"),
        Index("release_date", "release_date"),
        Index("issue_2", "issue"),
        Index("epiweek", "epiweek"),
        Index("region", "location"),
        Index("lag", "lag"),
    )
```

From https://github.com/cmu-delphi/delphi-epidata/blob/4dc5f129704337a83bbd1ecd3bf2ad573d1e1e5f/src/ddl/fluview.sql#L386-L391

& referencing 
* https://docs.sqlalchemy.org/en/14/core/constraints.html#unique-constraint
* https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#declarative-table-configuration

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -68,122 +68,112 @@
...
+                "lag": lag,
+                "rate_age_0": data[epiweek][0],
+                "rate_age_1": data[epiweek][1],
+                "rate_age_2": data[epiweek][2],
+                "rate_age_3": data[epiweek][3],
+                "rate_age_4": data[epiweek][4],
+                "rate_overall": data[epiweek][5],
+                "rate_age_5": data[epiweek][6],
+                "rate_age_6": data[epiweek][7],
+                "rate_age_7": data[epiweek][8],
```
krivard, line 132: question: is it worth refactoring `flusurv.extract_from_object` to produce a dict with `rate_*` keys instead of an array?

then including them here would be more straightforward

krivard (changes requested):
* 1 mandatory fix (uniqueness)
* 1 question to resolve before merge (`least` vs `update`)
* a bunch of optional stuff

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -68,122 +68,112 @@
...
+            }
+            existing_flusurv = session.query(FluSurv).filter_by(**args_meta)
+            if existing_flusurv.first() is not None:
+                existing_flusurv.update(args_update)
```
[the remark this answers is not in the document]
BrainIsDead: new version filter flusurvs and if  flusurv wit same issue, epiweek and location exists  updates it, else it will create new one

src/acquisition/flusurv/models.py
BrainIsDead: basically, they are the same but you are right ist more logical to use sqlalchemy.orm.declarative_base

src/acquisition/flusurv/models.py
```diff
@@ -0,0 +1,29 @@
...
+
+    id = Column(Integer, primary_key=True, autoincrement="auto", nullable=False)
+    release_date = Column(Date, nullable=False)
+    issue = Column(Integer, unique=True, nullable=False)
+    epiweek = Column(Integer, unique=True, nullable=False)
+    location = Column(String(length=32), unique=True, nullable=False)
```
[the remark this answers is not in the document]
BrainIsDead: if  it is analog unique_together you are right

src/acquisition/flusurv/flusurv_update.py
BrainIsDead: I thought about it's complex as hell almost unreadabble  mostly cause of complex response from flusurv_url

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -68,122 +68,112 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+from ...server._config import (
+    SQLALCHEMY_DATABASE_URI,
+    SQLALCHEMY_ENGINE_OPTIONS
+)
+from .models import FluSurv
 
+engine = create_engine(SQLALCHEMY_DATABASE_URI, **SQLALCHEMY_ENGINE_OPTIONS)
```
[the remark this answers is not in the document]
BrainIsDead: chat GPT recommends to do something like this 
```python
from sqlalchemy import create_engine

def get_engine():
    engine = create_engine('postgresql://user:password@localhost/mydatabase')
    return engine

with get_engine().connect() as conn:
    result = conn.execute("SELECT * FROM mytable")
    for row in result:
        print(row)
```
instead of global variables

but my suggestion is to make even deeper
```python
db = SQLAlchemy(metadata=meta)
ma = Marshmallow()
bp = Blueprint('bp', __name__, url_prefix='/api')
migrate = Migrate(db)


def create_app(config_object=Config):
    ''' Flask Application Factory

    Docs:
        https://flask.palletsprojects.com/en/2.0.x/patterns/appfactories/
    '''

    flask_app = Flask(__name__)
    flask_app.config.from_object(config_object())

    # init db and serializer
    db.init_app(flask_app)
    ma.init_app(flask_app)
    migrate.init_app(flask_app, db)

    # register app blueprints
    import app.event_impact.blueprints  # noqa

    # register blueprints
    flask_app.register_blueprint(bp)

    return flask_app
```

but it will affect a lot

src/acquisition/flusurv/models.py
BrainIsDead: I think it is fully independent

src/acquisition/flusurv/models.py
BrainIsDead: Done

src/acquisition/flusurv/models.py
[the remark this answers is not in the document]
BrainIsDead: Done

src/acquisition/flusurv/flusurv_update.py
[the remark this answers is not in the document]
krivard: checking my understanding -- you're recommending we use Flask on the acquisition side as well? We currently only use it for the API server.

src/acquisition/flusurv/flusurv_update.py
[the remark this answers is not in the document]
krivard: we want this behavior:

| (issue, epiweek, location) exists? | column | row action | column value | condition |
| - | - | - | - | - |
| no | * | insert | new value | * |
| yes | release date | update | old release date | old release date < new release date |
| yes | release date | update | new release date | new release date < old release date |
| yes | everything else | update | new value | * |

src/acquisition/flusurv/models.py
krivard: ideally we only have one copy of each table definition. if we wanted to drop the .sql file, what would be the procedure for spinning up an empty table based on the sqlalchemy definition?

krivard (commented):
whoops, submitted them one at a time. regardless, more questions

dshemetov:
Closing as we are migrating away from this repo and would prefer to keep it frozen at this time.

PR #1138, opened 2023-04-19
update delphi_utils version

melange396:
i assumed this was not version-pinned and would use whatever was most up to date.  i shoulda checked!  fwiw, we were pinned to a version from august 2022.

this should pick up the change from https://github.com/cmu-delphi/covidcast-indicators/pull/1835


nolangormley (approved):
LGTM

PR #1154, opened 2023-04-27
Sync to www-covidcast release v10.0.0

rzats:
Syncing covidcast_utils files to www-covidcast release v10.0.0.

### DIFF
--- a/src/server/endpoints/covidcast_utils/descriptions.raw.txt
+++ b/src/server/endpoints/covidcast_utils/descriptions.raw.txt
@@ -1,0 +1,125 @@
+---
+Name: Symptom Searches (Smell and Taste) on Google 
+Id: google-symptoms
+Signal: s05_smoothed_search
+Unit: scaled search volume
+UnitShort: 
+noMaps: true
+ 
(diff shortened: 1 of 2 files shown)


rzats:
This is an automated test PR for https://github.com/cmu-delphi/www-covidcast/pull/1216! We may or may not want to merge this - @melange396?

melange396:
it doesnt matter too much either way.  we dont need to because theyre already present in #1155 where they are actually used.  on the other hand, we could merge them in now if you want to try to see what happens when the workflow is run but the files have not changed.

krivard (commented):
what is the v.10.0.0 referring to? [www-covidcast is currently in v3.2.5](https://github.com/cmu-delphi/www-covidcast/releases)

rzats:
@krivard this "release version" is arbitrarily picked, since this is just a fully automated test PR. I can rename this PR/commit so there's no confusion about a 10.0.0 actually existing :)

krivard (commented):
we should probably close this without merging; if we merge it then #1155 will get confused about trying to add the same files

PR #1164, opened 2023-05-15
replace references to `delphi.cmu.edu/epidata` with `api.delphi.cmu.edu/epidata`

melange396:
- `delphi.cmu.edu/epidata` proxies to `api.covidcast.cmu.edu/epidata` but this includes an "extra hop" which presents a number of problems:
    - it adds a single point of failure.
    - the resource is not load-balanced but proxies to our load-balanced servers, thus defeating some of the other purposes of load-balancing.
    - complicates logging as the webapp logs will need to be cross referenced with proxy logs to get the full spectrum of information on usage.
    - aggregates any number of users' activity behind one IP address (the address of the proxy) for the purposes of rate limiting.

  - this change should happen in our code (especially client libs) and documentation.  outside of this repository, this includes but is not limited to:
    - https://github.com/cmu-delphi/epidatr/blob/02a688219c0deba8a451de2aa1877ee33a8d5672/R/constants.R#L3
    - https://github.com/cmu-delphi/epidatpy/blob/8659609c960cf1863f7c200ee6da9aa876a33b18/epidatpy/_constants.py#L9
    - likely more in: https://github.com/search?q=org%3Acmu-delphi+%22delphi.cmu.edu%22+language%3AR+OR+language%3Apython+OR+language%3AJavaScript+&type=code
    - ( https://github.com/cmu-delphi/covidcast seems to be ok )
    - ( documentation in https://cmu-delphi.github.io/delphi-epidata/api/README.html#the-api should be changed with this PR )

  - usage of this proxied url is not very high at the moment.  however, we are about to announce our "API Keys" release, which involves changes that will require many users to alter their interactions with our service.  they may choose to use our client libraries or reference our documentation, and it is advantageous for us if they do not increase activity to this proxied URI -- hence the desire to update our references to point to the non-proxied URI before such an announcement.  additionally, a significant portion of our current users still make use of very old versions of client libraries, so it is also important that we update our references sooner rather than later, to prevent users from becoming "cemented" to a newer library that still references the proxied URI.

  - we may want to consider using a hostname other than `api.covidcast.cmu.edu`, perhaps one that is not covid-specific
    - this change can happen in the future, and can be done with DNS aliases (CNAMES or similar)

  - ultimately, once activity to `delphi.cmu.edu/epidata` stops (or drops to a negligible amount), the proxying rule can be disabled or removed.
    - it looks like the config for this may be located here: https://github.com/cmu-delphi/delphi-admin/blob/5b7067c5343072bf4fdebc11d1ffdc00fdfc2644/ansible/templates/web-frontend-prod/web-frontend-main-p.conf.j2#L188-L190

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -70,7 +70,7 @@ Please note that our `endpoint` parameters were previously referenced as `source
 
 # The API
 
-The base URL is: https://delphi.cmu.edu/epidata/
+The base URL is: https://api.covidcast.cmu.edu/epidata/
 
 ## Specifying Epiweeks, Dates, and Lists
 
@@ -154,7 +154,7 @@ The parameters available for each source are documented in each linked source-sp
 # Example URLs
 
 ### FluView on 2015w01 (national)
-https://delphi.cmu.edu/epidata/fluview/?regions=nat&epiweeks=201501
+https://api.covidcast.cmu.edu/epidata/fluview/?regions=nat&epiweeks=201501
 
 ```json
 {
@@ -184,7 +184,7 @@ https://delphi.cmu.edu/epidata/fluview/?regions=nat&epiweeks=201501
 ```
 
 ### Wikipedia Access article "influenza" on 2020w01
-https://delphi.cmu.edu/epidata/wiki/?language=en&articles=influenza&epiweeks=202001
+https://api.covidcast.cmu.edu/epidata/wiki/?language=en&articles=influenza&epiweeks=202001
 
 ```json
 {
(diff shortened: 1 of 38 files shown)


dshemetov (approved):
These particular changes look good to me.

`api.covidcast.cmu.edu` seems fine, especially since we can add a new non-covid domain later using CNAMES.

dshemetov (dismissed):

krivard (approved):
👍

korlaxxalrok (approved):
:shipit: 💥 👍

melange396:
good catch on the new docs!!!

PR #1172, opened 2023-05-22
Make HTTP status 4xx responses uniform

krivard:
### Summary:

We were returning JSON for 429s, and HTML for 401s -- this PR has us return HTML for both.

I could've gone either way, but HTML is what delphi-epidata.R expects, and what I've been telling Dmitry, Logan, and Alex, so it won out

### Prerequisites:

- [x] It should be merged against the `api-keys` branch
- [x] Branch is up-to-date with the branch to be merged with
- [x] Build is successful

### DIFF
--- a/src/server/_config.py
+++ b/src/server/_config.py
@@ -57,7 +57,7 @@ REGION_TO_STATE = {
 }
 NATION_REGION = "nat"
 
-API_KEY_REQUIRED_STARTING_AT = date.fromisoformat(os.environ.get("API_REQUIRED_STARTING_AT", "2023-06-21"))
+API_KEY_REQUIRED_STARTING_AT = date.fromisoformat(os.environ.get("API_KEY_REQUIRED_STARTING_AT", "2023-06-21"))
 TEMPORARY_API_KEY = os.environ.get("TEMPORARY_API_KEY", "TEMP-API-KEY-EXPIRES-2023-06-28")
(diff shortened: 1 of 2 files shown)


src/server/_config.py
```diff
@@ -57,7 +57,7 @@
 }
 NATION_REGION = "nat"
 
-API_KEY_REQUIRED_STARTING_AT = date.fromisoformat(os.environ.get("API_REQUIRED_STARTING_AT", "2023-06-21"))
+API_KEY_REQUIRED_STARTING_AT = date.fromisoformat(os.environ.get("API_KEY_REQUIRED_STARTING_AT", "2023-06-21"))
```
melange396, line 60: oof, nice find.  that could have ended up being fairly confusing.  we should also remove or update the line in [.env.example](https://github.com/cmu-delphi/delphi-epidata/blob/1480948c1b253ba7d5733ea5618a4351924a87a8/.env.example#L4), as well as change it in [cmu-delphi/delphi-ansible-web](https://github.com/cmu-delphi/delphi-ansible-web/blob/2ed0f71123ac223e23bf2aae521e9c7710561649/vars.yml#L118)

melange396 (changes requested):
good call!  consistency is key.

i have a note for stuff related to that config variable name change as well

krivard:
Don't merge until staging tests are clear

melange396 (approved):

PR #1186, opened 2023-05-27
style(black): format acquisition with `black`, line-length=200

dshemetov:
Run the acquisition Python files through the code formatter black.

### Summary:

- Everything but covidcast (do that one after #1103).
- Also makes SonarCode happy.

### Prerequisites:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

PR #1189, opened 2023-06-05
style(black): format acquisition with `black`, line-length=100

dshemetov:
A version of #1186 with black's `line-length=100`. Recommend using GitHub's ignore whitespace feature when looking at diffs.

TL;DR: it mostly looks good. The main jank is, as Katie put it below, "repeated calls should use the same wrap mode".

- [x] Add .pylintrc configs from [covidcast-indicators](https://github.com/cmu-delphi/covidcast-indicators/blob/main/hhs_hosp/.pylintrc). Don't enforce pylint, just keep the same configs.
- [x] Apply [pyupgrade](https://github.com/asottile/pyupgrade) for a few minor changes.
- [x] Use [flynt](https://github.com/ikamensh/flynt) to automatically convert many of the % and .format strings into official f-strings. It was pretty straightforward `pip install flynt` and `flynt src/acquisition -a -l 999 --verbose` (-a is aggressive, it won't change % strings otherwise, -l 999 is line-length, it won't change long strings otherwise, --verbose is to have it log its changes). With this, it had a 94/98 success rate on %-style expressions, so I had to go fix a few manually.
- [x] Write the correct commits into ignore blame refs file, once this PR is finalized


src/acquisition/afhsb/afhsb_csv.py
```diff
@@ -19,174 +19,186 @@
...
-		for row in reader:
-			print(row.keys())
-			alpha2 = row['alpha-2']
-			alpha3 = row['alpha-3']
-			mapping[alpha2] = alpha3
-
-	return mapping
+    filename = "country_codes.csv"
+    mapping = dict()
+    with open(os.path.join(TARGET_DIR, filename)) as csvfile:
```
dshemetov, line 127: FWIW, I also used [`pyupgrade`](https://github.com/asottile/pyupgrade) here, which made a few changes like this

src/acquisition/afhsb/afhsb_csv.py
```diff
@@ -19,174 +19,186 @@
...
+                print(key, row[key])
+            continue
+        else:
+            country3 = country_mapping[country2]
+        new_row = {
+            "dmisid": row["DMIS ID"],
+            "country": country3,
+            "state": row["Facility State Code"],
+            "zip5": row["Facility 5-Digit ZIP Code"],
+        }
```
dshemetov, line 167: This looks better

src/acquisition/covid_hosp/facility/database.py
```diff
@@ -7,213 +7,467 @@
...
+        ),
+        Columndef(
+            "total_staffed_adult_icu_beds_7_day_coverage",
+            "total_staffed_adult_icu_beds_7_day_coverage",
+            int,
+        ),
+        Columndef(
+            "total_staffed_adult_icu_beds_7_day_sum", "total_staffed_adult_icu_beds_7_day_sum", int
+        ),
+        Columndef("zip", "zip", str),
```
dshemetov, line 462: The code wasn't consistent before, but even if it was, black would do this sort of inconsistent thing. Probably my least favorite part about it.

The thing black recommends here is to format these lines manually, then surround the whole block with fmt: off and fmt: on, so black doesn't touch it in the future.

src/acquisition/covid_hosp/state_daily/database.py
```diff
@@ -7,224 +7,502 @@
...
+        ),
+        Columndef(
+            "total_pediatric_patients_hospitalized_confirmed_covid_coverage",
+            "total_pediatric_patients_hosp_confirmed_covid_coverage",
+            int,
+        ),
+        Columndef("total_staffed_adult_icu_beds", "total_staffed_adult_icu_beds", int),
+        Columndef(
+            "total_staffed_adult_icu_beds_coverage", "total_staffed_adult_icu_beds_coverage", int
+        ),
```
dshemetov, line 496: same comment here as above

src/acquisition/covid_hosp/state_timeseries/database.py
```diff
@@ -7,223 +7,501 @@
...
+        ),
+        Columndef(
+            "total_pediatric_patients_hospitalized_confirmed_covid_coverage",
+            "total_pediatric_patients_hosp_confirmed_covid_coverage",
+            int,
+        ),
+        Columndef("total_staffed_adult_icu_beds", "total_staffed_adult_icu_beds", int),
+        Columndef(
+            "total_staffed_adult_icu_beds_coverage", "total_staffed_adult_icu_beds_coverage", int
+        ),
```
dshemetov, line 495: same comment as above

src/acquisition/ght/ght_update.py
```diff
@@ -88,304 +88,325 @@
...
+        action="store",
+        type=str,
+        default=None,
+        help='term/query/topic (ex: all; /m/0cycc; "flu fever")',
+    )
+    parser.add_argument("--first", "-f", default=None, type=int, help="first epiweek override")
+    parser.add_argument("--last", "-l", default=None, type=int, help="last epiweek override")
+    parser.add_argument(
+        "--country", "-c", default="US", type=str, help="location country (ex: US; BR)"
+    )
```
dshemetov, line 372: inconsistency here is annoying

src/acquisition/norostat/norostat_add_history.py
```diff
@@ -18,28 +18,42 @@
...
+    )
+    snapshot_version_counter = collections.Counter()
+    for subdir in os.listdir(snapshot_dir):
+        if re.match(r"[0-9]+", subdir) is not None:
+            # appears to be snapshot dir
+            snapshot_version_counter[subdir] = 0  # register that loop found this snapshot directory
+            for norostat_capitalization in ["norostat", "noroSTAT"]:
+                time.sleep(
+                    0.002
+                )  # ensure parse times are unique, assuming OS can accurately sleep and measure to ms precision
```
dshemetov, line 34: hm, i dont like that it's taking the inline comment into account for line wrapping

src/acquisition/covid_hosp/common/utils.py
```diff
@@ -8,219 +8,230 @@
...
+
+        dfs = [
+            df.set_index(key_cols) for df in dfs if not all(k in df.index.names for k in key_cols)
+        ]
+        result = dfs[0]
+        if logger and len(dfs) > 7:
+            logger.warning(
+                "expensive operation",
+                msg="concatenating more than 7 files may result in long running times",
+                count=len(dfs),
```
nmdefries, line 167: Wrapping here is better in this version

src/acquisition/fluview/fluview_update.py
```diff
@@ -130,398 +130,422 @@
...
+        "REGION",
+        "YEAR",
+        "WEEK",
+        "TOTAL SPECIMENS",
+        "TOTAL A",
+        "TOTAL B",
+        "PERCENT POSITIVE",
+        "PERCENT A",
+        "PERCENT B",
+    ]:
```
nmdefries, line 207: Better wrapping here

src/acquisition/ecdc/ecdc_db_update.py
```diff
@@ -138,39 +144,34 @@ def update_from_file(issue, date, dir, test_mode=False):
...
 
+
 def main():
     # args and usage
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        '--test',
-        action='store_true',
-        help='do dry run only, do not update the database'
+        "--test", action="store_true", help="do dry run only, do not update the database"
```
nmdefries, line 160: same with the vertical space comment

src/acquisition/fluview/fluview_update.py
```diff
@@ -130,398 +130,422 @@
...
+    for row in entries:
+        lag = delta_epiweeks(row["epiweek"], issue)
+        args = [
+            row["total_specimens"],
+            row["total_a"],
+            row["total_b"],
+            row["percent_positive"],
+            row["percent_a"],
+            row["percent_b"],
+        ]
```
nmdefries, line 355: Maybe not the best use of space but acceptable

src/acquisition/twtr/healthtweets.py
```diff
@@ -36,132 +36,221 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("username", action="store", type=str, help="healthtweets.org username")
+    parser.add_argument("password", action="store", type=str, help="healthtweets.org password")
+    parser.add_argument(
+        "state",
+        action="store",
+        type=str,
+        choices=list(HealthTweets.STATE_CODES.keys()),
+        help="U.S. state (ex: TX)",
+    )
```
nmdefries, line 228: very inconsistent

nmdefries (commented):
I like this one a lot better than with `line-length=200`. The formatting is acceptable overall and the wrapping style is more readable.

There are too many changes here to futz around with manually fixing and freezing more than a few individual lines, though.

I'd say the most worthwhile use of time would be deciding how we want to format the `Columndef("previous_day_admission_adult_covid_confirmed_20-29_7_day_sum", "previous_day_admission_adult_covid_confirmed_20_29_7_day_sum", int),` lines in all the `database.py` variants, and call it good after that.

dshemetov:
Absolutely agree with your review comment.

Formatting ColumnDef lines is actually moot, since we have [a refactor for these lines in the PR queue already](https://github.com/cmu-delphi/delphi-epidata/pull/1114).

So I'm just going to exclude covid_hosp from this PR, to avoid conflicts with that PR. We can format that code after those refactors.

src/acquisition/cdcp/cdc_dropbox_receiver.py
```diff
@@ -29,128 +29,128 @@
...
+    if len(save_list) == 0:
+        print("did not find any new data files")
+        return
+
+    # download new files, saving them inside of a new zip file
+    timestamp = get_timestamp_string()
+    zip_path = f"{DELPHI_BASE_DIR}/dropbox_{timestamp}.zip"
+    print(f"downloading into delphi:{zip_path}")
+    with ZipFile(zip_path, "w", ZIP_DEFLATED) as zf:
+        for name in save_list:
```
krivard, line 92: this is when I wish github let you provide manual alignment markers to make diffs make sense

src/acquisition/cdcp/cdc_extract.py
```diff
@@ -130,94 +130,99 @@ def store_result(cur, epiweek, state, num1, num2, num3, num4, num5, num6, num7,
...
+                num7 = get_num_hits(cur, epiweek, state, pages[6])
+                num8 = get_num_hits(cur, epiweek, state, pages[7])
+                total = get_total_hits(cur, epiweek, state)
+                store_result(
+                    cur, epiweek, state, num1, num2, num3, num4, num5, num6, num7, num8, total
+                )
+                print(
+                    " %d-%s: %d %d %d %d %d %d %d %d (%d)"
+                    % (epiweek, state, num1, num2, num3, num4, num5, num6, num7, num8, total)
+                )
```
krivard, line 191: black is worse here but the actual fix is to refactor to something like

```suggestion
                nums = []
                for i in range(8):
                    nums[i] = get_num_hits(cur, epiweek, state, pages[i])
                total = get_total_hits(cur, epiweek, state)
                store_result(cur, epiweek, state, *nums, total)
                print(f" {epiweek}-{state}: {' '.join(str(n) for n in nums)} ({total})")
```

src/acquisition/covid_hosp/common/database.py
```diff
@@ -15,263 +15,290 @@
...
+        for csv_name in self.key_columns:
+            dataframe.loc[:, csv_name] = dataframe[csv_name].map(
+                self.columns_and_types[csv_name].dtype
+            )
+
+        num_columns = 2 + len(dataframe_columns_and_types) + len(self.additional_fields)
+        value_placeholders = ", ".join(["%s"] * num_columns)
+        columns = ", ".join(
+            f"`{i.sql_name}`" for i in dataframe_columns_and_types + self.additional_fields
+        )
```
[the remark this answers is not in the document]
krivard: agreed, but covid_hosp should be left out of this PR until the logic refactor is done

src/acquisition/fluview/fluview_update.py
```diff
@@ -130,398 +130,422 @@
...
+    # args and usage
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--test", action="store_true", help="do dry run only, do not update the database"
+    )
+    parser.add_argument(
+        "--file", type=str, help="load an existing zip file (otherwise fetch current data)"
+    )
+    parser.add_argument(
+        "--issue", type=int, help="issue of the file (e.g. 201740); used iff --file is given"
```
krivard, line 526: hmmmm I'm skeptical about these. if it were just one, fine, but with three of them it's helpful to have whitespace highlight argument alignment

src/acquisition/afhsb/afhsb_csv.py
krivard: i like the pyupgrade changes

krivard (commented):
Places it falls over:

* repeated calls should use the same wrap mode
* same-line comments on too-long lines should be moved to the previous line before applying wrapping rules

otherwise seems pretty good!

src/acquisition/covid_hosp/common/database.py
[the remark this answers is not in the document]
dshemetov: yup, getting to it

src/acquisition/covid_hosp/common/database.py
```diff
@@ -15,263 +15,290 @@
...
+        sql = (
+            f"INSERT INTO `{self.table_name}` (`id`, `{self.publication_col_name}`, {columns}) "
+            f"VALUES ({value_placeholders})"
+        )
```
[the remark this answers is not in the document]
dshemetov: separating those two was already in the original code though. black added the parentheses, doubling the the vertical space of this code :/

dshemetov:
Trying to test and fix the [regex in question](https://github.com/cmu-delphi/delphi-epidata/blob/ds/format100/src/acquisition/norostat/norostat_raw.py#L83-L85) that SonarCloud doesn't like, but the norostat code doesn't even seem to be run correctly. It breaks in line 74 of that same file, where the output of `html_root.xpath(...)` is just `[]` for me 🤷 @krivard this makes me think this source doesn't run. Is it worth fixing?

krivard:
> this makes me think this source doesn't run. Is it worth fixing?

we're still running a norostat cronicle job daily, but it's been failing since November 2020. Checking with R&R whether to keep it or not.

dshemetov:
I used [`flynt`](https://github.com/ikamensh/flynt) to automatically convert many of the % and .format strings into official f-strings. It was pretty straightforward `pip install flynt` and `flynt src/acquisition -a -l 999 --verbose` (`-a` is aggressive, it won't change % strings otherwise, `-l 999` is line-length, it won't change long strings otherwise, `--verbose` is to have it log its changes). With this, it had a 94/98 success rate on %-style expressions, so I had to go fix a few manually.

krivard:
[We can kill norostat acquisition](https://github.com/cmu-delphi/delphi-epidata/issues/1205)

dshemetov:
Dropped Norostat changes from this and changed the base to the no-Norostat PR

src/acquisition/cdcp/cdc_dropbox_receiver.py
dshemetov: if you do "hide whitespace" up in the gear above, it makes these diffs way more readable

krivard (approved):
🚀 

thanks for fixing all the `fmt: on|off` cases!

nmdefries (approved):
Looks great!

PR #1191, opened 2023-06-07
properly handle and interpret X-Forwarded-For header

melange396:
configuration and handling to prevent potential ip address spoofing when behind certain kinds of proxies.

see loquacious comments inside the PR for more details.

### DIFF
--- a/src/server/_common.py
+++ b/src/server/_common.py
@@ -8,7 +8,7 @@ from werkzeug.exceptions import Unauthorized
 from werkzeug.local import LocalProxy
 
 from delphi.epidata.common.logger import get_structured_logger
-from ._config import SECRET, REVERSE_PROXIED
+from ._config import SECRET, REVERSE_PROXY_DEPTH
 from ._db import engine
 from ._exceptions import DatabaseErrorException, EpiDataException
 from ._security import current_user, _is_public_route, resolve_auth_token, show_no_api_key_warning, update_key_last_time_used, ERROR_MSG_INVALID_KEY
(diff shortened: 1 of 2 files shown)


src/server/_config.py
```diff
@@ -26,14 +26,37 @@
...
+
+it is important that REVERSE_PROXY_DEPTH is set accurately for two reasons...
+
+setting it too high (or to -1) will respect more of the entries in the "X-Forwarded-For" header than are appropriate.
+ this can allow remote addresses to be "spoofed" when a client fakes this header, carrying security/identity
+ implications.  in dev and testing, it is not particularly dangerous for this variable to be set to -1 (special case
+ for an "infinite" depth, where any and all proxy hops will be trusted).
+
+setting it too low can hinder logging accuracy -- that can cause an intermediate proxy IP address to be used as the
+ "real" client IP address.
```
krivard, line 53: ?

```suggestion
 "real" client IP address, which could cause requests to be rate-limited inappropriately.
```

src/server/_common.py
```diff
@@ -31,14 +31,29 @@ def _get_db() -> Connection:
...
         if "X-Forwarded-For" in req.headers:
-            return req.headers["X-Forwarded-For"].split(",")[
-                0
-            ]  # take the first (or only) address from the comma-sep list
+            full_proxy_chain = req.headers["X-Forwarded-For"].split(",")
+            # eliminate any extra addresses at the front of this list, as they could be spoofed.
+            if REVERSE_PROXY_DEPTH > 0:
+                depth = REVERSE_PROXY_DEPTH
+            else:
+                # special case for -1/negative: setting `depth` to 0 will not strip any items from the chain
```
krivard, line 46:
```suggestion
                # special case for -1/0/negative: setting `depth` to 0 will not strip any items from the chain
```

krivard (approved):
👍 
Minor comment fixes for clarity, only if accurate

src/server/_common.py
melange396: it cant have a value of zero here because of the `if REVERSE_PROXY_DEPTH:` above

krivard: 😱 

...that's probably fine

krivard (approved):
❤️ loquacious comments

PR #1193, opened 2023-06-09
update/fix PROXY_DEPTH default value, add 'diagnostics' endpoint

melange396:
(...and also some cleanup in admin endpoint file)

turns out...  the default value for `PROXY_DEPTH` set in https://github.com/cmu-delphi/delphi-epidata/pull/1191 is not high enough for the CC API servers and is likely not high enough for the AWS API servers.  iirc, the old value of "2" came from some testing on the staging setup, which is apparently not close enough to the configuration of either of the prod setups.

this change updates the value to match CC, which should be sufficient for AWS.  this change also adds a new endpoint (`api.delphi/epidata/admin/diagnostics`) which gives more visibility into the proxy chain of a running system (and maybe other things in the future).  it is only accessible with an admin token.


dmytrotsko (approved):
👍

PR #1202, opened 2023-06-21
Refactor covid_hosp auto columns

rzats:
### Summary:

Several refactors to `covid_hosp` files on the auto columns branch:
- Updates tests to get the number of columns directly from the YAML file - so they won't break when it changes
- Minimizes `Database` subclasses in `covid_hosp` (down to a single property) as the information stored there can now be obtained from the dataset definition YAML file
- Updates the base `Database` class to obtain the information in question
- Removes `Network` subclasses and condenses their logic into the base class
- Changes various tests as needed

### Prerequisites:

- [ ] ~~Unless it is a documentation hotfix it should be merged against the `dev` branch~~ `group/covid_hosp_auto_columns_2` base
- [x] Build is successful


src/acquisition/covid_hosp/common/database.py
```diff
@@ -12,17 +11,17 @@
...
 from delphi.epidata.common.logger import get_structured_logger
-
-Columndef = namedtuple("Columndef", "csv_name sql_name dtype")
+from delphi.epidata.common.covid_hosp.covid_hosp_schema_io import CovidHospSomething
 
 class Database:
 
+  DATASET_NAME = None
+  # Defined outside the constructor as this does not require a DB connection
+  METADATA_ID = CovidHospSomething().get_metadata_id(DATASET_NAME) if DATASET_NAME is not None else None
```
rzats, line 20: Isn't set alongside the other fields in the constructor, since it's only used by `Network` - and that doesn't require an actual `Database` instance, just the metadata ID string.

rzats:
Closing as the origin branch was remade.

PR #1211, opened 2023-06-26
Added host to database connection

dmytrotsko:
### Summary:

Specified database host in database connection strings to avoid wrong database using.

### Prerequisites:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [x] Build is successful
- [x] Code is cleaned up and formatted

PR #1212, opened 2023-06-26
Added host to database connection strings

dmytrotsko:
### Summary:

Added missing host to database connection strings to avoid wrong database connection
### Prerequisites:

- [x] Build is successful


melange396 (commented):
looks good...  but before merging (as in #1211), we need to figure out if there are data migrations that need to happen and plan/schedule those.  its also possible that many of these jobs may not even be on the run queue right now, but we will need someone with knowledge of and credentials for cronicle to help us determine that.

PR #1218, opened 2023-06-28
style(client): run py client through black

dshemetov:
Part of #962

### Summary:

- Run delphi_epidata.py through black and flynt
- Recommend "hide whitespace" setting for viewing diffs

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -20,684 +20,731 @@ from pkg_resources import get_distribution, DistributionNotFound
 # preference, even if you've installed it and then use this script independently
 # by accident.
 try:
-  _version = get_distribution('delphi-epidata').version
+    _version = get_distribution("delphi-epidata").version
 except DistributionNotFound:
-  _version = "0.script"
+    _version = "0.script"
 
-_HEADERS = {
-  "user-agent": "delphi_epidata/" + _version + " (Python)"
-}
+_HEADERS = {"user-agent": "delphi_epidata/" + _version + " (Python)"}
+
+
+class EpidataException(Exception):
+    pass
+
+
+class EpidataBadRequestException(EpidataException):
+    pass
+
+
+REGIONS_EPIWEEKS_REQUIRED = "`regions` and `epiweeks` are both required"
+ISSUES_LAG_EXCLUSIVE = "`issues` and `lag` are mutually exclusive"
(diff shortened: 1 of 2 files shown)


src/client/delphi_epidata.py
```diff
@@ -20,684 +20,737 @@
...
+        # Check parameters
+        if None in (
+            data_source,
+            signals,
+            time_type,
+            geo_type,
+            time_values,
+            geo_value,
+            sensor_names,
+        ):
```
nmdefries, line 693: nitpick [nonblocking]: Slight preference for having this all one one line. There's a similar `if` condition on line 563 that _is_ on one line. Would be nice to match.

nmdefries (dismissed):
Changes look good. A couple light/nonblocking comments.

src/client/delphi_epidata.py
```diff
@@ -20,684 +20,737 @@
...
-}
+_HEADERS = {"user-agent": "delphi_epidata/" + _version + " (Python)"}
+
+
+class EpidataException(Exception):
+    ...
+
+
+class EpidataBadRequestException(EpidataException):
+    ...
```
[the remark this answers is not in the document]
dshemetov: right, it just allows us to create more specific exception names. the ellipsis in python is alias for the "pass" keyword, which here is needed to terminate the class definition.

src/client/delphi_epidata.py
```diff
@@ -20,684 +20,731 @@
...
+    _version = "0.script"
 
-_HEADERS = {
-  "user-agent": "delphi_epidata/" + _version + " (Python)"
-}
+_HEADERS = {"user-agent": "delphi_epidata/" + _version + " (Python)"}
+
+
+class EpidataException(Exception):
+    ...
```
melange396, line 31: this looks like more than just a run through black!  at first i was quite impressed thinking that black did this automatically, but i came to my senses pretty quickly...

i think `pass` is a better token to use here because its not all punctuation, and i think its intent is a little clearer.  however, ymmv and its fine to leave it as-is.
```suggestion
    pass
```

dshemetov: blackGPT in 5 years probably will, but ye

and sure, i don't have a strong preference here.

nmdefries (approved):

nmdefries:
Oh, there's a conflict with the version number.

nmdefries (approved):

PR #1224, opened 2023-06-30
Set default older_than to tomorrow

krivard:
### Summary:

We currently wait until a day is over before pulling covid_hosp input files posted that day. Unfortunately, the new healthdata.gov posting schedule for covid_hosp data is Fridays and Mondays at ~12:10pm. Forecasting would like to use the fresh Monday data for the Monday forecasts they submit at 3pm ([Slack convo link](https://delphi-org.slack.com/archives/C013AH5N01E/p1688136661063719)). This means we should change the default pipeline behavior to permit same-day healthdata.gov files to be included in each update.

This PR updates the default selection criteria for which healthdata.gov files to import so that it includes data posted today. 

~WARNING: if healthdata.gov posts a file after 12:30 on a Monday or a Friday, **that file will never be picked up by the regular pipeline** and must be handled with a manual invocation of the pipeline that selects for the day the late file was posted.~ 7/17 edit: a fix for this edge case is in progress, thanks george!

This is a semi-hotfix. The covid_hosp acquisition materials are in the midst of a massive overhaul, but we'd like to implement this behavior change before the new code will be ready, so this PR modifies the old code.

### Prerequisites:

- [x] Build is successful


melange396:
**_this is untested in any way, and will probably need unit/integration test changes too._**

with the commit i just pushed, it "should" now look back to the same day as the last issue to see if there were any other new data revisions issued, instead of just looking back to the next day after the last issue.  any revisions from that last day that were already collected are skipped.

the new usage of the `with database.connect()` could maybe be widened in its scope (ie, pulled outisde of some number of the nested blocks), but i didnt want to keep the connection open while datasets are fetched and merged (which could potentially take a long time).

i think this also means, in normal usage, we should never again see the "no new issues; nothing to do" log message, but will now see the new log message "already collected revision:" on every run.

krivard:
oh noooooo this needs an ON DUPLICATE KEY UPDATE since we're no longer guaranteed to only insert once for each reference date + issue pair

krivard:
Okay @melange396 I've pushed up what I have and left a TODO where the ON DUPLICATE KEY should go

melange396:
i initially had a different change for the ON DUPLICATE KEY stuff where i rewrote more of the surrounding code for clarity and efficiency, but decided against that because of how it will complicate the integration with #1203 down the road.

dshemetov:
I have read through the failing test in `test_scenarios.py`. I have fixed one error and ran into another.

TL;DR: I don't think that the ON DUPLICATE KEY UPDATE part is working.

Here is my understanding:
- beginning in line 61, we add the rows from `dataset0.csv` into the db (the same data from two separate issues as specified by `metadata.csv`
- beginning in line 92, we test a second acquisition with the same metadata, just to make sure that a noop will occur; it is not necessary to choose a different dataset here, since the db insertion is rejected based on the metadata, not based on the values
- beginning in line 108, we test a third acquisition, but now with `metadata2.csv`; metadata2 has the same two issues as before and 2 new ones: one for a late 3/15 acquisition (that was missed the previous day) and one for a 3/16 acquisition; the data for both comes from `dataset1.csv`
  - the first two issues are ignored correctly
  - what _should_ then happen, is the 3/15 issue from before, with WY `critical_staffing_shortage_today_yes` value of 5, should be updated with a value of 8; the same should be true of the 3/16 issue; this is reflected in the assertions on lines 125 and 134; however, line 125 is failing with the 3/15 issue value staying at 5; this tells me that the `ON DUPLICATE KEY UPDATE` is currently not working correctly

I tinkered with it, but have no leads on what is wrong with the SQL statement.

melange396:
Looks like we arent even testing `ON DUPLICATE KEY UPDATE` (at least not in the place you're pointing out)...  For WY, `dataset0.csv` has value 5 and date 2020/12/09 (the date is under column "reporting_cutoff_start" which eventually gets renamed/mapped to "date" in the db), `dataset1.csv` has value 8 and date 2020/12/10...  So even with matching "issue"s, theres no key collision because the "date"s are different, and the value isn't overwritten.  Line 125 is correct in failing, the value should be 5.

dshemetov:
Good call @melange396, I fixed the test thanks to your pointer.

melange396 (approved):
LGTM, i think this is ready for merge and release!

PR #1228, opened 2023-07-10
Fix dashboard signals parsing for dashboard queries

dmytrotsko:
### Summary:

Fixed dashboard signals parsing for dashboard queries.
Used parse_source_signal_sets() function to parse (source, signal) pairs

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/src/server/_limiter.py
+++ b/src/server/_limiter.py
@@ -8,7 +8,15 @@ from ._common import app, get_real_ip_addr
 from ._config import RATE_LIMIT, RATELIMIT_STORAGE_URL, REDIS_HOST, REDIS_PASSWORD
 from ._exceptions import ValidationFailedException
 from ._params import extract_dates, extract_integers, extract_strings, parse_source_signal_sets
-from ._security import _is_public_route, current_user, require_api_key, show_no_api_key_warning, resolve_auth_token, ERROR_MSG_RATE_LIMIT, ERROR_MSG_MULTIPLES
+from ._security import (
+    _is_public_route,
+    current_user,
+    require_api_key,
+    show_no_api_key_warning,
+    resolve_auth_token,
+    ERROR_MSG_RATE_LIMIT,
+    ERROR_MSG_MULTIPLES,
+)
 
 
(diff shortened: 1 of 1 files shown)


src/server/_limiter.py
```diff
@@ -71,12 +71,13 @@ def get_multiples_count(request):
...
-        request_signals += extract_strings("signals")
-    if "data_source" in request.args:
-        request_signals = [f"{request.args['data_source']}:{request_signal}" for request_signal in request_signals]
+    request_args = request.args.keys()
+    if "signal" in request_args or "signals" in request_args:
+        source_signal_sets = parse_source_signal_sets()
+        for source_signal in source_signal_sets:
+            if isinstance(source_signal.signal, list):
+                for signal in source_signal.signal:
+                    request_signals.append(f"{source_signal.source}:{signal}")
```
melange396, line 80: we dont want to let users specify the wildcard "*" that matches all signals:
```suggestion
            if isinstance(source_signal.signal, bool):
                # wildcards are not allowed
                return False
            for signal in source_signal.signal:
                request_signals.append(f"{source_signal.source}:{signal}")
```

src/server/_limiter.py
```diff
@@ -71,12 +71,13 @@ def get_multiples_count(request):
...
     if len(request_signals) == 0:
         return False
     return all([signal in signals_allowlist for signal in request_signals])
```
melange396, line 83: (not directly related to this change, and i dont know why we didnt think of this before, but...) could we make this a little more efficient by using a set for `request_signals` instead of a list?
```suggestion
    return request_signals.issubset(signals_allowlist)
```

melange396 (changes requested):
beautiful!  one thing about handling the signal wildcard that is pretty important, and then two other little suggestions to think about.

src/server/_limiter.py
melange396: sorry, that wasnt specific enough:  we want to make sure users are rate limited when they specify the wildcard "*" that matches all signals

melange396 (approved):
looks great, nice work!

PR #1232, opened 2023-07-11
feat(R client): read from DELPHI_EPIDATA_KEY

dshemetov:
This should make CI integrations easier, by allowing us to set the env var to a secret. Also mimics behavior in [epidatr](https://github.com/cmu-delphi/epidatr/blob/dev/R/auth.R).

### Summary:

- read API key from DELPHI_EPIDATA_KEY env var by default, else None
  - don't disturb the existing read from "epidata.auth" R option

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/src/client/delphi_epidata.R
+++ b/src/client/delphi_epidata.R
@@ -18,6 +18,7 @@ Epidata <- (function() {
   client_version <- '4.1.5'
 
   auth <- getOption("epidata.auth", default = NA)
+  auth <- if (is.null(auth)) Sys.getenv("DELPHI_EPIDATA_KEY") else NA
 
   client_user_agent <- user_agent(paste("delphi_epidata/", client_version, " (R)", sep=""))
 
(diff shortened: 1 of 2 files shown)


krivard (approved):
👍 seems fine but I want george's thoughts too

if we keep it, please update the [corresponding docs ](https://github.com/cmu-delphi/delphi-epidata/blob/dev/docs/api/client_libraries.md)

melange396:
(note: im not really at all familiar with R test processes/procedures, but im presuming theyre something like the tests we use with the python code in this repository)

if the motivation is to be able to provide a secret API Key for CI tests, theres no need: those tests dont use production data or services, and we can insert any old "test" key into the local test database during the pre-test setups.

if using environment variables (for secrets) is a standard paradigm for R clients and/or packages, id say go for it, but that `getOption()` looks like it achieves that already.

dshemetov:
Yea, I'm just mimicking the convention of other packages - the R installers, for example, automatically pull out your GITHUB_PAT from your env vars, and so does the aws.s3 module. And even broader than R, the [aws cli](https://github.com/aws/aws-cli#configuration) does the same, so I figured this is an "industry-standard" thing 🤷

PR #1241, opened 2023-07-21
Add basic integration tests for all endpoints + API Keys tests

dmytrotsko:
### Summary:
Added  basic integration tests for all endpoints + API Keys tests

### Prerequisites:

- [x] Build is successful


src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+
+# third party
+import mysql.connector
+
+# first party
+from delphi.epidata.client.delphi_epidata import Epidata
+from delphi.epidata.server._limiter import limiter
+
+
+class BasicIntegrationTest(unittest.TestCase):
```
melange396, line 12:
```suggestion
class IntegrationTestBase(unittest.TestCase):
```
^ name suggestion to hopefully make it obvious that this is a "base" class.  (im also a little bit tempted to call it `DelphiTestBase`, but thats up to you)

---

this class has a number of things in common with our other `unittest.TestCase` subclass / test base class helper: [`CovidcastBase`](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/src/acquisition/covidcast/test_utils.py#L156).  `CovidcastBase` would best be redesigned as a subclass of the common `IntegrationTestBase` that then adds `epimetric_*` table truncations in setUp() and the additional row helper methods.

would it be difficult to combine them?

NOTE: maybe the suggested changes of combining / refactoring with `CovidcastBase` could be its own PR (if at all) after this one is merged.  also, `CovidcastBase` has a number of things i think should change, such as changing the name to `CovidcastTestBase` and relocating the file to `src/common/` instead of being under `src/acquisition/`.

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+    """Basic integration test class"""
+
+    def __init__(self, methodName: str = "runTest") -> None:
+        super().__init__(methodName)
+        self.delete_from_tables_list = []
+        self.truncate_tables_list = []
```
melange396, line 18: do we need both `DELETE` and `TRUNCATE`?  `TRUNCATE` is faster for bigger tables, but that shouldnt really matter in integration tests.  `TRUNCATE` will also reset `AUTO_INCREMENT` counters, but hopefully all of our tests are agnostic to any assigned `id` values and such.

it will be simpler to have only one of these, but if there is a good reason for having both, thats fine too.

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+        self.create_tables_list = []
+        self.role_name = None
+        self.epidata = Epidata
```
melange396, line 21: make this variable name a little more explanatory:
```suggestion
        self.epidata_client = Epidata
```

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+        self.epidata.auth = ("epidata", "key")
+
+    def delete_from_table(self, cur, table_name: str) -> None:
+        cur.execute(f"DELETE FROM `{table_name}`")
+
+    def truncate_table(self, cur, table_name: str) -> None:
+        cur.execute(f"TRUNCATE TABLE `{table_name}`")
+
+    def create_table(self, cur, create_table_stmt: str):
+        cur.execute(create_table_stmt)
```
melange396, line 32:
```suggestion
```
i think we can get rid of these methods since theyre pretty simple, and theyre only used one time each below

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def create_key_with_role(self, cur, role_name: str):
+        cur.execute("TRUNCATE TABLE `user_role`")
+        cur.execute("TRUNCATE TABLE `user_role_link`")
```
melange396, line 36: you should move these two `TRUNCATE`s to underneath `DELETE FROM api_user` in `setUp()`, so the table truncation only happens once per test, and then we should also be able to add multiple roles if needed in the future.

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+        cur.execute(f'INSERT INTO `api_user`(`api_key`, `email`) VALUES("{role_name}_key", "{role_name}_email")')
+        cur.execute(f'INSERT INTO `user_role`(`name`) VALUES("{role_name}")')
+        cur.execute(
+            f'INSERT INTO `user_role_link`(`user_id`, `role_id`) SELECT `api_user`.`id`, 1 FROM `api_user` WHERE `api_key`="{role_name}_key"'
```
melange396, line 40:
```suggestion
            f'INSERT INTO `user_role_link`(`user_id`, `role_id`) SELECT `api_user`.`id`, `user_role`.`id` FROM `api_user` JOIN `user_role` WHERE `api_user`.`api_key`="{role_name}_key" AND `user_role`.`name`="{role_name}"'
```
without the hardcoded role id of "1", this should be more flexible/generalizable

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,79 @@
...
+
+        cnx.commit()
+        cur.close()
+
+        self.cnx = cnx
+        self.cur = cnx.cursor()
+
+    @staticmethod
+    def _clear_limits() -> None:
+        limiter.storage.reset()
```
melange396, line 73: nice usage!

src/server/_config.py
```diff
@@ -93,8 +93,16 @@
 REDIS_HOST = os.environ.get("REDIS_HOST", "delphi_redis")
 REDIS_PASSWORD = os.environ.get("REDIS_PASSWORD", "1234")
 
+# testing mode to lower rate-limit in order to reduce number of required requests to hit the rate limit while runnig tests
+# by default is set to False when is not provided
```
melange396, line 97:
```suggestion
# mode to reduce number of required requests to hit rate limit while running tests,
# by default is set to False
```

.github/workflows/ci.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
 
       - name: Start delphi_web_epidata
         run: |
-          docker run --rm -d -p 10080:80 --env "MODULE_NAME=delphi.epidata.server.main" --env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" --env "FLASK_PREFIX=/epidata" --env "REDIS_HOST=delphi_redis" --env "REDIS_PASSWORD=1234" --env "API_KEY_REGISTER_WEBHOOK_TOKEN=abc" --env "API_KEY_ADMIN_PASSWORD=test_admin_password" --network delphi-net --name delphi_web_epidata delphi_web_epidata
+          docker run --rm -d -p 10080:80 --env "MODULE_NAME=delphi.epidata.server.main" --env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" --env "FLASK_PREFIX=/epidata" --env "REDIS_HOST=delphi_redis" --env "REDIS_PASSWORD=1234" --env "API_KEY_REGISTER_WEBHOOK_TOKEN=abc" --env "API_KEY_ADMIN_PASSWORD=test_admin_password" --env "TESTING_MODE=True" --network delphi-net --name delphi_web_epidata delphi_web_epidata
```
melange396, line 78: do we wanna make this an option in the `Makefile` as well?  (we definitely want to make it default in the Makefile)

another way to achieve the same effect (and only for tests that need to reach the request limiting threshold) would be to write into redis at the beginning of those tests to simulate a large number of requests...  if thats not too difficult to do, itll be more robust.

integrations/acquisition/covidcast/test_covidcast_meta_caching.py
```diff
@@ -58,6 +58,19 @@ def setUp(self):
...
+        password='pass',
+        host='delphi_database_epidata',
+        database='epidata')
+    epidata_cur = epidata_cnx.cursor()
+
+    epidata_cur.execute("DELETE FROM `api_user`")
+    epidata_cur.execute('INSERT INTO `api_user`(`api_key`, `email`) VALUES("key", "email")')
+    epidata_cnx.commit()
+    epidata_cur.close()
+    epidata_cnx.close()
```
melange396, line 72: you use this a couple times...  should we make it into a little helper method stored in `src/common/integration_test_base_class.py`?  it doesnt quite fit the requirements of https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming), so its fine to leave as-is.

integrations/server/test_api_keys.py
```diff
@@ -0,0 +1,243 @@
...
+        return response
+ 
+    def test_public_route(self):
+        """Test public route"""
+        public_route = "http://delphi_web_epidata/epidata/version"
+        status_codes = set()
+        for _ in range(10):
+            status_codes.add(self._make_request(public_route).status_code)
+        self.assertEqual(len(status_codes), 1)
+        self.assertEqual(next(iter(status_codes)), 200)
```
melange396, line 30:
```suggestion
        self.assertEqual(status_codes, {200})
```
this is a little more concise.  the `next(iter(...` syntax seems a little weird to me ¯\\\_(ツ)_/¯ 

(there are some similar usages of this below too)

integrations/server/test_api_keys.py
```diff
@@ -0,0 +1,243 @@
...
+            "time_type": "day",
+            "geo_type": "state",
+            "geo_value": "pa,ny",
+            "time_values": "20200406,20200407",
+        }
+        status_codes = set()
+        for _ in range(10):
+            status_codes.add(self._make_request(params=params).status_code)
+        self.assertEqual(len(status_codes), 2)
+        self.assertEqual(status_codes, {200, 429})
```
melange396, line 95:
```suggestion
        self.assertEqual(status_codes, {200, 429})
```
second line makes the first redundant

(more similar usages of this below too)

integrations/server/test_cdc.py
```diff
@@ -0,0 +1,43 @@
...
+from delphi.epidata.common.integration_test_base_class import BasicIntegrationTest
+
+
+class CdcTest(BasicIntegrationTest):
+    """Basic integration tests for cdc endpint."""
+
+    def setUp(self) -> None:
+        self.truncate_tables_list = ["cdc_extract"]
+        self.role_name = "cdc"
+        super().setUp()
```
melange396, line 11: if you follow a pattern similar to the [`localSetUp()` used in in `CovidcastBase`](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/src/acquisition/covidcast/test_utils.py#L168), you can avoid having to call `super()`

i believe you should also be able to set `truncate_tables_list` and `role_name` as class variables (outside of the class method scopes) to achieve the same effect, if you would rather try that instead

integrations/server/test_delphi.py
```diff
@@ -0,0 +1,63 @@
+# standard library
+import unittest
+
+# third party
```
melange396, line 4:
```suggestion
```

this can safely be removed, as far as i can tell

integrations/server/test_meta.py
```diff
@@ -0,0 +1,30 @@
...
+
+class MetaTest(BasicIntegrationTest):
+    """Basic integration tests for meta endpint."""
+
+    def setUp(self) -> None:
+        self.truncate_tables_list = ["forecasts", "fluview", "wiki", "wiki_meta", "twitter"]
+        super().setUp()
+
+    def test_meta(self):
+
```
melange396, line 13:
```suggestion
```
extra newline

integrations/server/test_norostat.py
```diff
@@ -0,0 +1,59 @@
...
+from delphi.epidata.common.integration_test_base_class import BasicIntegrationTest
+
+
+class NorostatTest(BasicIntegrationTest):
+    """Basic integration tests for norostat endpint."""
+
+    def setUp(self) -> None:
+        """Perform per-test setup."""
+
+        create_norostat_point_diffs = """
```
melange396, line 11: is this the same `CREATE TABLE` definition as in `test_meta_norostat.py`?  you could put the tests from both files into just one file to reduce repetition if you want!

melange396 (changes requested):
this is awesome!  You put a lot of work into this PR, and it will be very much worth the time when these tests help someone catch a bug later!

src/common/integration_test_base_class.py
dmytrotsko: Yeah, that renaming is reasonable.

Yeah, I have also thought about combining this class with CovidcastBase, but it will be better to do that in a separate PR(after this one is merged).

src/common/integration_test_base_class.py
dmytrotsko: That is unfortunate but yes, we need both of them. There are some tables with relations that can not be truncated(because of that relations) so the only way to clear them is to use `DELETE FROM <table>` statement.

.github/workflows/ci.yaml
dmytrotsko: Actually I believe that I have added this variable to Makefile as well, but for some reason it is not there now. Let me explore this a little bit more.

integrations/acquisition/covidcast/test_covidcast_meta_caching.py
dmytrotsko: Yeah, I added this in places where we `unittest.TestCase` is used and we don't have connection to the `epidata` database. This can be fixed in next round of refactoring.

integrations/server/test_api_keys.py
dmytrotsko: Nice comment. I believe that I had used this `next(iter())` construction in order to get first element of the set and getting element by index is not applicable for `set` data type. Don't unrderstand why I didn't try to cast it to list ¯\_(ツ)_/¯, but your solution looks even better.

integrations/server/test_cdc.py
dmytrotsko: Nice!

melange396 (changes requested):
looks great, just a few little things i noticed...

src/common/integration_test_base_class.py
```diff
@@ -0,0 +1,77 @@
...
+
+        cnx.commit()
+        cur.close()
+
+        self.cnx = cnx
+        self.cur = cnx.cursor()
+
+    def localSetUp(self):
+        # stub; override in subclasses to perform custom setup.
+        # runs after tables have been truncated but before database changes have been committed
```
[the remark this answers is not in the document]
dmytrotsko: Oops, my bad

integrations/server/test_meta_norostat.py
```diff
@@ -0,0 +1,87 @@
+# first party
+from delphi.epidata.common.integration_test_base_class import DelphiTestBase
+
+
+class MetaNorostatTest(DelphiTestBase):
```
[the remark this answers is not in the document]
dmytrotsko: I have deleted it, but forgot to include in the latest commit. Now it's removed.

melange396 (approved):
great job!

PR #1245, opened 2023-07-25
followup for #1203: deploy.json fix for src/common/*

melange396:
hopefully this works to make `github-deploy-repo` catch all the files from `src/common/` (#1203 added a file to `src/common/covid_hosp/`, and there are a total of 3 in that subdir now)

### DIFF
--- a/deploy.json
+++ b/deploy.json
@@ -24,7 +24,7 @@
       "dst": "[[package]]/common/",
       "match": "^.*\\.(py)$",
       "add-header-comment": true,
-	"recursive": true
+      "recursive": true
     },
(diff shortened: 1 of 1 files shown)


melange396:
> I am not 100% sure if `recursive` will do what we think it will do

i think/hope it will?  i cribbed it from some usages at the bottom of the same file, and a quick look at [the source](https://github.com/cmu-delphi/github-deploy-repo/blob/2c3e2f201fec7b96fa1e5412482b3ff510546d5b/src/actions/copymove.py#L142-L145) seems promising.

Issue #1247, opened 2023-07-26
`flusurv` data is stale


brookslogan:
The version of FluSurv-NET data available appears to be from 2021-05-28, containing data through the epiweek labeled with date 2020-04-19, while the [upstream source](https://gis.cdc.gov/GRASP/Fluview/FluHospRates.html) has data for the 2022/2023 flu season.

``` r
library(epidatr)
dat = flusurv(locations = "network_all", epiweeks = epirange(201701, 202301)) %>% fetch()
max(dat$release_date)
#> [1] "2021-05-28"
max(dat$epiweek)
#> [1] "2020-04-19"
names(dat)
#>  [1] "release_date" "location"     "issue"        "epiweek"      "lag"         
#>  [6] "rate_age_0"   "rate_age_1"   "rate_age_2"   "rate_age_3"   "rate_age_4"  
#> [11] "rate_overall"
```

<sup>Created on 2023-07-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

FluSurv-NET acquisition [broke](https://github.com/cmu-delphi/delphi-epidata/issues/242) circa 2020-10-09 but was patched to ignore age groups that were introduced then.  From the above sample, it looks like these/other age groups are still being ignored; upstream has 2 top-level age groups, 5 subgroups, and 8 subsubgroups; the API returns only 5 age groups.  I believe age group changes may have broken flusurv acquisition at some other point in time as well, so that might be a top suspect for the current breakage.

The [`fluview*` outage](https://github.com/cmu-delphi/delphi-epidata/issues/993) might be too late to be related.  FluSurv-NET reporting is not year-round; it typically starts at some point during the flu season when activity levels / influenza hospitalization numbers are deemed high enough (I don't remember the precise rule) and ends at/after the end of the flu season, with some break in issues and/or gap in measurements before the next season.

brookslogan:
Some other notes: [upstream](https://gis.cdc.gov/GRASP/Fluview/FluHospRates.html) source:
- has a few more breakdowns now which would be of interest: Sex, Race/Ethnicity, Season
- shows a blank plot for Cumulative Rate for 2020-2021 and "No Data Available" for corresponding Weekly Rates

nmdefries:
Looking at the oldest and newest logs available for flusurv acquisition (can't find any before April 2023):

[...]

Issue #242, opened 2020-10-09
flusurv acquisition is broken


dfarrow0:
The flusurv data acquisition steps have begun failing with the following stack trace:

```
current issue: 202040
fetching data for CA (2, 1)
[fetching flusurv data...]
[extracting values...]

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 189, in <module>
    main()
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 185, in main
    update(issue, args.location, args.test)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 99, in update
    data = flusurv.get_data(location_code)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv.py", line 194, in get_data
    data_out = extract_from_object(data_in)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv.py", line 162, in extract_from_object
    prev_rate = data_out[epiweek][age]
IndexError: list index out of range
```

---

Tip: reproduce locally by following [the dev guide](https://github.com/cmu-delphi/delphi-epidata/blob/main/docs/epidata_development.md) (i.e. bringing up database and web servers) and then running:
`docker run --rm delphi_python python3 -m delphi.epidata.acquisition.flusurv.flusurv_update CA`

---

The cause appears to be that the data now, as of 2020w40 (start of the 2020 flu season), includes 3 new age groups. This is great news in general. However, the definition of the new groups is unknown as [CDC's webapp](https://gis.cdc.gov/GRASP/Fluview/FluHospRates.html) doesn't appear to have been updated yet.

Once the site above is updated with definitions of the new age groups, we should capture them. This will require fairly straightforward updates to the acquisition script, the database table, the API server, and the API docs.

In the meantime, as a stop-gap, I'll submit a fix which simply ignores the new, unknown, age groups.

**TODO: revisit a week from today -- 2020-09-16**

dfarrow0:
webapp still not updated, new age groups still undefined. will check back in another week.

nmdefries:
[...]

PR #1278, opened 2023-08-23
Switch to new `flusurv` API endpoint

nmdefries:
### Summary:
Pull `flusurv` data from new CDC API endpoint. Ingest previously unlabelled age groupings. Ingest new race/eth, sex, and season breakdowns.

Closes https://github.com/cmu-delphi/delphi-epidata/issues/1247
Closes https://github.com/cmu-delphi/delphi-epidata/issues/242

Note: We would like to get in touch with CDC people responsible for maintaining their API so we can get advance notification of changes.

### Prerequisites:

- [ ] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [ ] Build is successful


src/acquisition/flusurv/new_grasp_location_result.json
```diff
@@ -0,0 +1,34 @@
+### New location-specific API response from https://gis.cdc.gov/GRASP/Flu3/PostPhase03DataTool?appVersion=Public
```
nmdefries, line 1: My json format is invalid, but I want to keep the comments describing sections. Leave as-is or convert to a `.txt` file?

melange396: does the python json library parse these files successfully?  they dont appear to be used anywhere, so im guessing theyre just for reference...  maybe they belong in a "reference" or "examples" subdirectory?

nmdefries: Yes, they're just for reference. Makes sense to move them :+1:

nmdefries: The CDC API response is pretty long and annoying to browse, so it was helpful to me to truncate and comment them. However, if these files seem unnecessary or too large, I'm also open to removing them.

src/acquisition/flusurv/flusurv.py
```diff
@@ -49,7 +51,7 @@
 
 # all currently available FluSurv locations and their associated codes
 # the number pair represents NetworkID and CatchmentID
-location_codes = {
+location_to_code = {
```
melange396, line 54: might we have to update this by hand in the future?  is there a programatic way to get this mapping?

src/acquisition/flusurv/flusurv.py
```diff
@@ -120,17 +122,46 @@ def fetch_json(path, payload, call_count=1, requests_impl=requests):
...
+    #    {
+    #        'default_data': {
+    #            'response': 'No Data'
+    #            }
+    #    }
+    #
+    # If data is returned, then data["default_data"] is a list
+    #  and data["default_data"]["response"] doesn't exist.
+    assert isinstance(result["default_data"], list) and len(result["default_data"]) > 0, \
+        f"Data was not correctly returned from the API for {location}"
```
melange396, line 154: should this always raise an exception?  that is, does a lack of data for a location indicate a problem on our end?

also, i think it is better to raise an explicit `Exception` here (as is done in other places in this file) instead of using the `assert` statement.

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
     """
-    Given a FluSurv data object, return hospitaliation rates.
+    Convert default data for a single location into an epiweek-grouped dictionary
 
-    The returned object is indexed first by epiweek, then by zero-indexed age
-    group.
-    """
+    Args:
+        data: The "default_data" element of a GRASP API response object,
+        as fetched with 'fetch_flusurv_location' or `fetch_flusurv_metadata`
```
melange396, line 182: distinguish the `metadata` arg in the docstring

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
+        epiweek = mmwrid_to_epiweek(obs["mmwrid"])
+        season = id_season_map[obs["seasonid"]]
+        groupname = groupids_to_name(
+            ageid = obs["ageid"], sexid = obs["sexid"], raceid = obs["raceid"],
+            id_label_map = id_label_map
+        )
+
+        # Set season description. This will be overwritten every iteration,
+        #  but should always have the same value per epiweek group.
+        data_out[epiweek]["season"] = season
```
melange396, line 233:
```suggestion
        data_out[epiweek]["season"] = id_season_map[obs["seasonid"]]
```
and delete the `season = ...` line above (or put this in its place)

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
+        if prev_rate is None:
+            # This is the first time to see a rate for this epiweek-group
+            #  combo
+            data_out[epiweek][groupname] = rate
+        elif prev_rate != rate:
+            # Skip and warn; a different rate was already found for this
+            # epiweek-group combo
+            warn((f"warning: Multiple rates seen for {epiweek} "
+                   f"{groupname}, but previous value {prev_rate} does not "
+                   f"equal new value {rate}. Using the first value."))
```
melange396, line 246: im guessing youve seen this in real data?  does it happen often and/or with large differences in value?  would it be worth halting acquisition when it happens, or maybe coming up with a way of escalating the message to where someone is sure to see it?

additionally, how would you feel about using [our logger module](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/common/logger.py) instead of the mixed `print()` and `warn()` statements?

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
+
+    # Sanity check the input. We expect to have populated our dictionary
+    if len(data_out.keys()) == 0:
+        raise Exception("no data loaded")
+
```
melange396, line 251:
```suggestion
```
probably not necessary if we know `len(data)` is non-zero ¯\\\_(ツ)_/¯  doesnt really hurt though

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
+    print(f"[successfully fetched data for {location}]")
     return data_out
 
 
-def get_current_issue():
-    """Scrape the current issue from the FluSurv main page."""
-
-    # fetch
-    data = fetch_json("GetPhase03InitApp?appVersion=Public", None)
+def get_current_issue(data):
```
melange396, line 278: this is only called on the response from `fetch_flusurv_metadata()`, right?
```suggestion
def get_current_issue(metadata):
```

src/acquisition/flusurv/flusurv.py
```diff
@@ -142,78 +173,192 @@ def mmwrid_to_epiweek(mmwrid):
...
+    id_to_label = defaultdict(lambda: defaultdict(lambda: None))
+    for season in metadata["seasons"]:
+        id_to_label[season["seasonid"]] = season["label"]
+
+    return id_to_label
+
+
+def groupids_to_name(ageid, sexid, raceid, id_label_map):
+    # Expect at least 2 of three ids to be 0
+    assert (ageid, sexid, raceid).count(0) >= 2, \
```
melange396, line 323: raise explicit `Exception` instead of using `assert`

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -81,6 +82,8 @@
 from delphi.utils.epiweek import delta_epiweeks
 
 
+max_age_to_consider_weeks = 52
```
melange396, line 85: use caps since its a constant
```suggestion
MAX_AGE_TO_CONSIDER_WEEKS = 52
```

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -113,38 +112,185 @@ def update(issue, location_name, test_mode=False):
...
+      `rate_race_hisp`,
+      `rate_race_asian`,
+      `rate_race_natamer`,
+
+      `rate_sex_male`,
+      `rate_sex_female`
     )
     VALUES (
-      %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s
+      %(release_date)s,
```
melange396, line 151: named instead of ordered parameters: 👍👍👍

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -113,38 +112,185 @@ def update(issue, location_name, test_mode=False):
...
+
+      `rate_sex_male` = coalesce(%(rate_sex_male)s, `rate_sex_male`),
+      `rate_sex_female` = coalesce(%(rate_sex_female)s, `rate_sex_female`)
     """
 
     # insert/update each row of data (one per epiweek)
     for epiweek in epiweeks:
         lag = delta_epiweeks(epiweek, issue)
-        if lag > 52:
+        if lag > max_age_to_consider_weeks:
```
melange396, line 220: does the `seasonids` creation/filtering negate the need for this check?

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -113,38 +112,185 @@ def update(issue, location_name, test_mode=False):
...
             # values (including duplicates) were stored on each run.
             continue
-        args_meta = [release_date, issue, epiweek, location, lag]
-        args_insert = data[epiweek]
-        args_update = [release_date] + data[epiweek]
-        cur.execute(sql, tuple(args_meta + args_insert + args_update))
+
+        # As of Sept 2023, for new data we expect to see these 23 groups, as
+        #  described in the top-level "master_lookup" element of the new GRASP API
+        #  (https://gis.cdc.gov/GRASP/Flu3/PostPhase03DataTool) response object:
```
melange396, line 228: a few things about this comment block:
 * its better to make it a footnote with the bulk of the text somewhere else in this file instead of interrupting the method (and loop!) with a huge comment.  maybe even make `n_expected_groups` into a top-level constant?
 * it might be easier to read as an unassigned multi-line string instead of using the octothorpe prefix on each line
 * its probably worth it to mention your sample json files in `./reference/`

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -113,38 +112,185 @@ def update(issue, location_name, test_mode=False):
...
+        #        {'label': '75-84 yr', 'ageid': 8, 'color_hexvalue': '#CAB2D6'},
+        #        {'label': '85+', 'ageid': 9, 'color_hexvalue': '#1f78b4'}
+        #    ]
+        #
+        #  In addition to the new age, race, and sex breakdowns, the group
+        #  id for overall reporting has changed from 6 to 0. Ageids 1-5
+        #  and 7-9 retain the same the same meanings.
+        n_expected_groups = 23
+        # Subtract one since we also store the season description in each epiweek value
+        if len(data[epiweek].keys() - 1) != n_expected_groups:
```
melange396, line 281: does this even work??  this subtract/minus notation can be used to remove an element from a set, but it doesnt seem to work for removing an integer from the set:
```
>>> d = {'a': 2, 1: 3}
>>> d.keys() 
dict_keys(['a', 1])
>>> d.keys() - 'a'
{1}
>>> d.keys() - 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'int' object is not iterable
```

i think maybe you wanted this instead?
```suggestion
        if len(data[epiweek].keys()) - 1 != n_expected_groups:
```

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -113,38 +112,185 @@ def update(issue, location_name, test_mode=False):
...
+            warnings.warn(
+                f"{location} {epiweek} data does not contain the expected {n_expected_groups} groups"
```
melange396, line 283: for ease of debugging problems we might see in production, print the number of groups seen, and possibly even the group names themselves.

you may want to differentiate between <23 groups (implying some are missing (would that be a fatal error in the `cur.execute()` call below, if a named value isnt in the params list?)) and >23 groups (implying new groups have been added, which we might want to investigate and eventually add to the db).

if you want to ***really*** over engineer this and super future-proof it, you could save the set of 23 expected keys instead of just the count, and print seen keys minus expected as well as expected keys minus seen.  that might be helpful in the possible event that groups are renamed (but the count will still be 23).

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -175,18 +321,28 @@ def main():
...
     # scrape current issue from the main page
-    issue = flusurv.get_current_issue()
+    issue = flusurv.get_current_issue(metadata)
     print(f"current issue: {int(issue)}")
 
+    # Ignore seasons with all dates older than one year
+    seasonids = {
+        season_blob["seasonid"] for season_blob in metadata["seasons"]
+        if delta_epiweeks(flusurv.mmwrid_to_epiweek(season_blob["endweek"]), issue) < max_age_to_consider_weeks
+    }
```
melange396, line 334: this is kind of a tricky, and maybe not worth the effort, but.......

does it make sense to push some of this functionality into `update()` or (more preferably) even further into `flusurv.get_data()`?

when this module has lots of intricate calls of specific stuff from `flusurv.py`, it loses its [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)#/media/File:CouplingVsCohesion.svg)  ...and then we might just as well combine the modules `flusurv.py` and `flusurv_update.py`.

i understand trying to make only one network call to retrieve `metadata` -- if youre worried about reprocessing stuff found in `metadata`, you could do something like making a `FlusurvMetadata` class that precomputes `issue` and `seasonids` and (from `flusurv.py`) `id_label_map` and `id_season_map` once each, and then pass an instance of that around.

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -175,18 +321,28 @@ def main():
...
         # all locations
-        for location in flusurv.location_codes.keys():
-            update(issue, location, args.test)
+        for location in flusurv.location_to_code.keys():
+            update(issue, location, seasonids, metadata, args.test)
     else:
         # single location
-        update(issue, args.location, args.test)
+        assert args.location in flusurv.location_to_code.keys(), \
+            f"Requested location {args.location} not available"
```
melange396, line 344: nice check!  how about raising a proper `Exception` instead of the `assert`?  also, the `.keys()` call after `in` isnt necessary.
```suggestion
        assert args.location in flusurv.location_to_code, \
            f"Requested location {args.location} not available"
```

melange396 (changes requested):
nice work!  i really appreciate the thorough commenting!

youll wanna pull in the changes from the `dev` branch, PR #1241 added [tests for the flusurv endpoint](https://github.com/cmu-delphi/delphi-epidata/blob/dev/integrations/server/test_flusurv.py).

src/acquisition/flusurv/flusurv.py
nmdefries: I reused logic that was already in the script; I haven't actually seen this in real data.

This was apparently set up to [match the behavior of the dashboard](https://github.com/cmu-delphi/delphi-epidata/blob/9f55c8197f7eea3e41e483c175e7cb87e0655e53/src/acquisition/flusurv/flusurv.py#L11-L15) that we scrape. Apparently we used to think the [data was versioned](https://github.com/cmu-delphi/delphi-epidata/blob/14c102a38c1a8d575b877f402fae676c9b7ea8ea/src/acquisition/flusurv/flusurv.py#L11-L17) or at least treat it like it was. It's unclear how much of this is sleuthing/speculation on our part, since the API we're using is a CDC internal API with no documentation. And the dashboard contains extra processing itself.

So... I could understand keeping this as-is, since the behavior matches the source's. It also would make sense for this to fail if we see contradictory values so we can investigate. Theoretically we have logs from all these runs that would tell us how often this happens, but without the script failing no one's going to look into it.

nmdefries: It makes sense to use more formal logging, but given this PR is already pretty large, maybe that should be a separate change. Unclear ATM how extensively I should refactor this (metadata class, e.g.), so I'll open an issue for now.

src/acquisition/flusurv/flusurv.py
nmdefries: Yeah, this can be constructed from the metadata JSON, but would involve passing this obj around a bit awkwardly. Could be a point in favor of a metadata class.

src/acquisition/flusurv/flusurv.py
nmdefries: > does a lack of data for a location indicate a problem on our end?

I don't know. Given that this has been effectively down for 2 years, I thought it would be good to over-alert rather than under-alert. Again, the API is internal to the CDC so we don't really know what normal behavior is for it.

> Exception -> assert

:+1:

src/acquisition/flusurv/flusurv.py
nmdefries: Same thing as above about wanting to over-alert, but I could be convinced to remove this particular check.

src/acquisition/flusurv/flusurv_update.py
nmdefries: The previous version made me itchy.

Even nicer would be if we could insert each value-name as a pair ("%(release_date)s as `release_date`, %(issue)s as `issue`, ..."). I don't suppose there's a way to do that in SQL. I looked briefly but couldn't find anything.

src/acquisition/flusurv/flusurv.py
brookslogan: Not sure these are the cases you're talking about here, but:
- FluSurv-NET isn't always the same set of states, but the set should only change once in a while, maybe exclusively between seasons.
- On their dashboard, I only remember seeing "No Data Available." for weekly rates in 2020-21, when you get to them by first looking at the 2020-21 Cumulative Rate then switching to Weekly Rate.

nmdefries: Okay, good to know that the location set does change. In that case, maybe we need to decide what locations to request based on what the API metadata dynamically says is available. Right now we use a hard-coded list, which means new locations would not be requested on our side. This means that erroring for missing locations also might be overkill -- maybe will switch this to a warning.

On the missing 2020-2021 data, I think this is a dashboard issue. I'll double-check, but the API response seems to include data fro that range.

src/acquisition/flusurv/flusurv.py
nmdefries: Logging changes proposed in https://github.com/cmu-delphi/delphi-epidata/issues/1294

nmdefries: Discussion about behavior when we see two values for the same geo-date https://github.com/cmu-delphi/delphi-epidata/issues/1331

src/acquisition/flusurv/flusurv_update.py
nmdefries: Going to leave `.keys()` for clarity.

src/acquisition/flusurv/flusurv_update.py
nmdefries: No, they're not exactly the same. Each seasonid corresponds to a flu season, which spans several months. So some data from a season could fall within the 52-week cutoff we're using and some could be outside that. The 52-week filter is required; the seasonid filter is not.

The seasonid filtering happens before we make any API calls. Given that the data spans a long time period (back to 2009?), I didn't want to fetch its whole history. Realistically, since this is weekly data, fetching only recent seasons probably doesn't make a big difference on runtime or memory usage. Just a preference, but I'm open to feedback on this. It does add complexity.

src/acquisition/flusurv/flusurv_update.py
nmdefries: Thanks for mentioning cohesion, I hadn't heard of that concept before. I created a metadata class like you suggested and also a data-fetcher class that uses the metadata one. Only the fetcher class is called from the wrapper/main file.

src/acquisition/flusurv/flusurv.py
nmdefries: I modified this so that we only request data for locations that are reported in the metadata to be available for recent dates. So if we don't get any data back, that actually is a problem.

melange396:
I wasnt paying attention and accidentally merged #1287 into this PR.  I then reverted #1287 in #1426.  This is a reminder that #1287 will eventually need to be re-merged into here (which is now probably best done by reverting the revert in #1426).  Sorry for the confusion!

aysim319:
ran the code locally and fails with 
```
found data for 82 epiweeks
[successfully fetched data for network_all]
rows before: 0
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 213, in <module>
    main()
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 204, in main
    update(fetcher, location, args.test)
  File "/usr/src/app/delphi/epidata/acquisition/flusurv/flusurv_update.py", line 155, in update
    raise Exception(
Exception: network_all 202401 data includes new group(s) {'rate_age_1t4', 'rate_age_gte75', 'rate_age_0tlt1'}
```

nmdefries:
> Exception: network_all 202401 data includes new group(s) {'rate_age_1t4', 'rate_age_gte75', 'rate_age_0tlt1'}

This should be resolved just by adding those signals to the list of those expected.

nmdefries:
This works now. I added a migration `.sql` script to add the new columns to the schema.

nmdefries:
@melange396 @aysim319 This is ready for review.

nmdefries:
I duplicated some of the work in https://github.com/cmu-delphi/delphi-epidata/pull/1287 in https://github.com/cmu-delphi/delphi-epidata/pull/1278/commits/c88c6fcd10a48200b45cc5c71080f8a2830026f4. (#1287 needs to be re-merged into this PR anyway, see [comment](https://github.com/cmu-delphi/delphi-epidata/pull/1278#issuecomment-2099518635).)

nmdefries:
These will need to be updated with new fields:
- https://github.com/cmu-delphi/delphi-epidata/pull/1287
- https://github.com/cmu-delphi/delphi-admin/pull/171
- https://github.com/cmu-delphi/epidatr/pull/210
- https://github.com/cmu-delphi/epidatpy/pull/26

src/acquisition/flusurv/reference/new_grasp_result.json
```diff
@@ -0,0 +1,167 @@
+### New API response from https://gis.cdc.gov/GRASP/Flu3/PostPhase03DataTool?appVersion=Public
```
nmdefries, line 4: note: the red highlighting here seems to be because the JSON format is invalid (JSON doesn't allow comments and also maybe single quotes are disallowed). However, this file is for reference only, so it doesn't really matter.

nmdefries:
Source data is [here](https://gis.cdc.gov/GRASP/Fluview/FluHospRates.html).

src/acquisition/flusurv/flusurv_update.py
```diff
@@ -165,6 +193,11 @@ def main():
         "location",
```
nmdefries, line 167: suggestion: default to "all" for convenience.

src/acquisition/flusurv/api.py
```diff
@@ -0,0 +1,345 @@
...
+    # the FluSurv AMF server
+    flusurv_url = FLUSURV_BASE_URL + path
+
+    # request headers
+    headers = {
+        "Accept-Encoding": "gzip",
+        "User-Agent": USER_AGENT,
+    }
+    if payload is not None:
+        headers["Content-Type"] = "application/json;charset=UTF-8"
```
nmdefries: suggestion: move to `else` block below.

src/acquisition/flusurv/api.py
```diff
@@ -0,0 +1,345 @@
...
+    epiweek_200340 = EpiDate(2003, 9, 28)
+    mmwrid_200340 = 2179
+    return epiweek_200340.add_weeks(mmwrid - mmwrid_200340).get_ew()
+
+
+class FlusurvMetadata:
+    def __init__(self, max_age_weeks):
+        self.metadata = self._fetch_flusurv_metadata()
+
+        self.issue = self._get_current_issue()
```
nmdefries: style: it is more idiomatic for these internal methods to set the attributes themselves, rather than assigning in `init`. Still would call the methods in `init`, though.

src/acquisition/flusurv/api.py
```diff
@@ -0,0 +1,345 @@
...
+            metadata: dictionary representing a JSON response from the FluSurv API
+        """
+        # extract
+        date = datetime.strptime(self.metadata["loaddatetime"], "%b %d, %Y")
+
+        # convert and return
+        return EpiDate(date.year, date.month, date.day).get_ew()
+
+    def _get_recent_seasonids(self):
+        # Ignore seasons with all dates older than one year
```
nmdefries: issue: 
```suggestion
        # Ignore seasons with all dates older than `self.max_age_weeks` (from user command line argument `max_age`)
```

src/acquisition/flusurv/reference/new_grasp_result.json
nmdefries: suggestion: put this comment in actual file for future reference.

src/acquisition/flusurv/api.py
```diff
@@ -0,0 +1,345 @@
...
+
+    # return the decoded json object
+    return resp.json()
+
+
+def mmwrid_to_epiweek(mmwrid):
+    """Convert a CDC week index into an epiweek."""
+
+    # Add the difference in IDs, which are sequential, to a reference
+    # epiweek, which is 2003w40 in this case.
```
nmdefries: suggestion: note that season 1/week 0 corresponds to some date in 1960.

src/acquisition/flusurv/reference/new_grasp_result.json
```diff
@@ -0,0 +1,167 @@
...
+    'seasons': [
+        {'description': 'Season 2006-07', 'enabled': True, 'endweek': 2387, 'label': '2006-07', 'seasonid': 46, 'startweek': 2336, 'IncludeWeeklyRatesAndStrata': True}, 
+        {'description': 'Season 2003-04', 'enabled': True, 'endweek': 2231, 'label': '2003-04', 'seasonid': 43, 'startweek': 2179, 'IncludeWeeklyRatesAndStrata': True}, 
+        {'description': 'Season 2009-10', 'enabled': True, 'endweek': 2544, 'label': '2009-10', 'seasonid': 49, 'startweek': 2488, 'IncludeWeeklyRatesAndStrata': True}, 
+        {'description': 'Season 2012-13', 'enabled': True, 'endweek': 2700, 'label': '2012-13', 'seasonid': 52, 'startweek': 2649, 'IncludeWeeklyRatesAndStrata': True}, 
+        {'description': 'Season 2015-16', 'enabled': True, 'endweek': 2857, 'label': '2015-16', 'seasonid': 55, 'startweek': 2806, 'IncludeWeeklyRatesAndStrata': True}, 
+        ...
+    ],
+    # Descriptions of data repporting groups (age, race, sex, and overall). Careful, some of these overlap.
+    # some of "parentidoverlap.
```
nmdefries: issue: typo

```suggestion
```

integrations/server/test_flusurv.py
```diff
@@ -11,8 +11,39 @@ def localSetUp(self):
     def test_flusurv(self):
         """Basic integration test for flusurv endpoint"""
         self.cur.execute(
-            "INSERT INTO `flusurv`(`release_date`, `issue`, `epiweek`, `location`, `lag`, `rate_age_0`, `rate_age_1`, `rate_age_2`, `rate_age_3`, `rate_age_4`, `rate_overall`, `rate_age_5`, `rate_age_6`, `rate_age_7`) VALUES(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
-            ("2012-11-02", "201243", "201143", "CA", "52", "0", "0", "0", "0.151", "0", "0.029", "0", "0", "0"),
+            """
```
nmdefries, line 16: suggestion [optional]: might also want to test out whole pipeline, where we mock the source API pull (getting bigger data from a local JSON). At least turn into a TODO.

src/acquisition/flusurv/flusurv_update.py
nmdefries: `argparser` requires this argument even if I set a default unless we convert location to a named argument. I'll leave this as-is so we don't have to change the production command that runs this.

PR #1253, opened 2023-08-08
Performance testing testing workflow

rzats:
Related: https://github.com/cmu-delphi/delphi-epidata/issues/1206

### Summary:

Adds a one-time testing workflow that will periodically run the performance scaling test rig and see how the results stack up. This is set to run every 30 minutes on August 9th, and can be safely deleted/reverted afterwards.

PR #1282, opened 2023-08-25
Remove temp testing workflow & route on-demand workflow to selfhosted runner

rzats:
### Summary:

- Removes the one-time "performance testing testing" workflow from https://github.com/cmu-delphi/delphi-epidata/pull/1253.
- Modifies the performance testing workflow to route it to our self-hosted test runner, `gh-runner-prod-cc-prv-01`.

### Prerequisites:

- [x] Build is successful


korlaxxalrok (approved):
Yep, looks good!

(This may not work quite as expected until we modify/remove the WireGuard setup, but if it can be left as is will "just work" then even better, so I guess we try and see what happens)

PR #1284, opened 2023-08-30
Fix ISO range parsing

dmytrotsko:
### Summary:

Fixed ISO range parsing for /covidcast endpoint

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/src/server/_params.py
+++ b/src/server/_params.py
@@ -426,18 +426,17 @@ def extract_dates(key: Union[str, Sequence[str]]) -> Optional[TimeValues]:
     values: TimeValues = []
 
     for part in parts:
-        if "*" in part:
-            values.append(part)
-            continue
+        if part == "*":
+            return ["*"]
         if ":" in part:
             # YYYY-MM-DD:YYYY-MM-DD
-            range_part = part.split(":", 2)
+            range_part = part.split(":", 1)
             r = _verify_range(parse_date(range_part[0]), parse_date(range_part[1]))
             if r is None:
                 return None
             values.append(r)
             continue
-        # parse another date formats
+        # parse other date formats
         r = parse_day_value(part)
         values.append(r)
     return values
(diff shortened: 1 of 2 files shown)


dshemetov:
Could you add some tests for this case [here](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/tests/server/test_params.py#L452)?

src/server/_params.py
```diff
@@ -425,44 +425,21 @@ def extract_dates(key: Union[str, Sequence[str]]) -> Optional[TimeValues]:
...
-        # the range is inverted, this is an error
-        raise ValidationFailedException(f"{key}: the given range is inverted")
-
     for part in parts:
-        if "-" not in part and ":" not in part:
-            # YYYYMMDD
-            values.append(parse_date(part))
+        if "*" in part:
+            values.append(part)
             continue
```
melange396, line 431:
```suggestion
        if "*" == part:
            return ["*"]
```
i think the `==` part is required, otherwise itll match a `part` like "x*y"

src/server/_params.py
```diff
@@ -425,44 +425,21 @@ def extract_dates(key: Union[str, Sequence[str]]) -> Optional[TimeValues]:
...
-            r = push_range(range_part[0], range_part[1])
+            r = _verify_range(parse_date(range_part[0]), parse_date(range_part[1]))
             if r is None:
                 return None
             values.append(r)
             continue
-        # YYYY-MM-DD
-        values.append(parse_date(part))
-    # success, return the list
+        # parse another date formats
```
melange396, line 440:
```suggestion
        # parse other date formats
```

src/server/_params.py
melange396: ...the `return` suggestion seems like a good way to short-circuit the rest of the argument processing -- if we have a wildcard to allow all dates, the other dates should be redundant

melange396 (changes requested):
looks good, just a few small things.  Dmitry's suggestion is worth doing too.

src/server/_params.py
dmytrotsko: Yeah, good point!

melange396 (approved):
LGTM!

PR #1288, opened 2023-09-19
Remove usage of PHP alias in the Python client

### LINKED ISSUE #1285: remove usage of php alias in the python client
Get rid of the use of the "`api.php`" generic compatibility alias in the python delphi_epidata client, referenced by [the `BASE_URL` class variable](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/src/client/delphi_epidata.py#L47).  Do this by changing the client to use the proper URL path suffix for each of the individual endpoints instead of specifying them in the `endpoint` parameter on calls to `_request()`.

This has a number of benefits, including:
  * a higher upper limit on the number of rows returned per request (see limits [here](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/src/server/_config.py#L12-L13))
  * better/easier logging aggregation and analysis (some request parameter values are used differently in the generic compatibility (php alias) endpoint vs in the other endpoints when used directly)
  * movement toward finally deprecating and removing the php alias endpoint entirely

The client is used by a number of integration tests in this repository.  Externally, the client is [used in the covidcast-indicators repo](https://github.com/cmu-delphi/covidcast-indicators/blob/39c02fa228fcb4de2b3f3d19469e862861351a18/hhs_hosp/delphi_hhs/run.py#L11), and [built upon in the core of the covidcast client](https://github.com/cmu-delphi/covidcast/blob/1e09e47df4748474c07e7c5ded07da6abb7964b1/Python-packages/covidcast-py/covidcast/covidcast.py#L9).  Changes may need to be made to those as well to ensure they support any edits made to the client.

rzats:
Closes #1285.

### Summary:

Removes all references to the `api.php` compatibility endpoint in `src/client/delphi_epidata.py`, instead redirecting all requests to their respective endpoints.

This has some implications on the outputs of the client, as it is taken out of "compatibility mode", which [would normally be set](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/main.py#L43) by the call to the `api.php` endpoint. This means:

- outputs in general [are slightly different](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/_printer.py#L27-L29), e.g., adding an empty `"epidata": []` field on missing rows
- several extra columns are [always included](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/endpoints/covidcast.py#L87-L92) in the `covidcast` endpoint

Those changes are reflected in the integration tests.

### Prerequisites:

- [x] Build is successful


melange396 (changes requested):
Looks good so far!

Because the client relied on the use of the compatibility endpoint, this PR effectively removes a LOT of test coverage for that endpoint...  Can you expand on the tiny bit of api.php-specific testing left in [`test_covidcast_endpoints.py`](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/integrations/server/test_covidcast_endpoints.py#L104-L116) to make up for some of the loss?  Using the `as_api_compatibility_row_*` methods from `src/common/covidcast_row.py` might make some of that easier.

Have you checked yet to make sure the covidcast client and covidcast-indicators are compatible with these changes (mentioned in the first comment of the associated issue, #1285)?  After you verify the covidcast client still works properly, you will need to change [this line in `signal_dash_data_generator.py`](https://github.com/cmu-delphi/delphi-epidata/blob/f7da6598a810be8df5374e3a71512c631c3a14f1/src/maintenance/signal_dash_data_generator.py#L22).

Also, a quick search through the code shows some comments still mention "api.php" where they probably shouldnt.

melange396:
> * outputs in general [are slightly different](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/_printer.py?rgh-link-date=2023-09-19T12%3A39%3A42Z#L27-L29), e.g., adding an empty `"epidata": []` field on missing rows
> * several extra columns are [always included](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/server/endpoints/covidcast.py?rgh-link-date=2023-09-19T12%3A39%3A42Z#L87-L92) in the `covidcast` endpoint

Good point!  Can you create `src/client/packaging/pypi/CHANGELOG.md` (perhaps using [this template](https://gist.github.com/juampynr/4c18214a8eb554084e21d6e288a18a2c)) and note that as of this version, returned results will now always have the `epidata` key in the root dict, and the other field keys at the appropriate locations?

src/client/delphi_epidata.py
```diff
@@ -136,18 +136,13 @@ def fluview(regions, epiweeks, issues=None, lag=None, auth=None):
         if auth is not None:
             params["auth"] = auth
         # Make the API call
-        return Epidata._request(params)
+        return Epidata._request(params, "fluview")
```
dshemetov, line 139:
```suggestion
        return Epidata._request("fluview", params)
```

here and the other endpoints

dshemetov (changes requested):
Looking good, but second @melange396's comments about checking downstream effects on the covidcast Python client (and [any other places](https://github.com/search?q=org%3Acmu-delphi+%22from+delphi_epidata+import+Epidata%22&type=code) that `from delphi_epidata import Epidata` such as [hhs](https://github.com/cmu-delphi/covidcast-indicators/blob/39c02fa228fcb4de2b3f3d19469e862861351a18/hhs_hosp/delphi_hhs/run.py#L11)).

rzats:
@melange396 @dshemetov fixed the issues you've mentioned, and also created two PRs to fix the compatibility changes which will be caused by this being merged:

https://github.com/cmu-delphi/covidcast-indicators/pull/1899
https://github.com/cmu-delphi/covidcast/pull/663

capnrefsmmat:
This is a breaking change to the Python client API; we'll want to make sure we update the version number on release accordingly.

melange396:
from https://github.com/cmu-delphi/covidcast/pull/663#discussion_r1334455733

@capnrefsmmat said:
> since the next version has a breaking API change, shouldn't we be incrementing more than the minor version number?

and i said:
> In even a fairly loose interpretation of the semantic versioning rules: yeah, we probably should be.
> 
> At the same time, bumping the major number would belie the fact that this is a relatively tiny change in the scope of the rest of the repository... Theres no new functionality and the only real effective change is to the signature of one public method, `Epidata.async_epidata()`. I would also argue that method is probably _**very**_ rarely used, though thats just based on a hunch -- i find it hard to believe that many (if any?) people are actually using it because it is poorly documented and it doesnt appear in established call chains, but i have no demonstrable evidence for this. [It _**IS**_ used [once in `covidcast-indicators`](https://github.com/cmu-delphi/covidcast-indicators/blob/39c02fa228fcb4de2b3f3d19469e862861351a18/nowcast/delphi_nowcast/epidata.py#L51), but we have full control over that code & execution.]
> 
> Still, your point stands and this change is worth more than a single increment of the patch-level version.
> 
> This would be a lot easier if the client and its versioning were decoupled from the rest of the API code, but we wont truly get that until we release and get adoption of `epidatpy`, which lives in its own repo.

melange396:
We ***COULD*** decouple the python client version from the repository version by changing the PyPI metadata [here](https://github.com/cmu-delphi/delphi-epidata/blob/b5128f09779de310ccd775a8a83edb5205372291/src/client/packaging/pypi/setup.py#L8) and [here](https://github.com/cmu-delphi/delphi-epidata/blob/b5128f09779de310ccd775a8a83edb5205372291/src/client/packaging/pypi/delphi_epidata/__init__.py#L4), removing those files from the automatic version bumping [here](https://github.com/cmu-delphi/delphi-epidata/blob/b5128f09779de310ccd775a8a83edb5205372291/.bumpversion.cfg#L14-L16), and then changing the version in the client code [here](https://github.com/cmu-delphi/delphi-epidata/blob/b5128f09779de310ccd775a8a83edb5205372291/src/client/delphi_epidata.py#L19-L25).

That would require us to remember to update the version number by hand in all those places if/when the python client code changes again (or come up with an automatic solution).  On the plus side, that would stop us from a ridiculous practice we have been doing pretty frequently in the past: changing the client version number when there are changes elsewhere in the repository even though the client code is literally identical.

dshemetov:
On the separate automatic solution for the client: we could probably have another `bumpversion.cfg` somewhere in the client folder. Unfortunately all the clients are in the same folder, so I think that means only one can have automated version bumping with this approach. If we are open to restructuring that folder somewhat (e.g. R, Py, JS folders), then we could automate them separately.

dshemetov:
TODO: look for code in the system that depends on the client package version number being coupled to the repo version. Probably need to check version-specifying config files in this repo (and in other repos).

melange396:
It looks like [our github action for publishing to PyPI](https://github.com/cmu-delphi/delphi-epidata/blob/94a2551315810ffa9399580f4592379c87f40a94/.github/workflows/release-helper.yml#L47-L85) will no-op if the version in `setup.py` isnt changed, thanks to the option `skip_existing: true` (i did a good bit of research to come to that conclusion but didnt actually test it).  Perhaps we should switch the automatic client release GH actions into manually triggered steps anyway?

Going with a new hand-coded client version of `>=5.*` seems excessive to me.  What if we went with `4.1001.10`?  Then the python client version would be the same as the next repository version modulo 1000 on the minor part, and the two versioning paths should be distinct until they realign with v5 (presuming we get to v5 before ~1k revisions of v4).  The one thousand point difference should convey the sentiment that this is a potentially significant upgrade.

Or we could sweep it under the rug by leaving the repository version linked to the python client version and just increment the patch number, and hope that nobody experiences untoward effects.  If some user's code breaks when they update to version `4.1.10`, they will either roll back to the previous version or fix their code.  ...But they will go through the exact same thing if we release the same changes as version `5.0.0` instead.  Strict semantic versioning makes more sense for projects that might introduce a new major or minor version but still perform follow-up updates to the lesser version branch -- for us, regardless of whether we move to `4.1.10` or `4.1001.10` or `5.1.1`, we arent going to fill in any intermediate versions between that and the current `4.1.9`.

In the past, we've done some potentially ["breaking" changes with just a patch-level version bump](https://github.com/cmu-delphi/delphi-epidata/commit/cdf3832cec31d3f96d0b00f62b48f9f5c171d55c) -- thats not exactly justification for doing it again, but it is a precedent.

melange396:
> TODO: look for code in the system that depends on the client package version number being coupled to the repo version.

Im not sure what you mean; I dont think anything DEPENDS on the client version being tied to the rep version, but rather they simply happen to be tied together as a consequence of the client and server code existing in the same place with a shared bumpversion config.

dshemetov:
The situation I was thinking of would be if some setup file somewhere asks for delphi-epidata repo version and buries the assumption that the client has the same version number. And vice versa.

dshemetov (commented):
This is looking good to me, but it seems like we need to make a decision on versioning. My personal take is

1. we should decouple this client from repo version and start doing SemVer proper
2. we take [this Python client release GitHub action](https://github.com/cmu-delphi/delphi-epidata/blob/94a2551315810ffa9399580f4592379c87f40a94/.github/workflows/release-helper.yml#L47-L85) and move it to a separate file that's triggered automatically (a) from the main branch, (b) when `delphi_epidata.py` changes
3. my order of preference for next version is v5.0.0 to v4.2.0 to v4.1001.0 to v4.1.10, but @melange396 has final say here

melange396 (changes requested):

dshemetov:
Should we mark functions downstream of `async_epidata` deprecated too? I'm thinking mostly of [`_async_fetch_epidata` in the Python client](https://github.com/cmu-delphi/covidcast/blob/f9a80e9d675c4f1734ddf69d046160d7c88facdc/Python-packages/covidcast-py/covidcast/covidcast.py#L456), probably don't need to be concerned about `nowcast`.

src/client/packaging/pypi/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 All notable future changes to the `delphi_epidata` python client will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [4.2.0] - 2023-09-27
+
+### Changed
+ 
+- Modify the signatures for several methods: endpoint no longer needs to be specified in the params dict under the "source" key, but becomes a mandatory parameter for the method:
+  - `async_epidata(param_list, batch_size=50)` → `async_epidata(endpoint, param_list, batch_size=50)`
```
dshemetov, line 11: This doesn't seem to match the function definition. Does this need to be updated or does the function need updating?

src/client/packaging/pypi/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 All notable future changes to the `delphi_epidata` python client will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [4.2.0] - 2023-09-27
```
dshemetov, line 6: We're not going to 4.2.0 right? The plan is to just go to 4.1.11?

integrations/server/test_api_keys.py
```diff
@@ -22,13 +21,12 @@ def test_public_route(self):
         public_route = "http://delphi_web_epidata/epidata/version"
         status_codes = set()
         for _ in range(10):
-            status_codes.add(self._make_request(public_route).status_code)
+            status_codes.add(self._make_request("version").status_code)
```
melange396, line 24: still need to delete `public_route` above

src/client/packaging/pypi/CHANGELOG.md
melange396: i have some text for the changelog that im about to push that addresses this

src/client/packaging/pypi/CHANGELOG.md
melange396: i have some text for the changelog that im about to push that addresses this

melange396 (changes requested):

melange396 (approved):
This looks good, but lets wait to merge it until the last few comments in https://github.com/cmu-delphi/covidcast/pull/663 are addressed

dshemetov (approved):

PR #1295, opened 2023-10-02
revert and deprecate async_epidata (+ Remove usage of PHP alias in the Python client)

melange396:
whoops, i wasnt paying attention on the command line and my commit message got slightly mangled, it should read:
"revert `async_epidata()` to previous behavior, and mark deprecated (`filterwarnings()` not fully tested)"

to be added to #1288 

i did not do much testing around the `warnings.filterwarnings()` call, but it is intended to print the deprecation warning for the method only once (on its first use), no matter how `async_epidata()` is called.

### DIFF
--- a/integrations/client/test_delphi_epidata.py
+++ b/integrations/client/test_delphi_epidata.py
@@ -342,12 +342,12 @@ class DelphiEpidataPythonClientTests(CovidcastBase):
     responses = [i[0]["epidata"] for i in test_output]
     # check response is same as standard covidcast call (minus fields omitted by the api.php endpoint),
     # using 24 calls to test batch sizing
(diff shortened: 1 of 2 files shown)


melange396:
@rzats can you make this failing test work?  I think it might be pretty easily fixable if you use `src/common/covidcast_row.py:as_api_compatibility_row_*()` in or around the `.assertEqual()` call.

dshemetov (approved):
Tested `filterwarnings` in a separate python file and it only printed the warning once 👍

rzats (approved):

melange396:
> Tested `filterwarnings` in a separate python file and it only printed the warning once 👍

awesome!!!  thank you @dshemetov !

PR #1303, opened 2023-10-10
Make DataSignal construction more flexible

melange396:
Pass only reflexively valid fields to the `DataSignal` constructor when loading from `db_signals.csv`, instead of using a rigid and brittle set of fields (columns) to exclude.  The system will now easily ignore new columns added to the CSV file, but for which it doesn't yet have a use.  This will make it effortlessly handle changes coming in #1259 (and beyond).

I considered logging a list of unused fields/columns, but that will already be noisy because of the existing `base_is_other` column, and i dont think itll be very helpful.


dmytrotsko (approved):
:+1:

Issue #1129, opened 2023-04-07
get rid of "__test_target__" et al


melange396:
many (if not all) of our integration and unit test files include the setting of a variable "`__test_target__`".  it is also frequently accompanied by a comment of a form similar to "```# py3tester coverage target (equivalent to `import *`)```".  they are no longer needed as we [removed our dependence on py3tester a while ago](https://github.com/cmu-delphi/delphi-epidata/pull/371).  since then, we have even superstitiously perpetuated these code snippets in new test files created.

`py3test` references should also be yeeted from the following files:
```
.github/workflows/ci.yaml
deploy.json
dev/docker/python/setup.sh
dev/local/install.sh
dev/local/setup.cfg
docs/new_endpoint_tutorial.md
docs/epidata_development.md

```

this is a subset of what i mentioned in [a comment on issue 965](https://github.com/cmu-delphi/delphi-epidata/issues/965#issuecomment-1273876351)

melange396:
an additional bonus to this would be that development docker images will be smaller and build faster!

krivard:
does github-deploy-repo run py3tester or pytest?

if it runs pytest then we can slap help-wanted on this; if it runs py3tester that may require a bit more work

dshemetov:
FWIW, I tried removing these before and ran into errors. I don't remember 100%, but I think the errors were only on staging, but not in Docker unit/integration tests, which makes me think some part of github-deploy-repo still uses py3tester

Issue #1042, opened 2022-12-02
Move acquisition deployment off of github-deploy-repo


krivard:
github-deploy-repo works via a webhook. The webhook is written in PHP, which we are slowly scrubbing from Delphi's active tech stack.

This webhook silently broke after the v4 deployment. While we have developed a fix, a better long-term solution would be to change the deployment mechanism for the acquisition system to something that more closely matches the approaches used by all our other active projects.

melange396:
it looks like the way tests are run by `github-deploy-repo` is not completely compatible with the way we do it in our current standard development cycle (see [this slack thread](https://delphi-org.slack.com/archives/C02U0AX8KJR/p1679335829521299) for some context).  if we do not completely move away from `github-deploy-repo` in the near future, we may consider removing its testing step(s), or at least be cognizant of the fact that they can be unreliable.

melange396:
part of moving off of github-deploy-repo would involve doing #1129

melange396:
this may require:
* #986

melange396:
after this is done (or as part of it), we can do
* #1129

melange396:
to sanitize https://github.com/cmu-delphi/github-deploy-repo from this repository, we should...

remove references to "`github-deploy-repo`" from:
https://github.com/cmu-delphi/delphi-epidata/blob/dev/.github/workflows/ci.yaml
https://github.com/cmu-delphi/delphi-epidata/blob/dev/dev/docker/python/setup.sh
https://github.com/cmu-delphi/delphi-epidata/blob/dev/dev/local/install.sh

remove references to "`github_deploy_repo`" from:
https://github.com/cmu-delphi/delphi-epidata/blob/dev/dev/local/setup.cfg

delete these two files:
https://github.com/cmu-delphi/delphi-epidata/blob/dev/deploy.json
https://github.com/cmu-delphi/delphi-epidata/blob/dev/integrations/test_deploy_syntax.py

and then do:
* #1129

korlaxxalrok:
For now we are still heavily reliant on github-deploy-repo to write the various parts of delphi-epidata (via deploy.json) to a filesystem so that the expected import locations and module paths work.

We also still use the webhook functionality, but only in production (and manually deploy to staging).

We can probably make it stop running tests since we do most of that already in Actions CI now, but we'll need to keep it around until we have a different way of getting our bits into an environment that can run them.

[...]

PR #935, opened 2022-06-18
Remove unused undefx depends from CI

dshemetov:
**Prerequisites**:

- [x] Unless it is a documentation hotfix it should be merged against the `dev` branch
- [x] Branch is up-to-date with the branch to be merged with, i.e. `dev`
- [ ] Build is successful
- [ ] Code is cleaned up and formatted

### Summary

Downstream of [this PR](https://github.com/cmu-delphi/operations/pull/70).

Needs:
- [ ] iterate on correct list dependencies
- [ ] consider not depending on `operations`

PR #1308, opened 2023-10-12
cruft removal attempt

melange396:
An attempt to remove a bunch of old cruft from our dependencies, tests, and builds...

Summary:
- Removes download/gitclone and usage of 5 unnecessary repos (`undefx/py3tester`, `undefx/undef-analysis`, `cmu-delphi/github-deploy-repo`, `cmu-delphi/flu-contest`, and `cmu-delphi/nowcast`) from the github actions for CI & performance tests, the `delphi_web_python` `Dockerfile` definition, and the dev env installation script.
- Adds local directory mounts to the test runner commands in the github CI action (which now better matches the dev env Makefile commands).
- Removes testing directive from the `github-deploy-repo` config (aka `./deploy.json`) so we can axe `py3tester` junk.  We run tests in many other places, so it isnt really necessary for g-d-r to do it too.
- Incorporates some file move commands into the `Dockerfile` for the `delphi_web_python` image, removing need for its old separate bash script (`dev/docker/python/setup.sh`)
- Passes tests using its own modified CI definitions!

Results:
- (Hopefully) speeds up building and testing, and reduces filesize/disk footprints of images and installations.
- Local dev envs can probably delete their copies of the 5 extraneous repos.

Caveats:
- Some local dev envs may have some small problems after this, if they use nonstandard procedures (particularly because of changes to the building and structure of the `delphi_web_python` docker image)

Precursor to:
* #1129
* #1042
  * Which then aids in:
    * 938
    * 939

Replaces:
* #935


deploy.json
```diff
@@ -241,10 +241,7 @@
       "dst": "[[stasis]]/integrations/",
       "match": "^.*\\.(py)$",
       "recursive": true
-    },
-
-    "// run unit and coverage tests",
-    {"type": "py3test"}
```
melange396, line 247: removing this will prevent `github-deploy-repo` from running local tests during its deployment process

melange396:
our local wizard, @korlaxxalrok , just successfully tested this on the staging server, so it should be good to go!

.github/workflows/ci.yaml
```diff
@@ -80,11 +53,17 @@ jobs:
 
       - name: Run Unit Tests
         run: |
-          docker run --rm --network delphi-net --env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" delphi_web_python python -m pytest --import-mode importlib repos/delphi/delphi-epidata/tests
+          docker run --rm --network delphi-net \
+                --mount type=bind,source=./repos/delphi/delphi-epidata,target=/usr/src/app/repos/delphi/delphi-epidata,readonly \
+                --mount type=bind,source=./repos/delphi/delphi-epidata/src,target=/usr/src/app/delphi/epidata,readonly \
```
dshemetov, line 58: Comment (non-blocking, probably be separate issue): we should move to having CI use the Makefile commands. It would help keep local testing identical with CI and would remove code duplication.

.github/workflows/ci.yaml
```diff
@@ -80,11 +53,17 @@ jobs:
...
+                --env "SQLALCHEMY_DATABASE_URI=mysql+mysqldb://user:pass@delphi_database_epidata:3306/epidata" --env "FLASK_SECRET=abc" delphi_web_python python -m pytest --import-mode importlib repos/delphi/delphi-epidata/tests
 
       - name: Run Integration Tests
         run: |
-          docker run --rm --network delphi-net delphi_web_python python -m pytest --import-mode importlib repos/delphi/delphi-epidata/integrations
+          docker run --rm --network delphi-net \
+                --mount type=bind,source=./repos/delphi/delphi-epidata,target=/usr/src/app/repos/delphi/delphi-epidata,readonly \
+                --mount type=bind,source=./repos/delphi/delphi-epidata/src,target=/usr/src/app/delphi/epidata,readonly \
```
dshemetov, line 65: Question (non-blocking, separate issue): any reason why unit tests and integrations are separated? In the Makefile both are run together, using that command here would remove this Docker command duplication.

dev/docker/python/Dockerfile
```diff
@@ -4,10 +4,13 @@ RUN apt-get update && apt-get install -y r-base && Rscript -e "install.packages(
...
 RUN ln -s -f /usr/share/zoneinfo/America/New_York /etc/localtime && \
-    chmod -R o+r repos/ && \
-    bash setup.sh && \
-    pip install --no-cache-dir -r repos/delphi/delphi-epidata/requirements.api.txt -r repos/delphi/delphi-epidata/requirements.dev.txt
+    mkdir delphi && \
+    chmod -R o+r delphi/
+
+COPY repos/delphi/operations/src delphi/operations
+COPY repos/delphi/utils/src delphi/utils
+COPY repos/delphi/delphi-epidata/src delphi/epidata
```
dshemetov, line 13: Comment (non-blocking): this COPY command is later overwritten by a runtime mount between the same folders we run tests. Consider getting rid of this.

Thinking out loud (non-blocking): looks like we're dropping the top-level `repos` dir within the Dockerfile. Want to make sure we think through the consequences of this. AFAICT, previously we copied the entire local `repos` folder over int `/usr/src/app/repos` and then moved pieces over to `/usr/src/app/delphi/` and `/usr/src/app/undefx/`. So it looks like we're skipping the intermediate step. So is there anything that still relies on that `repos` folder? The only thing I can find are the Docker testing commands that end with e.g. `python -m pytest --import-mode importlib repos/delphi/delphi-epidata/integrations`. But that is handled by the first mount command in those Docker commands: 

```sh
--mount type=bind,source=./repos/delphi/delphi-epidata,target=/usr/src/app/repos/delphi/delphi-epidata,readonly \
```

So the edit to my first sentence from this sequence of thoughts is: we're not dropping the `repos` dir, it just gets added by a mount at runtime.

Comment (non-blocking, separate issue): it's tough to mentally imagine the directory structure of Dockerfile when runtime mounts are involved (due to non-locality between Docker run command and the Dockerfile). Maybe we should add a comment at the end of this Dockerfile saying something like "in practice, when we run this, these additional directories are present". 

Comment (non-blocking, separate issue): a future future solution to this non-locality is to make a unified `docker-compose.yml` file.

dshemetov (approved):
Have a few tangential comments, some just me thinking through consequences and making sure we don't miss anything in the untangling. If you agree with any of my comments, let me know and I can make separate issues.

But overall, looks good and the fact that it runs in staging means 👍

dev/docker/python/Dockerfile
dshemetov: Docker compose PoC: #1315

PR #1311, opened 2023-10-12
Cleanup files from previous runs of performance test rig

rzats:
### Summary:

Adds an extra step to the performance test runner which cleans up files from previous (now self-hosted) runs.


korlaxxalrok (approved):
LGTM 👍 🚀

PR #1317, opened 2023-10-18
Fix week parsing for date parameters/arguments

melange396:
It looks like #1284 broke some week and week range parsing and was noticed in https://github.com/cmu-delphi/covidcast-indicators/issues/1904.

Heres evidence of the new test failing before i applied the fix:
https://github.com/cmu-delphi/delphi-epidata/actions/runs/6564148018/job/17829935532

This solution is a bit brittle, but ~all of `_params.py` is a mess and needs to be cleaned up, so i think this can fly for now.


dmytrotsko:
@melange396 thx for fixing that. I didn't even know that we have such `time_value` format

dmytrotsko (approved):
:+1:

PR #1326, opened 2023-10-24
Add one-time performance testing workflow

rzats:
### Summary:

Adds a one-time testing workflow that will periodically run the performance scaling test rig and see how the results stack up. This is set to run every 30 minutes on October 25th, and can be safely deleted/reverted afterwards. Similar to #1253.


korlaxxalrok (approved):
LGTM 👍

PR #1328, opened 2023-10-25
feat: add incremental linting and formatting

### LINKED ISSUE #962: Design and enforce linting standards

dshemetov:
fixes #962

### Summary:

* Add `inv lint` which uses [darker](https://github.com/akaihola/darker) and [lint-diffs](https://github.com/AtakamaLLC/lint-diffs/) to format and lint the changes on a given branch (by default relative to origin/dev). 
* Add CI action that will run `inv lint` on PRs.

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,26 +1,37 @@
 [tool.black]
-line-length = 100
+line-length = 120
 target-version = ['py38']
-include = 'server,tests/server'
+
+[tool.darker]
+revision = 'origin/dev...'
+color = true
+isort = true
+
+[tool.isort]
+profile = "black"
+known_third_party = ["pytest"]
 
 [tool.pylint]
-    [tool.pylint.'MESSAGES CONTROL']
-    max-line-length = 100
-    disable = [
-        'logging-format-interpolation',
-        # Allow pytest functions to be part of a class
-        'no-self-use',
-        'too-many-locals',
-        'too-many-arguments',
-        # Allow pytest classes to have one test
-        'too-few-public-methods',
-    ]
+[tool.pylint.main]
+max-line-length = 120
+disable = [
+    'logging-format-interpolation',
+    # Allow pytest functions to be part of a class
+    'no-self-use',
+    'too-many-locals',
+    'too-many-arguments',
+    'too-many-branches',
+    'too-many-statements',
+    # Allow pytest classes to have one test
+    'too-few-public-methods',
+]
+enable = 'useless-suppression'
 
-    [tool.pylint.'BASIC']
-    # Allow arbitrarily short-named variables.
-    variable-rgx = ['[a-z_][a-z0-9_]*']
-    argument-rgx = [ '[a-z_][a-z0-9_]*' ]
-    attr-rgx = ['[a-z_][a-z0-9_]*']
+[tool.pylint.basic]
+# Allow arbitrarily short-named variables.
+variable-rgx = '[A-Za-z_][a-z0-9_]*'
+argument-rgx = '[A-Za-z_][a-z0-9_]*'
+attr-rgx = '[A-Za-z_][a-z0-9_]*'
 
-    [tool.pylint.'DESIGN']
-    ignored-argument-names = ['(_.*|run_as_module)']
+[tool.pylint.design]
+ignored-argument-names = ['(_.*|run_as_module)']
--- a/tasks.py
+++ b/tasks.py
@@ -1,3 +1,8 @@
+"""Repo tasks."""
+
+import pathlib
+
+import requests
 from invoke import task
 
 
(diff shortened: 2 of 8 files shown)


tasks.py
```diff
@@ -26,3 +28,42 @@ def _migrate_file(url: str, filename: str):
 
     _migrate_file(sources_url, "db_sources.csv")
     _migrate_file(signal_url, "db_signals.csv")
+
+
+@task
+def lint(
+    c, incremental=True, format=True, revision="origin/dev...", diff=False, check=False
```
melange396, line 35:
```suggestion
    c, incremental=True, format=False, revision="origin/dev...", diff=True, check=False
```
can we make it so applying changes is not the default?  id rather not have my carefully handcrafted formatting wiped out because i forgot to include a command line argument.

also, are `format` and `diff` mutually exclusive?  do we ever want both to be `True`?  it seems like both `False` would be pointless?

what does `check` do?

tasks.py
```diff
@@ -26,3 +28,42 @@ def _migrate_file(url: str, filename: str):
...
+    if incremental:
+        if format:
+            c.run(f"darker --revision {revision} {diff} {check} .")
+        out = c.run(f"git diff -U0 {revision} | lint-diffs")
+        if out.stdout.strip() != "=== pylint: mine=0, always=0":
+            print(out.stdout)
+            sys.exit(1)
+    else:
+        if format:
+            reponse = input("This will format all files in this repo, continue? [y/N]")
```
melange396, line 64: scary!  nice failsafe :)

pyproject.toml
```diff
@@ -1,26 +1,36 @@
...
-    [tool.pylint.'BASIC']
-    # Allow arbitrarily short-named variables.
-    variable-rgx = ['[a-z_][a-z0-9_]*']
-    argument-rgx = [ '[a-z_][a-z0-9_]*' ]
-    attr-rgx = ['[a-z_][a-z0-9_]*']
+[tool.pylint.basic]
+# Allow arbitrarily short-named variables.
+variable-rgx = '[A-Za-z_][a-z0-9_]*'
+argument-rgx = '[A-Za-z_][a-z0-9_]*'
+attr-rgx = '[A-Za-z_][a-z0-9_]*'
```
melange396, line 33: i reserve the right to complain about these (or lack of them) as we use this more!

melange396 (commented):
cool stuff!  do you have a sample output?

pyproject.toml
dshemetov: Fair! Tbh, as per the comment, I think the main reason for this is we had some mathy code that used mathy variables like `X` or `W` for matrix operations and the linter complained. So we wound up with this essentially YOLO arbitrary names regex.

A possible alternative solution is to upgrade to [Pylint 3.0.0](https://pylint.readthedocs.io/en/latest/whatsnew/3/3.0/index.html#changes-requiring-user-actions):
> The invalid-name message no longer checks for a minimum length of 3 characters by default. (This was an unadvertised commingling of concerns between casing and name length, and users regularly reported this to be surprising.)

I could try bumping pylint in covidcast-indicators and if 3.0.0 passes there, we could probably then pin in both repos.

tasks.py
dshemetov: Thanks! I must confess that this in here because I already shot myself with this footgun xD

tasks.py
dshemetov: In hindsight, yea, good call on the no-auto-format. In personal projects I like format on save, but this totally makes sense for our gradual migration.

- `diff` puts `darker` and `black` into "show me the changes you would make" mode
- `lint-diffs` never formats, just complains (so it's essentially always in `diff` mode)
- `format` controls if `darker` or `black` are run (since they're the formatters)
- so you want `format=True` and `diff=True` if you want both `lint-diffs` and `black` to give you their complaints
- `check` makes `darker` or `black` return a non-zero exit code if any changes would be made (but without `diff` it doesn't return changes)

pyproject.toml
dshemetov: So upgrading pylint in `covidcast-indicators` produces a lot of new errors that will take a while to fix. I think we should just keep both `delphi-epidata` and `covidcast-indicators` at `pylint==2.8.3` for now, since that will allow us to bring this repo up incrementally and that will already be a nice improvement. After a bit of that, we can think about upgrading pylint and addressing the new issues. We can also tighten up these naming conventions in a PR that addresses just naming.

dshemetov:
@melange396 here's a sample output (many months later 😮‍💨)

```py
(venv) ➜  delphi-epidata git:(ds/lint) ✗ inv lint
--- src/client/delphi_epidata.py        2024-05-15 21:34:24.871474 +0000
+++ src/client/delphi_epidata.py        2024-05-15 21:50:15.409996 +0000
@@ -77,11 +77,13 @@
     @retry(reraise=True, stop=stop_after_attempt(2))
     def _request_with_retry(endpoint, params={}):
         """Make request with a retry if an exception is thrown."""
         request_url = f"{Epidata.BASE_URL}/{endpoint}/"
         if Epidata.debug:
-            Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS, auth=Epidata.auth) # add a little comment
+            Epidata.logger.info(
+                "Sending GET request", url=request_url, params=params, headers=_HEADERS, auth=Epidata.auth
+            )  # add a little comment
         if Epidata.sandbox:
             resp = requests.Response()
             resp._content = b'true'
             return resp
         start_time = time.time()
src/client/delphi_epidata.py:82:0: C0301: Line too long (146/120) (line-too-long)
src/client/delphi_epidata.py:316:0: C0325: Unnecessary parens after 'not' keyword (superfluous-parens)
tasks.py:45:0: C0304: Final newline missing (missing-final-newline)

=== pylint: mine=3, always=0
```

PR #1330, opened 2023-10-26
Prevent "308 PERMANENT REDIRECT" via client change

### LINKED ISSUE #1320: prevent "308 PERMANENT REDIRECT" via client change
Recent changes to the python client in #1288 make it so requests are made to endpoints directly, instead of using the `api.php` alias to do the proper routing.  However, the client is not currently appending a trailing slash to these URLs, which prompts a "308 PERMANENT REDIRECT" response from the server.

The client "correctly" follows this up by re-requesting with the trailing slash present, but this wastes time on the client side by doubling the number of requests made, and wastes cycles on the server side.

compare for yourself the difference in the results from the two commands:
`curl -vvvv https://api.delphi.cmu.edu/epidata/covidcast`
and
`curl -vvvv https://api.delphi.cmu.edu/epidata/covidcast/`

Once this fix is complete, it should also automatically get applied to the `covidcast` client as well, as soon as users of that package update their dependencies to include the appropriate `delphi-epidata` release.

rzats:
Closes #1320.

### Summary:

Appends a trailing slash to URLs requested by the Python client, which should prevent an automatic redirect and an extra request to the server.

### Prerequisites:

- [ ] Build is successful

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -70,7 +70,7 @@ class Epidata:
     @retry(reraise=True, stop=stop_after_attempt(2))
     def _request_with_retry(endpoint, params={}):
         """Make request with a retry if an exception is thrown."""
-        request_url = f"{Epidata.BASE_URL}/{endpoint}"
+        request_url = f"{Epidata.BASE_URL}/{endpoint}/"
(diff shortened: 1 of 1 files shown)


melange396 (approved):

melange396:
great, thanks!

PR #1334, opened 2023-11-03
changes to signals in rate-limit exempt list

melange396:
* preemptively adds `nchs-mortality:deaths_covid_incidence_prop` into the ratelimit-exempt signal set
* changes hospital admissions override signal for county-level visualizations to an active/current signal (from a `dsew-cpr` to `hospital-admissions:smoothed_adj_covid19_from_claims`)
* adds a fix to include override signals in the ratelimit-exempt set
* fixes tests to work around changes to exemption status of above signals


dsweber2 (approved):
mostly got stuck down a rabbit hole trying to make the requests for these myself. `smoothed_adj_covid19_from_claims` does have data with a time value of oct 29 (missing Alaska and Montana for some reason), so seems like a decent display candidate. Otherwise I'm trusting the tests

PR #1345, opened 2023-11-10
Switch to Sentry

korlaxxalrok:
### Summary:

We are migrating to Sentry for application observability needs.

The goal is to use it for both local development as well as (potentially) qa/staging/production/etc. environments.

This PR:

Removes and cleans up the New Relic config:

- Removes the New Relic wrapper and package.
- Returns the default application start process.

Adds the necessary Sentry bits:

- Adds the Sentry package for Flask.
- Adds and configures the Sentry SDK with useful defaults for local development.
- Exposes common knobs and levers as env vars for tuning when running in other environments.

Updates the local development setup to make it easy to instrument for Sentry (if desired):

- Updates the local development Makefile:
  - To create a `.env` file if one doesn't already exist. We need to be able to pass in the Sentry DSN (and maybe other config), and putting this into `.env` is best.
  - To start `delphi_web_epidata` with `--env-file` to load vars from `.env`

Adds comments and README updates to help with a basic Sentry setup in:

- `.env.example`
- `epidata_development.md`


requirements.dev.txt
```diff
@@ -21,3 +21,4 @@ selenium==4.7.2
 sqlalchemy-stubs>=0.3
 tenacity==7.0.0
 xlrd==2.0.1
+sentry-sdk[flask]
```
melange396, line 24:
```suggestion
```
you only need this in `requirements.api.txt` -- the only time `requirements.dev.txt` is used, its combined with `requirements.api.txt`: https://github.com/cmu-delphi/delphi-epidata/blob/08de322948ac6d55ad394d7c2fab62cbd257372b/dev/docker/python/Dockerfile#L13

requirements.api.txt
```diff
@@ -18,3 +17,4 @@ structlog==22.1.0
 tenacity==7.0.0
 typing-extensions
 werkzeug==2.2.3
+sentry-sdk[flask]
```
melange396, line 20: would you consider putting this in alphabetical order to assuage my ocd? 😵‍💫

src/server/main.py
```diff
@@ -13,6 +15,18 @@
...
 
+SENTRY_DSN = os.environ.get('SENTRY_DSN')
+if SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        traces_sample_rate=float(os.environ.get('SENTRY_TRACES_SAMPLE_RATE', 1.0)),
+        profiles_sample_rate=float(os.environ.get('SENTRY_PROFILES_SAMPLE_RATE', 1.0)),
+        environment=str(os.environ.get('SENTRY_ENVIRONMENT', 'development')),
+        debug=str(os.environ.get('SENTRY_DEBUG', 'True')),
+        attach_stacktrace=str(os.environ.get('SENTRY_ATTACH_STACKTRACE', 'True'))
```
melange396, line 26: do these options expressly accept strings? if they expect bools, any string except the empty string may evaluate to `True`...

also, is there a performance penalty with the `debug` option?  if so, maybe make it default to False and set the environment variable to True in our Makefile?  otherwise prod needs the option, and it feels better to me to make the development setup responsible for that.  ¯\\\_(ツ)_/¯  your call

melange396 (commented):
a few tiny things, but otherwise looks groovy!

korlaxxalrok:
I wan to work on this a bit more so converting to draft for a little bit.

melange396 (approved):
beautiful!

PR #1354, opened 2023-11-30
OKRF23-54 Add GitHub action to sync labeled issue to Jira board

dmytrotsko:
### Summary:

Added GitHub workflow/action to sync GitHub issue to Jira board when "jira" label is added to issue labels
### Prerequisites:

- [x] Build is successful


.github/actions/sync-issue-with-jira/action.yaml
```diff
@@ -0,0 +1,120 @@
...
+          echo "Not interested in this action, skipping"
+          exit 0
+        fi
+
+        # last one wins
+        echo "NeedsJiraUpdate=true" >> $GITHUB_ENV
+      shell: bash
+
+    - name: "Update jira"
+      if: ${{ env.NeedsJiraUpdate == 'true' }}
```
melange396, line 63: can we use something like this instead of the whole script step above?  if so, we should also preserve some of your comments there, especially the "Issue creation with label will trigger 2 events...".
```suggestion
      if: |
        github.event_name == "issues"
        && ! github.event.issue.pull_request
        && (contains(github.event.issue.labels.*.name, inputs.label) || github.event.label.name == inputs.label)
        && (github.event.action == "labeled" || github.event.action == "unlabeled" || github.event.action == "reopened" || github.event.action == "closed")
        && ( !(github.event.action == "labeled" || github.event.action == "unlabeled") || github.event.label.name == inputs.label )
        # line by line breakdown of above:
        #  - event is "issues"
        #  - and its not a pull request
        #  - and it has the keyword among its existing labels, OR this is an actual keyword [un]label action itself
        #  - and its action is one of "labeled", "unlabeled", "reopened", or "closed"
        #  - and if its action is "labeled" or "unlabeled", the label name matches the keyword
```

.github/workflows/sync-issue-with-jira.yml
```diff
@@ -0,0 +1,12 @@
+name: Sync GitHub issues to Jira
+on: [issues]
+
+jobs:
+  sync-issues:
+    name: Sync issues to Jira
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/sync_with_jira/
```
melange396, line 10: can we do all of this in just one file, or do we need to break it into two?

melange396 (changes requested):
looks pretty good, but i have a few questions...

.github/actions/sync-issue-with-jira/action.yaml
dmytrotsko: I think so, I will test that suggestion and if it works I will add those changes

dshemetov:
Not using Jira anymore

Issue #1355, opened 2023-12-05
Update database drivers


melange396:
We pin the version on 3 database-specific (MySQL) packages.  They are all getting long in the tooth.  Updating them could give us better performance, but might also cause problems or require non-trivial changes:

* https://github.com/cmu-delphi/delphi-epidata/blob/bc7e37590d22c541d87430684e12c747b71c6fd0/requirements.api.txt#L7
  * 1.5y old
  * https://pypi.org/project/mysqlclient/#history

* https://github.com/cmu-delphi/delphi-epidata/blob/bc7e37590d22c541d87430684e12c747b71c6fd0/requirements.dev.txt#L16
  * almost 3y old!
  * https://pypi.org/project/pymysql/#history

* https://github.com/cmu-delphi/delphi-epidata/blob/bc7e37590d22c541d87430684e12c747b71c6fd0/requirements.dev.txt#L13
  * ~4.5y old!!!
  * was: https://pypi.org/project/mysql-connector/#history
  * replaced by: https://pypi.org/project/mysql-connector-python/#history

melange396:
I went looking for places these db packages are actually referenced/used...

`pymysql` seems vestigial and doesnt appear to be used anywhere, and thus should probably be removed.

In all of our python ( https://github.com/search?q=repo%3Acmu-delphi%2Fdelphi-epidata+%2Fmysql%2F+language%3Apython&type=code ), we really only use `mysql-connector`/`mysql-connector-python` aka `mysql.connector`.  This amounts to just the acquisition and test code.

In non-python ( https://github.com/search?q=repo%3Acmu-delphi%2Fdelphi-epidata+%2Fmysql%2F+-language%3Apython&type=code ), the only places we see `mysqlclient` aka `mysqldb` are in requirements files and as a 'dialect' in SQLAlchemy DBURIs.

[...]

PR #1356, opened 2023-12-05
use REPLACE INTO instead of INSERT INTO...UPDATE in covid_hosp acquisition

melange396:
The change from #1224 does not work on production, presumably because of the new superlong INSERT INTO...UPDATE statement that it introduced.  With it, [this `executemany()` call](https://github.com/cmu-delphi/delphi-epidata/blob/bc7e37590d22c541d87430684e12c747b71c6fd0/src/acquisition/covid_hosp/common/database.py#L214) hangs indefinitely and pegs the CPU (this even happens with a batch size of 2).  The covid_hosp tables have [a large number of lengthy column names](https://github.com/cmu-delphi/delphi-epidata/blob/bc7e37590d22c541d87430684e12c747b71c6fd0/src/ddl/covid_hosp.sql#L380-L500), and each gets repeated thrice in the INSERT INTO...UPDATE syntax, making for a ~20k character long SQL statement (and this doesn't even consider the literal values that get inserted).  It stands to reason that we didn't see this problem in our tests because we don't use an obnoxiously long list of sample columns there (perhaps we should).

I have a hunch that this could be a driver issue, so updates there might've remedied the situation, but i didn't want to crawl down that rabbit hole.  We can potentially explore this with https://github.com/cmu-delphi/delphi-epidata/issues/1355.

The fix uses MySQL's REPLACE INTO statement instead, which doesn't require enumerating column names.  When there is a UNIQUE key collision, INSERT INTO...UPDATE updates the columns as defined in the UPDATE clause, while REPLACE INTO deletes the colliding row and inserts the new one.

Also included in this PR is improved logging which lets us differentiate between new and inserted rows, and it adds a stub for removing covid_hosp datasets from the db (which may be helpful for doing patches in the future).

### DIFF
--- a/src/acquisition/covid_hosp/common/database.py
+++ b/src/acquisition/covid_hosp/common/database.py
@@ -159,26 +159,6 @@ class Database:
           (%s, %s, %s, %s, %s, NOW())
       ''', (self.table_name, self.hhs_dataset_id, publication_date, revision, meta_json))
 
-  def remove_issues(self, issue_date):
-    # TODO: this is *VERY* incomplete!  SQL statements are never even evaluated!
-    # delete from metadata table where issue date matches
-    a = f"DELETE FROM `covid_hosp_meta` WHERE dataset_name='{self.table_name}' AND publication_date='{issue_date}'"
-    if self.aggregate_key_cols:
-      # TODO: restrict this to just UNIQUE columns from aggregate keys table?
-      # create (empty) `some_temp_table` like `{self.table_name}_key`
-      b = f"CREATE TABLE some_temp_table AS SELECT {self.aggregate_key_cols} FROM `{self.table_name}_key` WHERE FALSE"
-      # save aggregate keys from what we are about to delete
-      c = f"SELECT {self.aggregate_key_cols} INTO some_temp_table FROM `{self.table_name}` WHERE `{self.publication_col_name}`={issue_date} GROUP BY {self.aggregate_key_cols}"
-      # TODO: combine two SQL queries above into one?
-    # delete from main data table where issue matches
-    d = f"DELETE FROM `{self.table_name}` WHERE `{self.publication_col_name}`={issue_date}"
-    if self.aggregate_key_cols:
-      # delete from saved aggregate keys where the key still exists
-      e = f"DELETE FROM some_temp_table JOIN `{self.table_name}` USING ({self.aggregate_key_cols})"
-      # delete from aggregate key table anything left in saved keys (which should be aggregate keys that only existed in the issue we deleted)
-      f = f"DELETE FROM `{self.table_name}_key` JOIN some_temp_table USING ({self.aggregate_key_cols})"
-      g = "DROP TABLE some_temp_table"
-
   def insert_dataset(self, publication_date, dataframe, logger=False):
     """Add a dataset to the database.
 
--- a/src/acquisition/covid_hosp/common/database.py
+++ b/src/acquisition/covid_hosp/common/database.py
@@ -205,7 +205,7 @@ class Database:
       dataframe.loc[:, csv_name] = dataframe[csv_name].map(self.columns_and_types[csv_name].dtype)
 
     col_names = [f'`{i.sql_name}`' for i in dataframe_columns_and_types + self.additional_fields]
-    value_placeholders = ', '.join(['%s'] * (2 + len(columns))) # extra 2 for `id` and `self.publication_col_name` cols
+    value_placeholders = ', '.join(['%s'] * (2 + len(col_names))) # extra 2 for `id` and `self.publication_col_name` cols
     columnstring = ', '.join(col_names)
     sql = f'REPLACE INTO `{self.table_name}` (`id`, `{self.publication_col_name}`, {columnstring}) VALUES ({value_placeholders})'
     id_and_publication_date = (0, publication_date)
--- a/src/acquisition/covid_hosp/common/database.py
+++ b/src/acquisition/covid_hosp/common/database.py
@@ -244,7 +244,8 @@ class Database:
       if many_values:
         cursor.executemany(sql, many_values)
         rows_affected += cursor.rowcount
-      logger.info('rows affected', count=rows_affected)
+      if logger:
+        logger.info('rows affected', count=rows_affected)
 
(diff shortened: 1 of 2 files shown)


src/acquisition/covid_hosp/common/database.py
```diff
@@ -159,6 +159,26 @@ def insert_metadata(self, publication_date, revision, meta_json, logger=False):
...
+  def remove_issues(self, issue_date):
+    # TODO: this is *VERY* incomplete!  SQL statements are never even evaluated!
+    # delete from metadata table where issue date matches
+    a = f"DELETE FROM `covid_hosp_meta` WHERE dataset_name='{self.table_name}' AND publication_date='{issue_date}'"
+    if self.aggregate_key_cols:
+      # TODO: restrict this to just UNIQUE columns from aggregate keys table?
+      # create (empty) `some_temp_table` like `{self.table_name}_key`
+      b = f"CREATE TABLE some_temp_table AS SELECT {self.aggregate_key_cols} FROM `{self.table_name}_key` WHERE FALSE"
+      # save aggregate keys from what we are about to delete
+      c = f"SELECT {self.aggregate_key_cols} INTO some_temp_table FROM `{self.table_name}` WHERE `{self.publication_col_name}`={issue_date} GROUP BY {self.aggregate_key_cols}"
```
dshemetov, line 171: Should this be INSERT INTO? [SELECT INTO](https://dev.mysql.com/doc/refman/8.0/en/select-into.html) seems to only store to vars or files.

dshemetov:
```suggestion
      c = f"INSERT {self.aggregate_key_cols} INTO some_temp_table FROM `{self.table_name}` WHERE `{self.publication_col_name}`={issue_date}"
```

Should this be INSERT INTO? [SELECT INTO](https://dev.mysql.com/doc/refman/8.0/en/select-into.html) seems to only store to vars or files. Not sure if we need the GROUP BY either.

src/acquisition/covid_hosp/common/database.py
```diff
@@ -159,6 +159,26 @@ def insert_metadata(self, publication_date, revision, meta_json, logger=False):
...
+      # TODO: combine two SQL queries above into one?
+    # delete from main data table where issue matches
+    d = f"DELETE FROM `{self.table_name}` WHERE `{self.publication_col_name}`={issue_date}"
+    if self.aggregate_key_cols:
+      # delete from saved aggregate keys where the key still exists
+      e = f"DELETE FROM some_temp_table JOIN `{self.table_name}` USING ({self.aggregate_key_cols})"
+      # delete from aggregate key table anything left in saved keys (which should be aggregate keys that only existed in the issue we deleted)
+      f = f"DELETE FROM `{self.table_name}_key` JOIN some_temp_table USING ({self.aggregate_key_cols})"
+      g = "DROP TABLE some_temp_table"
```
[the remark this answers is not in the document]
melange396: good suggestion, but this is just a work-in-progress for now...  as you probably noticed, these are all just strings and the function doesnt actually accomplish anything.  im gonna remove the method and save it for later.

dshemetov (approved):
No notes on the main code, I read through it line by line and it looks correct to me.

melange396:
This indeed is/was a driver issue...  It was corrected in `mysql-connector-python` (AKA `mysql.connector` in the python code) version `8.0.33`; see the [relevant text in the release notes](https://dev.mysql.com/doc/relnotes/connector-python/en/news-8-0-33.html#:~:text=Executing%20executemany()%20with%20insert%20statements%20that%20used%20ON%20DUPLICATE%20KEY%20UPDATE%20demonstrated%20poor%20performance%3B%20this%20was%20improved%20by%20fixing%20the%20associated%20regular%20expressions%20that%20match%20INSERT%20statements.%20(Bug%20%2399575%2C%20Bug%20%2331355895)), the [filed bug](https://bugs.mysql.com/bug.php?id=99575), and its [fix](https://github.com/mysql/mysql-connector-python/commit/49b861e31b75417dae55248d2c4f8667e69af1d3).

The prod and staging "automation" servers currently run `mysql-connector-python` version `8.0.21` from 2020 (and they almost certainly also did when #1224 went live and broke).  That version may have been committed to by ansible when it evaluated something like [this](https://github.com/cmu-delphi/delphi-admin/blob/764342f1699a8ce9241a7adce2153626e765dfcf/ansible/build-automation-aws.yaml#L210) or [this](https://github.com/cmu-delphi/delphi-admin/blob/764342f1699a8ce9241a7adce2153626e765dfcf/ansible/roles/delphi.pyenv/tasks/pyenv-python-packages.yaml#L8) for these machines, or it could have even been installed or updated manually by someone.  New versions of our own code are kept up-to-date on these machines by our old friend https://github.com/cmu-delphi/github-deploy-repo (🧄), which does not appear to have any support for maintaining dependencies (much less their versions).

The much older version `2.2.9` of the `mysql-connector` package (still AKA `mysql.connector` in code) is [pinned in our development requirements](https://github.com/cmu-delphi/delphi-epidata/blob/9c5ef79c6d6bf0eec84657ceacc4c322f1cd0089/requirements.dev.txt#L13).  That version does not seem to be affected by the `executemany(INSERT...ON DUPLICATE)` bug, which added to the difficulty of isolating the problematic behavior.

We should try to keep our production and development environments more in sync so that we don't run into inconsistent and confusing situations like this again.  We have talked about this a number of times before for various reasons, but i don't think we ever came to a consensus.  Some possibilities for addressing it include:
- Run our production automation jobs out of our prebuilt docker containers instead of on "bare hardware".  They will then have the same versions of requirements/dependencies as are in dev.  This might be "the right way to do things" but it sounds complicated and is probably a longer-term project.
- Add a `pip` step like [this](https://github.com/cmu-delphi/delphi-epidata/blob/9c5ef79c6d6bf0eec84657ceacc4c322f1cd0089/dev/docker/python/Dockerfile#L13) to the end of any github-deploy-repo runs.  This sounds easy(ish?) to do, though there could be unforeseen consequences.  It is also a temporary band-aid, since g-d-r is already in the crosshairs in #1042.
- Maybe get ansible to do it?

I will probably turn much of the text of this comment into a GH issue of its own, too. cc: @korlaxxalrok

PR #1413, opened 2024-03-27
Add `min_issue` to metadata

### LINKED ISSUE #232: Add min_issue to covidcast metadata endpoint

nmdefries:
Closes https://github.com/cmu-delphi/delphi-epidata/issues/232

### Summary:
Calculate and include `min_issue` in metadata.

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/docs/api/covidcast_meta.md
+++ b/docs/api/covidcast_meta.md
@@ -38,6 +38,7 @@ None required.
 | `epidata[].max_value`     | maximum value                                                       | float            |
 | `epidata[].mean_value`    | mean of value                                                       | float            |
 | `epidata[].stdev_value`   | standard deviation of value                                         | float            |
+| `epidata[].min_issue`     | oldest date data was issued (e.g., 20200710)                        | integer          |
 | `epidata[].max_issue`     | most recent date data was issued (e.g., 20200710)                   | integer          |
 | `epidata[].min_lag`       | smallest lag from observation to issue, in `time_type` units        | integer          |
 | `epidata[].max_lag`       | largest lag from observation to issue, in `time_type` units         | integer          |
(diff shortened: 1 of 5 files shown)


melange396:
I know why that test is failing, and its related to a semi-bug that i thought i had documented already (but i apparently didn't).  We can work around it with a small change to the test -- i will show you tomorrow or monday.

melange396:
The test failure stems from this issue i just created: https://github.com/cmu-delphi/delphi-epidata/issues/1427

You can fix this PR in the meantime by changing the way rows are created and inserted by the failing test...  Right now, [it creates the rows by a list comprehension and then inserts them all at once](https://github.com/cmu-delphi/delphi-epidata/blob/a2dcfad838870ce05da45d4d3d9138e5cb60a36d/integrations/client/test_delphi_epidata.py#L365-L373).  I think that if you move the call to `self._insert_rows()` inside of one or both of those `for` loops, it should do the trick (🤞).

nmdefries:
Changed that [here](https://github.com/cmu-delphi/delphi-epidata/pull/1413/commits/5f64b54eb0d448a2e6415a992ef546723f77f73c). Based on the previous failure, I think that's the only spot that needs to change.

docs/api/covidcast_meta.md
```diff
@@ -51,19 +52,22 @@ https://api.delphi.cmu.edu/epidata/covidcast_meta/
...
-      "min_time": 20200201
+      "min_time": 20200201,
+      "max_time": 20240321,
+      "num_locations": 2586,
+      "min_value": 0.0,
+      "max_value": 87.670832,
+      "mean_value": 2.4972178,
+      "stdev_value": 3.6433783,
+      "last_update": 1711412545,
+      "max_issue": 20240325,
```
melange396, line 68:
```suggestion
      "min_issue": 20200507,
      "max_issue": 20240325,
```
(i looked this up in the db)

integrations/client/test_delphi_epidata.py
```diff
@@ -283,15 +283,14 @@ def test_covidcast_meta(self):
...
-    ]
-    self._insert_rows(rows)
+    for i in range(3):
+      for t in range(3):
+        row = CovidcastTestRow.make_default_row(
+          time_value=DEFAULT_TIME_VALUE + t,
+          issue=DEFAULT_ISSUE + i,
+          value=t*10 + i
+        )
+        self._insert_rows([row])
```
melange396, line 293: the test is failing because the data points all get re-issued, so the minimum issue across all of the "latest" data is actually that most recent issue.  we can add another (unique) day that doesnt have its issue updated.

i think this should work:
```suggestion
        self._insert_rows([row])
    self._insert_rows([CovidCastTestRow.make_default_row(
      time_value=DEFAULT_TIME_VALUE-1,
      issue=DEFAULT_ISSUE,
      value=12
    )])
```
and then youll have to change below to include
```py
      min_time=DEFAULT_TIME_VALUE-1,
```
and probably a new `stdev_value` too

nmdefries: Thanks for the various fixes. I've been struggling with this test

melange396:
In its current state, this isnt going to be entirely accurate in many cases (though possibly not all cases) ; to get the guaranteed absolute "first" issue, we will need to query the `full` table instead of the `latest` as it is now, which will require an additional or more complicated SQL statement.

dshemetov:
Closing as stale

PR #1418, opened 2024-04-18
Add inspection modes to Python client library

### LINKED ISSUE #1321: Add "inspection" mode to client libraries
Add an inspection mode (like a debug mode) to client libraries so we can "see" (and/or log) what exact http[s] requests are made during its operation.

This can help identify redundant/repeated requests, or shed light on which might be taking unexpected amounts of time to return.  It can also elucidate behaviors in [the `covidcast` client](https://github.com/cmu-delphi/covidcast/blob/d2aa655c01e8cc411202fad82e89f6ad9f091d0d/Python-packages/covidcast-py/covidcast/covidcast.py#L417-L463), where a single method call may be decomposed into multiple http requests.

A simple implementation of this could just print info for each request as it is made (URL, method, headers, response time, response bytes/rows, etc).  An additional deeper feature could print this info, BUT return empty/dummy data and not actually make the request.  The latter could be useful for testing long-running or expensive data retrieval scripts without making the user wait for results or putting load on the server (obviously, this may not be very useful in instances when later query parameters are dependent on earlier query results).

rzats:
Closes #1321.

### Summary:

Adds two debug modes to the Python client library **(JS and R clients TBD)**:

- `debug` logs info about HTTP requests and responses.
- `sandbox` prevents any HTTP requests from actually executing, allowing users to test a long-running or expensive script without any actual server load.

### Prerequisites:

- [X] Build is successful

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -79,18 +79,24 @@ class Epidata:
         """Make request with a retry if an exception is thrown."""
         request_url = f"{Epidata.BASE_URL}/{endpoint}/"
         if Epidata.debug:
-            Epidata.logger.info(f"Sending GET request to URL: {request_url} | params: {params} | headers: {_HEADERS}")
+            Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS)
         if Epidata.sandbox:
-            return
+            resp = requests.Response()
+            resp._content = b'true'
+            return resp
         start_time = time.time()
         req = requests.get(request_url, params, auth=Epidata.auth, headers=_HEADERS)
         if req.status_code == 414:
             if Epidata.debug:
-                Epidata.logger.info(f"Received {req.status_code} response ({len(req.content)} bytes) in {(time.time() - start_time):.3f} seconds")
-                Epidata.logger.info(f"Sending POST request to URL: {request_url} | params: {params} | headers: {_HEADERS}")
+                Epidata.logger.info("Received 414 response, retrying as POST request", url=request_url, params=params, headers=_HEADERS)
             req = requests.post(request_url, params, auth=Epidata.auth, headers=_HEADERS)
         if Epidata.debug:
-            Epidata.logger.info(f"Received {req.status_code} response ({len(req.content)} bytes) in {(time.time() - start_time):.3f} seconds")
+            Epidata.logger.info(
+                "Received response",
+                status_code=req.status_code,
+                len=len(req.content),
+                time=round(time.time() - start_time, 4)
+            )
         # handle 401 and 429
         req.raise_for_status()
         return req
@@ -105,8 +111,6 @@ class Epidata:
         """
         try:
             result = Epidata._request_with_retry(endpoint, params)
-            if Epidata.sandbox:
-                return
         except Exception as e:
             return {"result": 0, "message": "error: " + str(e)}
         if params is not None and "format" in params and params["format"] == "csv":
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -10,11 +10,14 @@ Notes:
 
 # External modules
 import requests
+import time
 import asyncio
 from tenacity import retry, stop_after_attempt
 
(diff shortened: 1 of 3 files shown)


src/client/delphi_epidata.py
```diff
@@ -71,9 +78,19 @@ def _list(values):
     def _request_with_retry(endpoint, params={}):
         """Make request with a retry if an exception is thrown."""
         request_url = f"{Epidata.BASE_URL}/{endpoint}/"
+        if Epidata.debug:
+            Epidata.logger.info(f"Sending GET request to URL: {request_url} | params: {params} | headers: {_HEADERS}")
```
melange396, line 82: should probably log `Epidata.auth` here too since its used in the request

src/client/delphi_epidata.py
```diff
@@ -71,9 +78,19 @@ def _list(values):
...
+        if Epidata.sandbox:
+            return
```
melange396, line 84: how about returning a dummy `Response` object (since thats whats expected downstream)?  this one should be minimal but mostly valid json:
```suggestion
            resp = requests.Response()
            resp._content = b'true'
            return resp
```

src/client/delphi_epidata.py
```diff
@@ -71,9 +78,19 @@ def _list(values):
...
+        start_time = time.time()
         req = requests.get(request_url, params, auth=Epidata.auth, headers=_HEADERS)
         if req.status_code == 414:
+            if Epidata.debug:
+                Epidata.logger.info(f"Received {req.status_code} response ({len(req.content)} bytes) in {(time.time() - start_time):.3f} seconds")
+                Epidata.logger.info(f"Sending POST request to URL: {request_url} | params: {params} | headers: {_HEADERS}")
```
melange396, line 90: one log message should be sufficient
```suggestion
                Epidata.logger.info(f"Received 414 response, retrying as POST request to URL: {request_url} | params: {params} | headers: {_HEADERS}")
```

src/client/delphi_epidata.py
```diff
@@ -88,6 +105,8 @@ def _request(endpoint, params={}):
         """
         try:
             result = Epidata._request_with_retry(endpoint, params)
+            if Epidata.sandbox:
+                return
```
melange396, line 109:
```suggestion
```
if ._request_with_retry() returns a proper Response object, you can hopefully remove this?

melange396 (changes requested):
Pretty nice!

Could you, instead of or in addition to being in the main log messages, put all the variables reported into their own keyword arguments?  That will allow for easier post-processing, among other things.

src/client/delphi_epidata.py
dshemetov: suggest(non-blocking): since we're using [structlog](https://www.structlog.org/en/stable/getting-started.html#building-a-context), you can feed params in like a dict and it will handle the formatting, like
```py
Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS)
```

dshemetov (dismissed):
Small suggestion about log message formatting, otherwise looks good 👍

dshemetov (dismissed):
lgtm

src/client/delphi_epidata.py
```diff
@@ -71,9 +78,25 @@ def _list(values):
     def _request_with_retry(endpoint, params={}):
         """Make request with a retry if an exception is thrown."""
         request_url = f"{Epidata.BASE_URL}/{endpoint}/"
+        if Epidata.debug:
+            Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS)
```
melange396, line 82:
```suggestion
            Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS, auth=Epidata.auth)
```

melange396 (dismissed):
add in the `auth` logging, and then i think this is done!

rzats:
@melange396 @dshemetov re-requesting review as the latest commit invalidated the "approved" checkmarks :)

Issue #1386, opened 2024-02-08
Log "Referer" header to help track tool usage


melange396:
Extract and record the ["`Referer` [sic]" HTTP request header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) (or, in its absence, the ["Origin" header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin)) to help us better and more easily measure and analyze the usage of our web application tools (like the COVIDcast Dashboard, for instance).

This can be accomplished in the server code [here](https://github.com/cmu-delphi/delphi-epidata/blob/55fb756e3b3775da908feebf6a8c001aa5d19c44/src/server/_common.py#L117-L118) and [here](https://github.com/cmu-delphi/delphi-epidata/blob/55fb756e3b3775da908feebf6a8c001aa5d19c44/src/server/_common.py#L157-L158).

To get more granularity and thus utility out of this, we may also need to change the [Referrer-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) on our webapps (i have not investigated enough to know if this is actually necessary, but i strongly suspect that it is).

melange396:
Lets keep this open for now, until we determine if we have to (and how to) deal with the "Referrer-Policy" stuff

PR #1429, opened 2024-05-09
Log HTTP referer and origin

rzats:
Related: #1386.

### Summary:

Logs the HTTP Referer (or referrer) header as `req_referrer` and Origin header as `req_origin`, helping us track where requests come from. The easiest way to test this is via Postman:

Screenshot 2024-05-09 at 14 53 50

This generates the following log string:

```
{"method": "GET", "url": "http://localhost:10080/epidata/signal_dashboard_status/", "form_args": {}, "req_length": null, "remote_addr": "172.18.0.1", "real_remote_addr": "172.18.0.1", "user_agent": "PostmanRuntime/7.30.1", "api_key": null, "user_id": "None", "req_referrer": "https://test.com/test", "event": "Received API request", "logger": "server_api", "level": "info", "pid": 8, "timestamp": "2024-05-09T11:28:19.206454Z"}
```
```
{ ... "req_referrer": "https://test.com/test" ... }
```

An automated version of this test has also been included. If no referrer is present, the HTTP `Origin` is used instead.

As mentioned in #1386, we _might_ also need to modify the referrer policy across our webapps in order to properly set the relevant fields. The [default referrer policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) is:
```
Send the origin, path, and querystring when performing a same-origin request. For cross-origin requests send the origin (only) when the protocol security level stays same (HTTPS→HTTPS). Don't send the [Referer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) header to less secure destinations (HTTPS→HTTP).
```

Since the API endpoint is HTTPS, for some webapps these headers are likely to be already set out of the box, but further testing will still be necessary.

### Prerequisites:

- [ ] Build is successful

### DIFF
--- a/src/server/_common.py
+++ b/src/server/_common.py
@@ -68,7 +68,7 @@ def log_info_with_request(message, **kwargs):
         remote_addr=request.remote_addr,
         real_remote_addr=get_real_ip_addr(request),
         user_agent=request.user_agent.string,
-        req_referrer=request.referrer or request.origin,
+        referrer=request.referrer or request.origin,
         api_key=resolve_auth_token(),
         user_id=(current_user and current_user.id),
         **kwargs
@@ -115,7 +115,6 @@ def before_request_execute():
     user = current_user
     api_key = resolve_auth_token()
 
-    # TODO: replace this next call with: log_info_with_request("Received API request")
     log_info_with_request("Received API request")
 
     if not _is_public_route() and api_key and not user:
--- a/src/server/_common.py
+++ b/src/server/_common.py
@@ -125,7 +125,9 @@ def before_request_execute():
         real_remote_addr=get_real_ip_addr(request),
         user_agent=request.user_agent.string,
         api_key=api_key,
-        user_id=(user and user.id)
+        user_id=(user and user.id),
+        req_referrer=request.referrer,
+        req_origin=request.environ.get('HTTP_ORIGIN', '')
     )
 
     if not _is_public_route() and api_key and not user:
@@ -171,6 +173,8 @@ def after_request_execute(response):
         response_status=response.status,
         content_length=response.calculate_content_length(),
         elapsed_time_ms=total_time,
+        req_referrer=request.referrer,
+        req_origin=request.environ.get('HTTP_ORIGIN', '')
     )
     return response
 
--- a/src/server/_common.py
+++ b/src/server/_common.py
@@ -68,6 +68,7 @@ def log_info_with_request(message, **kwargs):
         remote_addr=request.remote_addr,
         real_remote_addr=get_real_ip_addr(request),
         user_agent=request.user_agent.string,
+        req_referrer=request.referrer or request.origin,
         api_key=resolve_auth_token(),
         user_id=(current_user and current_user.id),
         **kwargs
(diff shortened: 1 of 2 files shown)


dshemetov (approved):

src/server/_common.py
```diff
@@ -125,7 +125,9 @@ def before_request_execute():
         real_remote_addr=get_real_ip_addr(request),
         user_agent=request.user_agent.string,
         api_key=api_key,
-        user_id=(user and user.id)
+        user_id=(user and user.id),
+        req_referrer=request.referrer,
+        req_origin=request.environ.get('HTTP_ORIGIN', '')
```
melange396, line 130: - Can you move this to right underneath `user_agent=...` (since they come from the same place (request headers) and because they both identify client context)?

- Why not just use [`request.origin`](https://flask.palletsprojects.com/en/3.0.x/api/#flask.Request.origin)?  Presumably so you can default to the empty string, but why prefer that over `None`?

- We dont need to keep both values; if both are provided, refer[r]er should be a superstring of origin.
```suggestion
        refer_origin=request.referrer or request.origin,
```

melange396 (changes requested):
Please also add to the [`log_info_with_request()` method](https://github.com/cmu-delphi/delphi-epidata/blob/rzatserkovnyi/referer-origin/src/server/_common.py#L60C5-L60C23)!

tests/server/test_validate.py
```diff
@@ -60,3 +61,19 @@ def test_require_any(self):
         with self.subTest("one options given with is empty but ok"):
             with app.test_request_context("/?abc="):
                 self.assertTrue(require_any(request, "abc", empty=True))
+
+    def test_origin_headers(self):
+        with self.subTest("referer and origin"):
```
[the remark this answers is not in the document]
rzats: I did end up having a few different tests!

src/server/_common.py
rzats: Fixed!

src/server/_common.py
```diff
@@ -115,18 +116,7 @@ def before_request_execute():
...
-        url=request.url,
-        form_args=request.form,
-        req_length=request.content_length,
-        remote_addr=request.remote_addr,
-        real_remote_addr=get_real_ip_addr(request),
-        user_agent=request.user_agent.string,
-        api_key=api_key,
-        user_id=(user and user.id)
-    )
+    log_info_with_request("Received API request")
```
melange396, line 119: nice!  (but plz remove TODO above)

src/server/_common.py
```diff
@@ -68,6 +68,7 @@ def log_info_with_request(message, **kwargs):
         remote_addr=request.remote_addr,
         real_remote_addr=get_real_ip_addr(request),
         user_agent=request.user_agent.string,
+        req_referrer=request.referrer or request.origin,
```
melange396, line 71: can we just call this `referrer`?
```suggestion
        referrer=request.referrer or request.origin,
```

tests/server/test_validate.py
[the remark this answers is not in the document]
melange396: awesome

melange396 (changes requested):
just a couple quick changes and we should be good :)

dshemetov (approved):

melange396 (approved):

PR #1436, opened 2024-05-14
Use bumpversion for epidata client

### LINKED ISSUE #1431: Fix python client's user-agent version
Change the way the python client determines its own version number.

The current implementation uses `importlib` to [get the "installed version" of the package and set it in the user-agent string](https://github.com/cmu-delphi/delphi-epidata/blob/a2dcfad838870ce05da45d4d3d9138e5cb60a36d/src/client/delphi_epidata.py#L22-L30).  This is apparently fragile enough that is misreporting its version number to the API server (in a number of places on our own systems, no less!).  Maybe it [worked better in the past](https://github.com/cmu-delphi/delphi-epidata/pull/145/files) before it was [updated to stop using `pkg_resources`](https://github.com/cmu-delphi/delphi-epidata/pull/1363).  ¯\\\_(ツ)_/¯ 

On a single one of our automation servers, different jobs make requests that announce user-agents of both `delphi_epidata/0.0.11 (Python)` and `delphi_epidata/0.0.12 (Python)` (among others!).  To make it even more ridiculous, the addition of the "` (Python)`" suffix didn't come around until version 4.1.0.  This does suggest the possibility of a misconfiguration of some sort on the machine, but the way the client is implemented makes that even more difficult to track down.

Unless someone has a better idea, we can keep something like the existing `try` structure and attempt to get [the version from the package](https://github.com/cmu-delphi/delphi-epidata/blob/a2dcfad838870ce05da45d4d3d9138e5cb60a36d/src/client/packaging/pypi/delphi_epidata/__init__.py#L4) but fall back to the "0.script" text if it doesnt work, or we could just be lazy and add `delphi_epidata.py` to [the bumpversion config](https://github.com/cmu-delphi/delphi-epidata/blob/dev/.bumpversion.cfg).

rzats:
Closes #1431.

### Summary:

Updates the versioning in the Python client to the same method used in the JS and R clients. With this method, the version is set to a simple hardcoded string; however, during a new Delphi release, GitHub Actions will run [bump2version](https://pypi.org/project/bump2version/) during the [release process](https://github.com/cmu-delphi/delphi-epidata/blob/dev/.github/workflows/release-helper.yml#L31-L33) and automatically update this string in the code.

This can be replicated in a local environment; the following code should bump the version in the Python client (as well as the `.bumpversion.cfg` file itself, and 7 other files) to 4.1.21:
```
python3 -m pip install bump2version
bump2version --current-version 4.1.20 patch
```

### Prerequisites:

- [x] Build is successful


melange396 (changes requested):
We are already duplicating the version number twice in this package, once in the `__init__.py` and once in the `setup.py`, which is kinda ugly.  i would feel better if we could avoid adding a third copy of it.  What if we keep the literal in `delphi_epidata.py` and make https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/client/packaging/pypi/delphi_epidata/__init__.py and just look like:
```py
from .delphi_epidata import Epidata, __version__

name = "delphi_epidata"
```

melange396:
and dont forget to add to `CHANGELOG.md`!

src/client/delphi_epidata.py
```diff
@@ -15,17 +15,10 @@
...
 from delphi.epidata.common.logger import get_structured_logger
 
-# Obtain package version for the user-agent. Uses the installed version by
-# preference, even if you've installed it and then use this script independently
-# by accident.
-try:
-    _version = version("delphi-epidata")
-except PackageNotFoundError:
-    _version = "0.script"
+_version = "4.1.20"
```
melange396, line 21: to agree with the change in `__init__.py`:
```suggestion
__version__ = "4.1.20"
```
(and then change the other references to this variable as well)

.bumpversion.cfg
```diff
@@ -9,6 +9,8 @@ tag = False
 
 [bumpversion:file:src/client/delphi_epidata.R]
 
+[bumpversion:file:src/client/delphi_epidata.py]
```
melange396, line 12: you can remove the line for `__init__.py` a few lines below since it doesnt have the version number hardcoded in it any more

melange396 (changes requested):

src/client/delphi_epidata.py
```diff
@@ -15,19 +15,12 @@
...
 from delphi.epidata.common.logger import get_structured_logger
 
-# Obtain package version for the user-agent. Uses the installed version by
-# preference, even if you've installed it and then use this script independently
-# by accident.
-try:
-    _version = version("delphi-epidata")
-except PackageNotFoundError:
-    _version = "0.script"
+_version_ = "4.1.20"
```
melange396, line 21: still doesnt match whats in `__init__.py`!
```suggestion
__version__ = "4.1.20"
```
the convention is for this to be a "dunder", as seen in the good ol:
```py
if __name__ == '__main__':
```

melange396 (changes requested):

Issue #1296, opened 2023-10-02
Python client -- issues with version coupling and automatic releases


melange396:
The version of the `delphi_epidata` python client is [bound to](https://github.com/cmu-delphi/delphi-epidata/blob/b5128f09779de310ccd775a8a83edb5205372291/.bumpversion.cfg#L14-L16) the version of the repository.  This means that a new version of the client is [automatically published](https://github.com/cmu-delphi/delphi-epidata/blob/94a2551315810ffa9399580f4592379c87f40a94/.github/workflows/release-helper.yml#L47-L85) to [PyPI](https://pypi.org/project/delphi-epidata/) whenever we do a release of `delphi-epidata`, even if the client code has not actually changed.  This can confuse or annoy users, as it might encourage them to do frequent but unnecessary upgrades.  There are also concerns regarding what to do if we make changes to the client interface that might break users' existing workflows -- in such a case, [Semantic Versioning](https://semver.org/) necessitates incrementing the "major" version number, but this may be undesirable to apply to the entire repository.

We can address this in a number of ways:
- Uncouple versions between the two, which lets us number the client version however we need or want.  This would involve removing the client from the existing "`bumpversion`" configuration, and would require creating a different release process (either automatic or manual -- automatic being more work up front, while manual requires remembering to perform some number of extra steps when changing the client).
- Keep the versions coupled, but only do the PyPI release if/when the client code actually changes (again, either automatic or manual with the same caveats).  Changes to the client (other than just version number) would be found in the file https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/client/delphi_epidata.py and anything under the directory https://github.com/cmu-delphi/delphi-epidata/tree/dev/src/client/packaging/pypi .  I suggest this path, at least initially.

This discussion came up in #1288.  Additional points are made and details listed in that PR, starting with comment https://github.com/cmu-delphi/delphi-epidata/pull/1288#issuecomment-1731792278 .

Similar concerns may exist for the other clients in this repo.

melange396:
We are going to do the second bullet point above, as it integrates well with checking against the most current PyPI release as mentioned in https://github.com/cmu-delphi/delphi-epidata/issues/1281#issuecomment-2104903462

rzats:
@melange396 one quick question before I create a PR for this!

[...]

PR #1456, opened 2024-05-29
One-time version check

### LINKED ISSUE #1281: client version checking
have the [python, r, and javascript clients](https://github.com/cmu-delphi/delphi-epidata/tree/dev/src/client) check their version against the [current running version on the API server](https://github.com/cmu-delphi/delphi-epidata/blob/9954505be2e2754a4bf7c887fab41ad1a74931cc/src/server/main.py#L56-L58), and print a message informing the user that an update is available that may include fixes or new functionality.

a downside to this is that the version number is incremented ***MUCH*** more frequently than changes are made to the client code, so we may often be prompting updates that are pointless.

rzats:
Closes #1281

### Summary:

Adds a one-time check to the Epidata client, which logs a warning when the current "remote" server version (https://api.delphi.cmu.edu/epidata/version) does not match the client version.

_This is limited to debug mode for now but can be easily extended to run by default._

Also adds a test to verify this behavior.

### Prerequisites:

- [x] Build is successful


melange396:
Let's do #1296 first and then you can revisit this

rzats:
@melange396 since #1296 is merged, we can move forward with this PR; lmk if the current iteration looks good!

src/client/delphi_epidata.py
```diff
@@ -102,6 +103,20 @@ def _request(endpoint, params={}):
         capabilities, but fall back to POST if the request is too
         long and returns a 414.
         """
+        # One-time check to verify the client version is up to date.
+        if Epidata.debug and not Epidata.sandbox and not Epidata.version_check:
+            try:
+                server_version = Epidata._request_with_retry("version").json()
+                if server_version["version"] != Epidata.client_version:
```
dshemetov, line 110: Now that we've decoupled the server version from the client version, this check won't be very informative - the versions will often not match, but that won't mean that the client isn't up to date. We need to find another way to query the latest client version, I think.

melange396: https://github.com/cmu-delphi/delphi-epidata/issues/1281#issuecomment-2104903462

src/client/delphi_epidata.py
```diff
@@ -102,6 +103,20 @@ def _request(endpoint, params={}):
...
+            try:
+                server_version = requests.get('https://pypi.org/pypi/delphi-epidata/json').json()['info']['version']
+                if server_version != Epidata.client_version:
+                    Epidata.logger.warn(
+                        "Client version does not match server version",
+                        client_version=Epidata.client_version,
+                        server_version=server_version
+                    )
+            except Exception as e:
+                Epidata.logger.warn("Could not get current server version", exception=str(e))
```
dshemetov, line 117:
```suggestion
                latest_version = requests.get('https://pypi.org/pypi/delphi-epidata/json').json()['info']['version']
                if latest_version != Epidata.client_version:
                    Epidata.logger.warn(
                        "Client version not up to date",
                        client_version=Epidata.client_version,
                        latest_version=latest_version
                    )
            except Exception as e:
                Epidata.logger.warn("Error getting latest client version", exception=str(e))
```

dshemetov (changes requested):
Do yall only need this for internal use? If so, then this is good to go. If not, then we should not hide the version check behind a debug flag.

src/client/delphi_epidata.py
```diff
@@ -107,6 +108,20 @@ def _request(endpoint, params={}):
         capabilities, but fall back to POST if the request is too
         long and returns a 414.
         """
+        # One-time check to verify the client version is up to date.
+        if Epidata.debug and not Epidata.sandbox and not Epidata.version_check:
+            try:
+                latest_version = requests.get('https://pypi.org/pypi/delphi-epidata/json').json()['info']['version']
+                if latest_version != Epidata.client_version:
+                    Epidata.logger.warn(
```
dshemetov, line 116:
```suggestion
                    Epidata.log.warn(
```

src/client/delphi_epidata.py
```diff
@@ -107,6 +108,20 @@ def _request(endpoint, params={}):
...
+                        "Client version not up to date",
+                        client_version=Epidata.client_version,
+                        latest_version=latest_version
+                    )
+            except Exception as e:
+                Epidata.logger.warn("Error getting latest client version", exception=str(e))
```
dshemetov, line 122:
```suggestion
                Epidata.log.warn("Error getting latest client version", exception=str(e))
```

dshemetov (changes requested):

melange396 (changes requested):
This functionality has enough purpose (and code!) for it to be in its own method and not taking up >50% of the lines inside the `_request()` method.

The intent is for this to be helpful to the user -- without their needing to ask for it -- thus it would be better for the flag/argument to be opt-out instead of opt-in, if we give the option at all.

I had envisioned that we would do this on module load [only once, on the initial load into the namespace].  That might make it hard to do opt-in/out behavior, but it could make for a more natural interaction with the user, especially if they are using it in an interactive environment.  What do you think?

src/client/delphi_epidata.py
```diff
@@ -54,6 +52,21 @@ def log(evt, **kwargs):
...
+            latest_version = requests.get('https://pypi.org/pypi/delphi-epidata/json').json()['info']['version']
+            if latest_version != __version__:
+                Epidata.log(
+                    "Client version not up to date",
+                    client_version=__version__,
+                    latest_version=latest_version
+                )
+        except Exception as e:
+            Epidata.log("Error getting latest client version", exception=str(e))
+    _version_check.__func__() # run this once on module load
```
rzats, line 68: In Python <= 3.9, static methods cannot be directly called within the class body.
In Python >= 3.10, it works as you'd expect: `_version_check()`

dshemetov:
```suggestion
    # Run this once on module load. Using dunder method for Python <= 3.9 compatibility 
    # https://stackoverflow.com/a/12718272
    _version_check.__func__()
```

integrations/client/test_delphi_epidata.py
```diff
@@ -306,6 +306,24 @@ def test_sandbox(self, get, post):
       Epidata.debug = False
       Epidata.sandbox = False
 
+  @patch('requests.get')
+  def test_version_check(self, get):
+    """Test that in debug + sandbox mode request params are correctly logged, but no queries are sent."""
```
melange396, line 311: this comment doesnt seem applicable here

integrations/client/test_delphi_epidata.py
```diff
@@ -306,6 +306,24 @@ def test_sandbox(self, get, post):
...
+      def raise_for_status(self): pass
+      def json(self): return json.loads(self.content)
+    get.reset_mock()
+    get.return_value = MockJson(b'{"info": {"version": "0.0.1"}}', 200)
+    Epidata._version_check()
+    captured = self.capsys.readouterr()
+    output = captured.err.splitlines()
+    self.assertEqual(len(output), 1)
+    self.assertIn("Client version not up to date", output[0])
+    self.assertIn("\'latest_version\': \'0.0.1\'", output[0])
```
melange396, line 325: would it make sense to add another test that tries to re-import the module to ensure the message is not repeated?

src/client/delphi_epidata.py
```diff
@@ -54,6 +52,21 @@ def log(evt, **kwargs):
         kwargs['timestamp'] = time.strftime("%Y-%m-%d %H:%M:%S %z")
         return sys.stderr.write(str(kwargs) + "\n")
 
+    # One-time check to verify the client version is up to date.
```
melange396, line 55:
```suggestion
    # Check that this client's version matches the most recent available, runs just once per program execution (on initial module load).
```

src/client/delphi_epidata.py
melange396: plz also add newline above function call!

would it be any better/easier to avoid this pattern here and just call `Epidata._version_check()` at the end of the file (after the class definition)?

melange396 (changes requested):
plz also add an entry to CHANGELOG!

melange396 (approved):

melange396 (approved):

melange396 (approved):

Issue #939, opened 2022-06-27
Infra refactor: convert `acquisition` to an installable package


krivard:
The current deployment system fakes a `delphi.epidata.acquisition` package by moving directories around. This results in a system that runs correctly, but which actively resists IDE helper utilities for import tracking and call-reference chasing, making it difficult for new developers to come onboard, and easy to make silly mistakes.

We should find a way to make acquisition a real package. This may also mean replacing `github_deploy_repo` as our primary deployment mechanism.

dshemetov:
See [this comment](https://github.com/cmu-delphi/delphi-epidata/pull/936#discussion_r907605941) specifically.

dshemetov:
Is it true that most of the necessity for directory restructuring comes from this repo having a hyphen, which python doesn't allow in its package/module names? That's the impression I get from [this file](https://github.com/cmu-delphi/operations/blob/main/dev/docker/python/assets/setup.sh) that does the directory setup.

dshemetov:
@ryantibs' [suggestion](https://github.com/cmu-delphi/epidatr/issues/31#issuecomment-1186053373) to rename `delphi-epidata` to `epidata` fits into this work.

melange396:
related: #1037 and potentially #1004

PR #1470, opened 2024-06-05
feat: port pyclient build config to pyproject.toml

dshemetov:
### Summary:

I hoped the more modern config's [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) would work, but it only works if `__version__` is defined explicitly in `__init__.py`, not imported from `delphi_epidata.py` (importing `delphi_epidata.py` errors at build-time because the dependencies in that file aren't installed in the build environment (just setuptools)).

So, this doesn't come with a useful feature, it's just a lateral move to the more modern packaging config. 

On the plus side, I learned a lot by reading parts of these guides:

- [PyPA Sample Project](https://github.com/pypa/sampleproject)
- [setuptools docs](https://setuptools.pypa.io/en/latest/index.html)
- [PyPA Packaging User Guide](https://packaging.python.org/en/latest/)

Might be handy in the future when we do #939 and other such changes.

### Prerequisites:

- [x] Build is successful


src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+
+# Specification (technical, formal):
+# https://packaging.python.org/en/latest/specifications/pyproject-toml/
+
+
+# Choosing a build backend:
+# https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend
+[build-system]
+# A list of packages that are needed to build your package:
+requires = ["setuptools>=61.0.0"] # REQUIRED if [build-system] table is used
```
melange396, line 12: what does that version get us?  the sample/template `pyproject.toml` [doesnt specify it](https://github.com/pypa/sampleproject/blob/a1f00056ebc65d17ae69c5aba324ef4f0382138a/pyproject.toml#L12)

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+# that file directly.
+#
+# This field corresponds to the "Description" metadata field:
+# https://packaging.python.org/specifications/core-metadata/#description-optional
+readme = "README.md"
+
+# Specify which Python versions you support. In contrast to the
+# 'Programming Language' classifiers in this file, 'pip install' will check this
+# and refuse to install the project if the version does not match. See
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires
```
melange396, line 56: these are simple fields, all those comments arent necessary and make it harder to find relevant info
```suggestion
name = "delphi_epidata" # REQUIRED, is the only field that cannot be marked as dynamic.
version = "4.1.23"
description = "A programmatic interface to Delphi's Epidata API."
readme = "README.md"
```

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+requires-python = ">=3.8"
```
melange396, line 57: what is the earliest version we can use here?  im a little worried that this could be a breaking change for a lot of people, especially since we didnt specify a python version constraint before.

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+
+# This is either text indicating the license for the distribution, or a file
+# that contains the license.
+# https://packaging.python.org/en/latest/specifications/core-metadata/#license
+license = { file = "LICENSE.txt" }
```
melange396, line 62:
```suggestion
license = { file = "LICENSE" }
```
our license file doesnt have the txt suffix

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+
+# This should be your name or the name of the organization who originally
+# authored the project, and a valid email address corresponding to the name
+# listed.
```
melange396, line 66: more unnecessary comments
```suggestion
```

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+authors = [{ name = "David Farrow", email = "dfarrow0@gmail.com" }]
+
+# This should be your name or the names of the organization who currently
+# maintains the project, and a valid email address corresponding to the name
+# listed.
+# maintainers = [
+#     { name = "A. Great Maintainer", email = "maintainer@example.com" },
+# ]
```
melange396, line 74:
```suggestion
maintainers = [
    { name = "Delphi Support", email = "delphi-support+pypi@andrew.cmu.edu" },
]
```

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+
+    # Specify the Python versions you support here. In particular, ensure
+    # that you indicate you support Python 3. These classifiers are *not*
+    # checked by "pip install". See instead "requires-python" key in this file.
+    "Programming Language :: Python",
+    "Operating System :: OS Independent",
+    "Intended Audience :: Science/Research",
+    "Natural Language :: English",
+    "Topic :: Scientific/Engineering :: Bio-Informatics",
+]
```
melange396, line 95: get rid of more comments, add python v3 as one comment suggested.  i dont wanna claim this package is "stable" but im not sure if "beta" is appropriate either...  i wouldnt be upset if we just omitted the development status tag ¯\\\_(ツ)_/¯ 
```suggestion
# For a list of valid classifiers, see https://pypi.org/classifiers/
classifiers = [
    "Development Status :: 4 - Beta",
    "License :: OSI Approved :: MIT License",
    "Programming Language :: Python",
    "Programming Language :: Python :: 3",
    "Operating System :: OS Independent",
    "Intended Audience :: Science/Research",
    "Natural Language :: English",
    "Topic :: Scientific/Engineering :: Bio-Informatics",
]
```

src/client/packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+# List URLs that are relevant to your project
+#
+# This field corresponds to the "Project-URL" and "Home-Page" metadata fields:
+# https://packaging.python.org/specifications/core-metadata/#project-url-multiple-use
+# https://packaging.python.org/specifications/core-metadata/#home-page-optional
+#
+# Examples listed include a pattern for specifying where the package tracks
+# issues, where the source is hosted, where to say thanks to the package
+# maintainers, and where to support the project financially. The key is
+# what's used to render the link text on PyPI.
```
melange396, line 115: more superfluous comments
```suggestion
dependencies = ["aiohttp", "delphi-utils", "requests>=2.7.0", "tenacity"]

```

melange396 (changes requested):
this looks great, i really like it!  there are waaaaaay too many comments in here tho....

src/client/packaging/pypi/pyproject.toml
dshemetov: Ah yea, it's not necessary here, I'll remove it. It's needed to support the dynamic version feature (see the callout box in bullet 1 [here](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/) "As of the release of setuptools 61.0.0...").

src/client/packaging/pypi/pyproject.toml
dshemetov: So the client file claims that it works with Python 2 and the old setup.py file made no restrictions on Python version. It's probably best to remove this 👍

melange396 (changes requested):
looks good, except i dont think this line is going to work any more: https://github.com/cmu-delphi/delphi-epidata/blob/69835d1d7795eaf9a710d9f4903fef22a07e8fdf/.github/workflows/release-helper.yml#L67

which probably also means updating this line:
https://github.com/cmu-delphi/delphi-epidata/blob/69835d1d7795eaf9a710d9f4903fef22a07e8fdf/.github/workflows/release-helper.yml#L60

and we should put an entry in https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/client/packaging/pypi/CHANGELOG.md

dshemetov:
Good point, thanks! I swapped the commands over to the new ones ([sdist](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#source-distributions) and [wheel](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#pure-python-wheels)) and added the new `build` dependency to CI. 

Also updated the publish action based on hints from my IDE:

image

And updated changelog too.

melange396:
we can always fix the date on release day

melange396 (approved):

PR #1473, opened 2024-06-07
fix(pyclient ci): validate version arg in create-release

dshemetov:
### Summary:

Fix the release CI to accept any version number and major, minor, patch as its argument. My test script for the versionNumber logic is below.

Notes:

- [set-output was deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), so moved over to the new thing.
- tried [bump-my-version](https://github.com/callowayproject/bump-my-version), but it wasn't a significant improvement and had dependency resolver issues in `pip install -r requirements.api.txt -r requirements.dev.txt` in our Dockerfile.
- [regex tests](https://regexr.com/81dmt). 

```sh
# Try (major|minor|patch|5.3.0|blah|3.2.2.2)
versionName=major
allowed_pattern="^(major|minor|patch|[0-9]+\.[0-9]+\.[0-9]+)$"

if [[ ! $versionName =~ $allowed_pattern ]]; then
  echo "\nInvalid version name: $versionName"
  exit 1
fi

if [[ $versionName =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
  NEXT_TAG="$versionName"
else
  NEXT_TAG=$(bump2version --dry-run --list $versionName | grep ^new_version | sed -r s,"^.*=",,)
fi
echo "next_tag=$NEXT_TAG"
```

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       versionName:
-        description: 'Semantic Version Number (i.e., 5.5.0 or patch, minor, major, prepatch, preminor, premajor, prerelease)'
+        description: 'Semantic Version Number (i.e., 5.5.0 or patch, minor, major)'
         required: true
         default: patch
 
@@ -27,10 +27,23 @@ jobs:
           python-version: 3.8
       - name: Change version number
         id: version
+        # See this issue for explanation and testing:
+        # https://github.com/cmu-delphi/delphi-epidata/pull/1473
         run: |
-          python -m pip install bump2version
-          echo -n "::set-output name=next_tag::"
-          bump2version --list ${{ github.event.inputs.versionName }} | grep new_version | sed -r s,"^.*=",,
+          python -m pip install bump-my-version
+          echo -n "next_tag=" >> $GITHUB_OUTPUT
+          allowed_pattern="^(major|minor|patch|[0-9]+\.[0-9]+\.[0-9]+)$"
+
+          if [[ ! ${{ github.event.inputs.versionName }} =~ $allowed_pattern ]]; then
+            echo "\nInvalid version name: ${{ github.event.inputs.versionName }}"
+            exit 1
+          fi
+
+          if [[ ${{ github.event.inputs.versionName }} =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "${{ github.event.inputs.versionName }}"
+          else
+            echo $(bump-my-version show-bump | grep ${{ github.event.inputs.versionName }} | sed -r s,"^.*─ ",,)
+          fi
       - name: Get main branch SHA
         id: base-sha
         run: echo "sha=$(git rev-parse origin/main)" >> $GITHUB_OUTPUT
@@ -46,7 +59,7 @@ jobs:
         if: steps.changed-py.outputs.any_changed == 'true'
         run: |
           cd src/client/packaging/pypi
-          bump2version --allow-dirty --new-version ${{ steps.version.outputs.next_tag }} ignore_part
+          bump-my-version --allow-dirty --new-version ${{ steps.version.outputs.next_tag }}
       - name: Create pull request into prod
         uses: peter-evans/create-pull-request@v3
         with:
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -30,8 +30,7 @@ jobs:
         # See this issue for explanation and testing:
         # https://github.com/cmu-delphi/delphi-epidata/pull/1473
         run: |
-          python -m pip install bump-my-version
-          echo -n "next_tag=" >> $GITHUB_OUTPUT
+          python -m pip install bump2version
           allowed_pattern="^(major|minor|patch|[0-9]+\.[0-9]+\.[0-9]+)$"
 
           if [[ ! ${{ github.event.inputs.versionName }} =~ $allowed_pattern ]]; then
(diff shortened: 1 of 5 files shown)


.github/workflows/create-release.yml
```diff
@@ -27,10 +27,23 @@ jobs:
...
+          if [[ ! ${{ github.event.inputs.versionName }} =~ $allowed_pattern ]]; then
+            echo "\nInvalid version name: ${{ github.event.inputs.versionName }}"
+            exit 1
+          fi
+
+          if [[ ${{ github.event.inputs.versionName }} =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "${{ github.event.inputs.versionName }}"
+          else
+            echo $(bump-my-version show-bump | grep ${{ github.event.inputs.versionName }} | sed -r s,"^.*─ ",,)
+          fi
```
melange396, line 46: `bump-my-version` doesnt look a whole lot better than `bump2version`...  what if we just keep using bump2 but with the logic to either extract a numeric version from the input w/ your regex, or otherwise try to generate a new numeric version?

```suggestion
          python -m pip install bump2version
          if [[ ${{ github.event.inputs.versionName }} =~ "^[0-9]+\.[0-9]+\.[0-9]+$" ]]; then
            NEXT_TAG="${{ github.event.inputs.versionName }}"
          else
            NEXT_TAG=$(bump2version --dry-run --list ${{ github.event.inputs.versionName }} | grep ^new_version | sed -r s,"^.*=",,)
          fi
          bump2version --new-version $NEXT_TAG _ignored_arg_
          echo "next_tag=$NEXT_TAG" >> $GITHUB_OUTPUT
```

and then below:
`bump2version --allow-dirty --new-version ${{ steps.version.outputs.next_tag }} _ignored_arg_`

dshemetov: Sure, I pushed a fix.

rzats (approved):
lgtm!

src/client/packaging/pypi/pyproject.toml
```diff
@@ -42,3 +42,27 @@ dependencies = ["aiohttp", "delphi-utils", "requests>=2.7.0", "tenacity"]
 [project.urls]
 "Homepage" = "https://github.com/cmu-delphi/delphi-epidata"
 "Changelog" = "https://github.com/cmu-delphi/delphi-epidata/blob/main/src/client/packaging/pypi/CHANGELOG.md"
+
```
melange396, line 45: you should revert these changes

src/client/packaging/pypi/.bumpversion.cfg
```diff
@@ -1,8 +0,0 @@
-[bumpversion]
```
melange396, line 1: you should revert these changes (un-delete the file)

.github/workflows/create-release.yml
```diff
@@ -27,10 +27,23 @@ jobs:
...
+            echo "\nInvalid version name: ${{ github.event.inputs.versionName }}"
+            exit 1
+          fi
+
+          if [[ ${{ github.event.inputs.versionName }} =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            NEXT_TAG="${{ github.event.inputs.versionName }}"
+          else
+            NEXT_TAG=$(bump2version --dry-run --list ${{ github.event.inputs.versionName }} | grep ^new_version | sed -r s,"^.*=",,)
+          fi
+          echo "next_tag=$NEXT_TAG" >> $GITHUB_OUTPUT
```
melange396, line 46: we need to make sure the new version gets applied:
```suggestion
          if [[ ${{ github.event.inputs.versionName }} =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            # use given version number
            NEXT_TAG="${{ github.event.inputs.versionName }}"
          else
            # calculate new version number based on given tag
            NEXT_TAG=$(bump2version --dry-run --list ${{ github.event.inputs.versionName }} | grep ^new_version | sed -r s,"^.*=",,)
          fi
          # apply given or calculated version number
          bump2version --new-version $NEXT_TAG _ignored_arg_
          # save version number for later
          echo "next_tag=$NEXT_TAG" >> $GITHUB_OUTPUT
```

melange396 (changes requested):

.github/workflows/create-release.yml
dshemetov: We never actually applied the version in this step before, so I don't think is right. The actual version bump happens in the  "Bump Python versions if client files changed" step below.

melange396: we did, it was in old line 33: `bump2version --list ${{ github.event.inputs.versionName }} | grep new_version | sed -r s,"^.*=",,`

the step below is just for the client version, this is for the server and acquisition code.

dshemetov: That line doesn't actually bump version, it just extracts the version number out and I'm already doing it in new line 44. The 
```
          # apply given or calculated version number
          bump2version --new-version $NEXT_TAG _ignored_arg_
```
in your suggestion was never done in the original action in this step and I don't think we need it. Are you intending to change the behavior of the original CI here? This PR is just about fixing the CI argument validation.

dshemetov: Ah I misunderstood, thought that `bump2version --list ${{ github.event.inputs.versionName }} | grep new_version | sed -r s,"^.*=",,` didn't actually operate on the files and make changes.

requirements.dev.txt
```diff
@@ -1,5 +1,6 @@
 aiohttp==3.9.4
 black>=20.8b1
+bump2version==1.0.1
```
melange396, line 3: you can take this back out; it doesnt get used anywhere in this repo except for those two gh actions files, where it gets installed w/o versioning by calls to `python -m pip install bump2version`

dshemetov: cool, thanks for checking

dshemetov:
Cleaned up the commit log by squashing the edits.

PR #1482, opened 2024-06-20
Add mailing list info to API key registration email

melange396:
This just updates our repo's copy of the js file, the running Google Apps Script has already been modified.


nmdefries (approved):
Looks good!

Issue #1467, opened 2024-06-03
Tons of new dependencies introduced in requirements for delphi-epidata


boxydog:
Thanks for your work on epidata.

Between `delphi-epidata` 4.1.20 and 4.1.23, you seem to have introduced at least 66 new dependencies, see below. (I have a few in our project already, so you may have introduced more than 66.)

I am concerned that introducing all these new dependencies into my production application will make it less likely to be reliable.

Is there any way to get a more slimmed down version of `delphi-epidata`?

```
$ diff tmp/[ab]
5a6
> astroid
6a8,9
> boto3
> botocore
7a11
> cffi
8a13,22
> clarabel
> click
> click-plugins
> cligj
> contourpy
> coverage[toml]
> covidcast
> cryptography
> cvxpy
> cycler
9a24,25
> delphi-utils
> descartes
16a33
> ecos
17a35,37
> fiona
> fonttools
> freezegun
18a39,41
> geopandas
> gitdb
> gitpython
23a47,52
> imageio
> imageio-ffmpeg
> iniconfig
> isort
> jinja2
> jmespath
25a55,61
> kiwisolver
> lazy-object-proxy
> markupsafe
> matplotlib
> mccabe
> mock
> moto
28a65
> osqp
29a67,69
> pandas
> pillow
> pluggy
30a71
> pycparser
32a74,83
> pydocstyle
> pylint
> pyparsing
> pyproj
> pytest
> pytest-cov
> python-dateutil
> pytz
> pyyaml
> qdldl
34a86,88
> requests-mock
> responses
> s3transfer
36a91,95
> scs
> shapely
> six
> slackclient
> smmap
37a97
> snowballstemmer
38a99
> structlog
41a103
> toml
44a107
> werkzeug
45a109,111
> wrapt
> xlrd
> xmltodict
```

boxydog:
I suspect this was from more closely tying `delphi-utils` this repo. It might be #1460?

dshemetov:
Thank you for the report. This will take us a bit of time to address. A few possible workarounds for now:

* pin the `delphi-epidata` client version to 4.1.20; the main update in 4.1.23 is some logging needed for our internal use; that client version should stay compatible for the foreseeable future (we can let you know in this thread if that changes)
[...]

PR #1486, opened 2024-06-25
py client: drop delphi_utils requirement, log to stderr instead

melange396:
This drops the requirement for the bloated `delphi_utils` package in the python client, and replaces the `delphi_utils.logger` with printing to STDERR instead.

Addresses #1467 

I havent tested this yet, and i imagine it is going to fail CI.

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -8,6 +8,8 @@ Notes:
  - Compatible with Python 2 and 3.
 """
 
+import sys
+
 # External modules
 import requests
 import time
@@ -15,8 +17,6 @@ import asyncio
 from tenacity import retry, stop_after_attempt
 
 from aiohttp import ClientSession, TCPConnector, BasicAuth
-
-from delphi_utils.logger import get_structured_logger
 
 __version__ = "4.1.23"
 
@@ -45,9 +45,14 @@ class Epidata:
 
     client_version = __version__
 
-    logger = get_structured_logger('delphi_epidata_client')
     debug = False # if True, prints extra logging statements
     sandbox = False # if True, will not execute any queries
+
+    @staticmethod
+    def log(evt, **kwargs):
+        kwargs['event'] = evt
+        kwargs['timestamp'] = time.strftime("%Y-%m-%d %H:%M:%S %z")
+        return sys.stderr.write(str(kwargs) + "\n")
 
     # Helper function to cast values and/or ranges to strings
     @staticmethod
@@ -72,7 +77,7 @@ class Epidata:
         """Make request with a retry if an exception is thrown."""
         request_url = f"{Epidata.BASE_URL}/{endpoint}/"
         if Epidata.debug:
-            Epidata.logger.info("Sending GET request", url=request_url, params=params, headers=_HEADERS, auth=Epidata.auth)
+            Epidata.log("Sending GET request", url=request_url, params=params, headers=_HEADERS, auth=Epidata.auth)
         if Epidata.sandbox:
             resp = requests.Response()
             resp._content = b'true'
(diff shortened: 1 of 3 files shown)


src/client/delphi_epidata.py
```diff
@@ -45,10 +45,15 @@ class Epidata:
...
 
-    logger = get_structured_logger('delphi_epidata_client')
     debug = False # if True, prints extra logging statements
     sandbox = False # if True, will not execute any queries
 
+    @staticmethod
+    def log(evt, **kwargs):
+        kwargs['event'] = evt
+        kwargs['timestamp'] = time.strftime("%Y-%m-%d %H:%M:%S %z")
+        return sys.stderr.write(str(kwargs) + "\n")
```
melange396, line 55: this might need a `flush()` call to keep it synced
```suggestion
        sys.stderr.write(str(kwargs) + "\n")
        sys.stderr.flush()
```

dshemetov:
Seems like a simple enough workaround 👍 .

---

I was curious to learn a little about stderr and structlog. A few basic notes I found helpful, for future reference:
- `get_structured_logger()` is set to always use [`logger.StreamHandler()`](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/_delphi_utils_python/delphi_utils/logger.py#L77) (I'm actually not sure if [the branch with `logging.FileHandler()`](https://github.com/cmu-delphi/covidcast-indicators/blob/84d059751b646c0075f1a384741f2c1d80981269/_delphi_utils_python/delphi_utils/logger.py#L121-L122) is ever hit, since I think `system_logger.handlers` always has the `StreamHandler()` as per the config above)
- [`logger.StreamHandler()`](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler) defaults to `stderr` as its output
- so I think the end result is that structlog does our formatting and string rendering and the standard logging library then forwards that string to stderr (structlog has some suggested configurations and ours seems closest to [this one](https://www.structlog.org/en/stable/standard-library.html#rendering-within-structlog))

melange396:
@dshemetov youre right, the `FileHandler` stuff is basically unreachable.  One of the googlebros [borked it](https://github.com/cmu-delphi/covidcast-indicators/pull/980/files), but it wasnt much better before that anyway...  I guess we are lucky that we never needed to directly write to a log file with this!

dshemetov:
Ah well this actually answers a question that came up in a conversation with @aysim319: I was wondering whether any of [these `get_structured_logger` calls](https://github.com/search?q=repo:cmu-delphi/covidcast-indicators+get_structured_logger&type=code) actually depended on that filename argument in any important way and looks like the answer is no. Since it's been like this for 3 years, we must not rely on those file logs. On the plus side, that suggests that we can remove that argument, so the logger doesn't depend on the params file, so it can be pulled out of the `run` function, set as a global in each module, and then we can stop passing the logger around from function to function.

src/client/delphi_epidata.py
dshemetov: TL;DR: we don't need the `flush()`.

```py
>>> sys.stderr
<_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>
>>> sys.stderr.line_buffering
True
```

[`io.TextIOWrapper`](https://docs.python.org/3/library/io.html#io.TextIOWrapper) says this about `line_buffering=True`

> If line_buffering is True, [flush()](https://docs.python.org/3/library/io.html#io.IOBase.flush) is implied when a call to write contains a newline character or a carriage return.

dshemetov (approved):
Tests look good, new `log()` function with `str({...})` looks good. Low risk, non-user facing change, removes lots of dependencies.

PR #1497, opened 2024-07-19
py client version check fixes and cleanup

melange396:
fixes on top of #1456, to address CI failures seen in #1496

### DIFF
--- a/src/client/delphi_epidata.py
+++ b/src/client/delphi_epidata.py
@@ -43,8 +43,6 @@ class Epidata:
     BASE_URL = "https://api.delphi.cmu.edu/epidata"
     auth = None
 
-    client_version = __version__
-
     debug = False # if True, prints extra logging statements
     sandbox = False # if True, will not execute any queries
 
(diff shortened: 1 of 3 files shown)


melange396:
i believe this test is now broken (or is at least inadequate) :
https://github.com/cmu-delphi/delphi-epidata/blob/21c7df6dd46cd14240fa4fd566fc0fc2629131de/integrations/client/test_delphi_epidata.py#L336-L340

src/client/delphi_epidata.py
```diff
@@ -46,29 +46,38 @@ class Epidata:
...
+
     @staticmethod
     def log(evt, **kwargs):
         kwargs['event'] = evt
         kwargs['timestamp'] = time.strftime("%Y-%m-%d %H:%M:%S %z")
         return sys.stderr.write(str(kwargs) + "\n")
 
-    # Check that this client's version matches the most recent available, runs just once per program execution (on initial module load).
+    # Check that this client's version matches the most recent available.
+    # This is indended to run just once per program execution, on initial module load.
```
dshemetov, line 58:
```suggestion
    # This is intended to run just once per program execution, on initial module load.
```

dshemetov (changes requested):
I was surprised that this test didn't error:
```py
  @patch('delphi.epidata.client.delphi_epidata.Epidata._version_check')
  def test_version_check_once(self, version_check):
    """Test that the _version_check() function is only called once on initial module import."""
    from delphi.epidata.client.delphi_epidata import Epidata
    version_check.assert_not_called()
```
Since we moved the flag check inside the function, `_version_check()` should be called and then take the quick exit branch. But this doesn't error, so `_version_check()` was never called, which lead me to think that Python doesn't rerun a module once it's imported and this turns out to be true.

> The first place checked during import search is [sys.modules](https://docs.python.org/3/library/sys.html#sys.modules). This mapping serves as a cache of all modules that have been previously imported, including the intermediate paths. So if foo.bar.baz was previously imported, [sys.modules](https://docs.python.org/3/library/sys.html#sys.modules) will contain entries for foo, foo.bar, and foo.bar.baz. Each key will have as its value the corresponding module object.

From [The module cache](https://docs.python.org/3/reference/import.html#the-module-cache). 

You can test it like this:

```py
# a.py
class Test:
    @staticmethod
    def hello():
        print("Hello")

class Test2:
    pass

Test.hello()

# b.py
from a import Test
from a import Test2

➜  epidatpy git:(dev) ✗ python b.py
Hello
```

This holds even a is imported in different files. 

So it seems like we don't actually need this flag and my bad for suggesting all this unneeded stuff. We probably don't need the unit tests associated with the check running only once either, since we don't need to test Python's import system. Would be good to write a comment explaining that `_version_check` will automatically only run on the first import.

minhkhul (approved):
~Agree w Rostyslav on unit test. Also, just curious, what does `e___vdc` stands for?~ 
Edit: just read through Dmitry's suggestions.

dshemetov (approved):

minhkhul (approved):

PR #1542, opened 2024-09-13
Add rvdss indicator/data source 

cchuong:
### Summary:

Code to create a new datasource for the epidata AP. The data is respiratory virus detections in Canada, reported by the Public Health Agency of Canada (PHAC). 

### Prerequisites:

- [ ] Build is successful


src/acquisition/rvdss/rvdss_historic.py
```diff
@@ -0,0 +1,652 @@
...
+
+def abbreviate_geo(full_name):
+    lowercase=full_name.lower()
+    
+    if "newfoundland" in lowercase:
+        abbrev =  "nl"
+    elif "prince edward island" in lowercase:
+        abbrev =  "pe"
+    elif "nova scotia" in lowercase:
+        abbrev =  "ns"
```
nmdefries, line 100: suggestion: We'll want this to be in a dictionary in `constants.py`, like [this](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/acquisition/fluview/fluview_locations.py#L17).

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,199 @@
...
+    elif "pr" in lowercase:
+        abbrev = "prairies" 
+    elif "terr" in lowercase:
+        abbrev = "territories" 
+    else:
+        abbrev=lowercase
+    return(abbrev)
+
+def create_geo_types(geo,default_geo):
+    regions = ['atlantic','atl','province of québec','québec','qc','province of ontario','ontario','on',
```
nmdefries, line 79: question: I know we [previously decided](https://delphi-org.slack.com/archives/C078XQGDHJR/p1724969838674169) that we'd mix 2-letter province abbreviations and full region names when returning regional data (since some provinces are their own region), but I'm questioning that now. Should we just mix 2-letter province abbrevs with longer regional abbreviations (like `atl`, `terr`, `pr` listed here)?

nmdefries: Decided in discussions with Dan that, since regions don't have standard abbreviations, we should just list out their full names.

nmdefries:
We also want to have some unit tests [here](https://github.com/cmu-delphi/delphi-epidata/tree/dev/tests/acquisition). This can be complicated for functions that normally interact with external components (websites, databases, etc), but we want to at least test functions that don't require that. (External components can be faked with `mock`ing packages, but it's not essential.)

nmdefries (commented):
I renamed some things, moved some variables to `constants.py`, and added some comments. I will be looking at this more, but wanted to get you started with some feedback.

Overall, I'm having trouble understanding the flow of the code. I think this could be improved by adding more [docstrings](https://peps.python.org/pep-0257/#what-is-a-docstring) and comments in general, mid-level functions, and clear function naming.

Docstrings: the shortest form of this is one line describing what the function does, the core goal. We don't necessarily need to describe all the args, like

```
    Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)
```

although it can be helpful for more complex and user-facing fns.

Comments: Comments give context to the code. They can say _why_ we're doing this. They can describe broadly what we're doing in a chunk of code (not details of implementation). They can discuss cons or gotchas of the current approach. They can present alternatives to the current approach (generally when the alternatives were considered but were too complex for the current need).

Take [this fn](https://github.com/cmu-delphi/delphi-epidata/blob/cd8308781effbf5fda3b60365b8a71f4f9e3823c/src/acquisition/rvdss/rvdss_historic.py#L86-L105), for example. Why are we processing table captions? What is a table caption? What do the "table_identifiers" represent? What is the `sum([all...` chunk doing (broadly)?

You don't need to answer all these questions in the code, but it's good to insert comments where you think future readers will be confused.

Mid-level functions (optional but something to think about): The helper functions you've defined so far are mostly low level (`get_report_date`), and the higher level functions in this package are pretty long. It makes me think that that they can/should be broken down into more functions that call several of the low-level fns at once.

Function naming (recommended but optional): Try to make fn names very clearly say what they are doing. For example, [`get_report_date` sounds](https://github.com/cmu-delphi/delphi-epidata/blob/cd8308781effbf5fda3b60365b8a71f4f9e3823c/src/acquisition/rvdss/rvdss_historic.py#L70-L83) to me like it is fetching the report date from the website ("get") . However, it's actually interpreting a given start_year-week combo. A better name might be "parse_report_date". Actually, I'm confused by the term "report date", because that makes me think of the date the entire report was posted, but I think this is handling time values (weeks within a given report). So maybe even "parse_time_value"? Or (this is too long, but brainstorming here...) "wrap_early_weeknums_to_next_year".

nmdefries:
Plan to do code walkthrough next week!

src/ddl/rvdss.sql
```diff
@@ -0,0 +1,169 @@
...
+  `geo_type` char(6) NOT NULL,
+  `geo_value` char(35) NOT NULL,
+  `sarscov2_tests` int(10) NOT NULL,
+  `sarscov2_positive_tests` int(10) NOT NULL,
+  `sarscov2_pct_positive` int(10) NOT NULL,
+  `flu_tests` int(10) NOT NULL,
+  `flua_tests` int(10) NOT NULL,
+  `flub_tests` int(10) NOT NULL,
+  `flu_positive_tests` int(10) NOT NULL,
+  `flu_pct_positive` int(10) NOT NULL,
```
nmdefries: issue: All value columns are being stored as `int`, but some of them should be `float`s (`double`), like percent positive. [Example](https://github.com/cmu-delphi/delphi-epidata/blob/dev/src/ddl/fluview.sql#L265-L270)

nmdefries: I think for other datasets, we store all value fields as `float`s.

testdata/acquisition/rvdss/RVD_WeeklyData.csv
```diff
@@ -0,0 +1,2001 @@
+region,province,week,weekorder,year,date,virus,tests,detections,percentpositive
```
nmdefries, line 1: issue: this file (as well as `testdata/acquisition/rvdss/RVD_WeeklyData_Formatted.csv` to a lesser extent) is pretty long for a test file. Issues are: because the test data is complex, it can be hard to debug if there's an issue. And the file is large which will make the repo larger (eventually slows down cloning, pushing, etc).

So if we can, it would be better to make this file shorter/smaller. If we need all lines, then we could try compressing (e.g. `csv.gz`)

testdata/acquisition/rvdss/RVD_CurrentWeekTable_Formatted.csv
```diff
@@ -0,0 +1,1026 @@
+epiweek,time_value,issue,geo_type,geo_value,week,weekorder,sarscov2_tests,sarscov2_positive_tests,flu_tests,flu_positive_tests,fluah1n1pdm09_positive_tests,fluah3_positive_tests,fluauns_positive_tests,flua_positive_tests,flub_positive_tests,rsv_tests,rsv_positive_tests,hpiv_tests,hpiv1_positive_tests,hpiv2_positive_tests,hpiv3_positive_tests,hpiv4_positive_tests,hpivother_positive_tests,adv_tests,adv_positive_tests,hmpv_tests,hmpv_positive_tests,evrv_tests,evrv_positive_tests,hcov_tests,hcov_positive_tests,year
```
nmdefries, line 1: issue: same comment about large size

testdata/acquisition/rvdss/RVD_CurrentWeekTable.csv
```diff
@@ -0,0 +1,1026 @@
+week,weekorder,date,ReportingLaboratory,test_sarscov2,pos_sarscov2,test_flu,pos_flu,pos_flua_h1,pos_flua_h3,pos_flua_uns,pos_flua,pos_flub,test_rsv,pos_rsv,test_hpiv,pos_hpiv_1,pos_hpiv_2,pos_hpiv_3,pos_hpiv_4,pos_hpiv_other,test_adv,pos_adv,test_hmpv,pos_hmpv,test_evrv,pos_evrv,test_hcov,pos_hcov
```
nmdefries, line 1: issue: same comment about large size

src/acquisition/rvdss/constants.py
```diff
@@ -0,0 +1,135 @@
...
+    "phol timmins":"timmins phl",
+    "phol toronto": "cphl toronto"
+ }
+
+# Regions are groups of provinces that are geographically close together. Some single provinces are reported as their own region (e.g. Québec, Ontario).
+REGIONS = ['atlantic','atl','at','province of québec','québec','qc','province of ontario','ontario','on',
+            'prairies', 'pr', "british columbia",'bc',"territories",'terr']
+NATION = ["canada","can",'ca']
+
+PROVINCES = ['nl','pe','ns','nb','qc','on','mb','sk','ab','bc','yt','nt','nu']
```
nmdefries, line 137: question: If we have the map above that standardizes geo names/spellings, do we need to list all possible options again under `REGIONS` and `NATION`? If yes, do we need to do the same for `PROVINCES`? If we need to do this, I can suggest a way to derive all options from the `GEOS` map above.

src/acquisition/rvdss/constants.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Construct dashboard and data report URLS.
+DASHBOARD_BASE_URL = "https://health-infobase.canada.ca/src/data/respiratory-virus-detections/"
+DASHBOARD_W_DATE_URL = DASHBOARD_BASE_URL + "archive/{date}/"
+
+# May not need this since we write a function for this in pull_historic
```
nmdefries: todo: looks like we're still using this in `pull_historic.py`. Do we want to keep this var or switch to using the function?

src/acquisition/rvdss/constants.py
```diff
@@ -0,0 +1,135 @@
...
+ALTERNATIVE_SEASON_BASE_URL = "www.phac-aspc.gc.ca/bid-bmi/dsd-dsm/rvdi-divr/"
+HISTORIC_SEASON_REPORTS_URL = SEASON_BASE_URL+"/en/public-health/services/surveillance/respiratory-virus-detections-canada/{year_range}.html"
+DASHBOARD_ARCHIVED_DATES_URL= "https://health-infobase.canada.ca/src/js/respiratory-virus-detections/ArchiveData.json"
+
+# Each URL created here points to a list of all data reports made during that
+# season, e.g.
+# https://www.canada.ca/en/public-health/services/surveillance/respiratory-virus-detections-canada/2014-2015.html.
+# The Public Health Agency of Canada site switched in 2024 to reporting
+# disease data in a dashboard with a static URL. Therefore, this collection
+# of URLs does _NOT_ need to be updated. It is used for fetching historical
```
nmdefries: praise: appreciate the clear comment!

src/acquisition/rvdss/constants.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Regions are groups of provinces that are geographically close together. Some single provinces are reported as their own region (e.g. Québec, Ontario).
+REGIONS = ['atlantic','atl','at','province of québec','québec','qc','province of ontario','ontario','on',
+            'prairies', 'pr', "british columbia",'bc',"territories",'terr']
+NATION = ["canada","can",'ca']
+
+PROVINCES = ['nl','pe','ns','nb','qc','on','mb','sk','ab','bc','yt','nt','nu']
+
+# Construct dashboard and data report URLS.
+DASHBOARD_BASE_URL = "https://health-infobase.canada.ca/src/data/respiratory-virus-detections/"
```
nmdefries: note: I moved all the URLs, shorter lists of things, etc up to the top of the file for easier reading

nmdefries (commented):
Formatting comments. Looking at logic separately.

src/acquisition/rvdss/run.py
```diff
@@ -0,0 +1,174 @@
...
+
+def update_current_data(logger):
+    logger.info("Updating current data")
+    ## Check if data for current update date has already been fetched
+    headers = {
+        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36'
+    }
+
+    update_date = get_dashboard_update_date(DASHBOARD_BASE_URL, headers)
+    already_updated = check_most_recent_update_date(update_date,UPDATE_DATES_FILE)
```
nmdefries, line 25: issue: need to create this file before running in production otherwise this will fail. `open(<file>, "a")`, as below, creates the file if it doesn't exist, but the `open()` call in `check_most_recent_update_date` doesn't set the mode. We probably want to set the mode to "a+" ([options](https://stackoverflow.com/a/1466036)). That also creates the file if it doesn't exist.

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,416 @@
...
+    
+    if pattern1.match(update_date_url_response.text):
+        update_date = datetime.strptime(update_date_url_response.text,"%Y-%m-%d %H:%M:%S").strftime("%Y-%m-%d")
+    elif pattern2.match(update_date_url_response.text):
+        update_date = datetime.strptime(update_date_url_response.text,"%m/%d/%Y %H:%M:%S").strftime("%Y-%m-%d")
+    
+    return(update_date)
+
+def check_most_recent_update_date(date,date_file):
+    with open(date_file) as file:
```
nmdefries, line 99: issue: need to create this file before running in production otherwise this will fail. We probably want to set the mode to "a+" (options). That creates the file if it doesn't exist.

src/acquisition/rvdss/run.py
```diff
@@ -0,0 +1,174 @@
...
+        
+            data_dict={"positive":pos_data,"respiratory_detection":resp_data} 
+            
+            # Combine all rables into a single table
+            data = combine_tables(data_dict)
+            data['time_type'] = "week"
+            
+            #update database
+            update(data,logger)
+        logger.info("Finished patching in {season[0]}-{season[1]} season")
```
nmdefries, line 116: issue:
```suggestion
        logger.info(f"Finished patching in {start}-{end} season")
```

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    if geo in NATION:
+        geo_type="nation"
+    elif geo in REGIONS:
+        geo_type="region"
+    else:
+        geo_type = default_geo
+    return(geo_type)
+
+def check_date_format(date_string):
+    if not re.search("[0-9]{4}-[0-9]{2}-[0-9]{2}",date_string):
```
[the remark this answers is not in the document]
nmdefries: bumping this. I think a try-catch will be better for maintainability/readability.

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,416 @@
...
+def convert_date_to_int(date_string):
+    if re.search("[0-9]{4}-[0-9]{2}-[0-9]{2}",str(date_string)):
+        new_date = int(re.sub("-","",str(date_string)))
+    else:
+        new_date=date_string
+    return(new_date)
+
+def get_dashboard_update_date(base_url,headers):
+    # Get update date
+    update_date_url =  base_url + "RVD_UpdateDate.csv"
```
nmdefries, line 85: suggestion: filename better as a constant.

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,416 @@
...
+    positive_data = get_positive_data(url,headers,update_date)
+
+    return {
+        "respiratory_detection": detections_data,
+        "positive": positive_data,
+        # "count": None, # Dashboards don't contain this data.
+    }
+
+def fetch_current_dashboard_data(url):
+    headers = {
```
nmdefries, line 410: suggestion: since we already defined headers in run.py, can pass those to this function instead of redefining headers.

nmdefries: same with update_date below

src/acquisition/rvdss/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        new_data["rsv_pct_positive"] =  np.divide(new_data["rsv_positive_tests"], new_data["rsv_tests"], 
+                      out=np.zeros_like(new_data["rsv_positive_tests"],dtype=float), 
+                      where=new_data["rsv_tests"]!=0)*100
+    
+    if "hpiv1_positive_tests" in new_data.columns and "hpiv_tests" in new_data.columns:
+        new_data["hpiv_positive_tests"] =  new_data["hpiv1_positive_tests"] + new_data["hpiv2_positive_tests"]+ new_data["hpiv3_positive_tests"]+new_data["hpiv4_positive_tests"]+new_data["hpivother_positive_tests"]
+    
+        new_data["hpiv_pct_positive"] =  np.divide(new_data["hpiv_positive_tests"], new_data["hpiv_tests"], 
+                      out=np.zeros_like(new_data["hpiv_positive_tests"],dtype=float), 
+                      where=new_data["hpiv_tests"]!=0)*100
```
nmdefries, line 328: question/issue: it looks to me like when denominator (total tests) is 0, we set the output to 0. However we should set it to missing (NULL, NaN, etc).

Same for all the np.divid/zeros_like calculations.

nmdefries (dismissed):
There's a few issues with behavior (file creation, etc, and one data handling step). Other comments are about style and are optional.

I'm going to go ahead and approve this, but please fix the content issues before merging. Tagging @aysim319 to keep an eye on this while I'm out.

Note: SonarCloud is failing due to some test files that we need. Have already talked to Brian about this and we can ignore (and maybe just remove SonarClout)

src/acquisition/rvdss/run.py
cchuong: The mode has been set to "a+" for any calls to `open()`.

src/acquisition/rvdss/utils.py
cchuong: Resolved as above

src/acquisition/rvdss/run.py
cchuong: Nice catch! Suggested change made

src/acquisition/rvdss/utils.py
cchuong: Headers are now passed directly to `fetch_current_dashboard_data`

src/acquisition/rvdss/utils.py
cchuong: Changed to a constant

src/acquisition/rvdss/constants.py
cchuong: Let's keep the variable since the dates are static.

src/acquisition/rvdss/utils.py
cchuong: Output is now set to nan when denominator is 0.

nmdefries (approved):

PR #1570, opened 2025-01-06
adding docs

aysim319:
add docs for NHSN data

### DIFF
--- a/docs/api/covidcast-signals/nhsn.md
+++ b/docs/api/covidcast-signals/nhsn.md
@@ -1,0 +1,76 @@
+---
+title: NHSN ED Visits
+parent: Data Sources and Signals
+grand_parent: Main Endpoint (COVIDcast)
+nav_order: 1
+---
+# National Syndromic Surveillance Program Emergency Department Visits
+{: .no_toc}
+
+* **Source name:** `nhsn`
+* **Earliest issue available:** August 08, 2020
+* **Number of data revisions since 18 Nov 2024:** 0
+* **Date of last change:** Never
+* **Available for:** state, hhs, nation (see [geography coding docs](../covidcast_geography.md))
+* **Time type:** week (see [date format docs](../covidcast_times.md))
+* **License:** [Public Domain US Government](https://www.usa.gov/government-works)
+
+## Overview
+
+[The National Healthcare Safety Network (NHSN)](https://www.cdc.gov/nhsn/index.html) is an effort to track epidemiologically relevant conditions.
+This dataset represents preliminary weekly hospital respiratory data and metrics aggregated to national and state/territory levels reported to CDC’s National Health Safety Network (NHSN) beginning August 2020.
+Each signal below is derived from one of two following datasets:
+- Main: [Weekly Hospital Respiratory Data (HRD) Metrics by Jurisdiction, National Healthcare Safety Network (NHSN)](https://data.cdc.gov/Public-Health-Surveillance/Weekly-Hospital-Respiratory-Data-HRD-Metrics-by-Ju/ua7e-t2fy/about_data)
+- Preliminary: [Weekly Hospital Respiratory Data (HRD) Metrics by Jurisdiction, National Healthcare Safety Network (NHSN) (Preliminary)](https://data.cdc.gov/Public-Health-Surveillance/Weekly-Hospital-Respiratory-Data-HRD-Metrics-by-Ju/mpgq-jmmr/about_data). Signals derived from the preliminary dataset have suffix `_prelim` in their signal names.
+
+Both datasets started reporting data in late 2022. As of May 2024, NHSN received data from 78% of US EDs.
+
+| Signal                          | Description                                                                                                                                                                         |
+|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `confirmed_admissions_covid_ew`              | Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week **Earliest date available:** 2020-08-08                                                                                                     |
+| `confirmed_admissions_covid_ew_prelim`          | Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday). <br/> **Earliest date available:** 2020-08-08  |
+| `confirmed_admissions_flu_ew`           | Total number of patients hospitalized with confirmed influenza captured for the Wednesday of the reporting week  <br/> **Earliest date available:** 2020-08-08                                                                                       |
+| `confirmed_admissions_flu_ew_prelim`     | Total number of patients hospitalized with confirmed influenza captured for the Wednesday of the reporting week  preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday).<br/> **Earliest date available:** 2020-08-08 |
+
+## Table of Contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+## Estimation
+
+The confirmed signal `confirmed_admissions_covid_ew` is the value of the `totalconfc19newadm`
+The confirmed signal `confirmed_admissions_flu_ew` is the value of the `totalconfflunewadm`
+
+
+### Geographic weighting
+As the original data is a percentage and raw case counts are not available, `hrr`,`msa`, and `hhs` values are computed from county-level data using a weighted mean. Each county is assigned a weight equal to its population in the last census (2020). Unreported counties are implicitly treated as having a weight of 0 or a value equal to the group mean.
+
+This weighting approach assumes that the number of ED visits is proportional to the overall population of a county, i.e. the per-capita ED visit rate is the same for all counties, which may not be the case (for example, denser counties may have easier access to EDs and thus higher rates of ED visits per capita).
+
+state-level data is **not** simply an average of the county-level data, but may contain facilities omitted at the regional level. For example, state-level values are available for California, even though no California county data is reported; see the [missingness section below](#missingness) for a list of such states.
+
+
+## Missingness
+
+Data for reporting dates through April 30, 2024 represent data reported during a previous mandated reporting period as specified by the HHS Secretary.
+Data for reporting dates May 1, 2024 – October 31, 2024 represent voluntarily reported data in the absence of a mandate. 
+Data for reporting dates beginning November 1, 2024 represent data reported during a current mandated reporting period.
+All data and metrics capturing information on respiratory syncytial virus (RSV) were voluntarily reported until November 1, 2024.
+
+
+## Lag and Backfill
+
+The weekly signal is primarily reported on Friday, adding data from the prior week.
+For example, on Friday, 2024-04-19, the source added new data from the week ending 2024-04-13.
+
+## Limitations
+
+
+## Source and Licensing
+
+This source is derived from the CDC's [Weekly Hospital Respiratory Data (HRD) Metrics by Jurisdiction, National Healthcare Safety Network (NHSN)](https://data.cdc.gov/Public-Health-Surveillance/Weekly-Hospital-Respiratory-Data-HRD-Metrics-by-Ju/ua7e-t2fy/about_data) and
+[Weekly Hospital Respiratory Data (HRD) Metrics by Jurisdiction, National Healthcare Safety Network (NHSN) (Preliminary)](https://data.cdc.gov/Public-Health-Surveillance/Weekly-Hospital-Respiratory-Data-HRD-Metrics-by-Ju/mpgq-jmmr/about_data).
+
+This data was originally published by the CDC, and is made available here as a convenience to the forecasting community under the terms of the original license, which is [U.S. Government Public Domain](https://www.usa.gov/government-copyright).


aysim319:
also noticed that NSSP isn't mentioned in the https://cmu-delphi.github.io/delphi-epidata/api/covidcast_signals.html at the dashboad signals page when is it in the public dashboard.

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,69 @@
...
+
+Both datasets started reporting data in late 2022. As of May 2024, NHSN received data from 78% of US EDs.
+
+| Signal                          | Description                                                                                                                                                                         |
+|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `confirmed_admissions_covid_ew`              | Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week **Earliest date available:** 2020-08-08                                                                                                     |
+| `confirmed_admissions_covid_ew_prelim`          | Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday). <br/> **Earliest date available:** 2020-08-08  |
```
dsweber2, line 31: This is somewhat hard to follow. Might want to include a specific example, e.g. "As an example, for the issue week of `2024-12-08`, the specific release date was `2024-12-11` (the Wednesday of that week)". I'm not 100% sure this is what you're saying though.
```suggestion
| `confirmed_admissions_covid_ew_prelim`          | Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week. Preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday). <br/> **Earliest date available:** 2020-08-08  |
```

dsweber2: I'm also assuming we need the issue dates to be weekly? It'd probably be more accurate to have them daily, but that may be a larger issue

aysim319: I took the wording directly from the cdc website. It updates on **wednesday about the previous epiweek (so the numbers are generally bit lower) compared to the main source which the same data is published on **fridays

** the actual updates has been inconsistent 

The issue dates are indeed weekly;

nmdefries: I'm also finding the wording hard to follow.

```
Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week
```

should be more like

```
Total number of patients hospitalized with confirmed COVID-19 over the entire week (Sunday – Saturday). Only includes hospitalizations reported before Wednesday of the following week.
```

For the "prelim" one,

```
Total number of patients hospitalized with confirmed COVID-19 captured for the Wednesday of the reporting week preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday).
```

to

```
Total number of patients hospitalized with confirmed COVID-19 over the entire week (Sunday – Saturday). This is preliminary data that is <...whatever that means>. Only includes hospitalizations reported before Wednesday of the following week.
```

Make sure to clarify what "preliminary data" means -- my initial point of confusion is that the prelim and non-prelim data seem to be reported at the same time and for the same week. So what's the difference?

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,69 @@
...
+| `confirmed_admissions_flu_ew`           | Total number of patients hospitalized with confirmed influenza captured for the Wednesday of the reporting week  <br/> **Earliest date available:** 2020-08-08                                                                                       |
+| `confirmed_admissions_flu_ew_prelim`     | Total number of patients hospitalized with confirmed influenza captured for the Wednesday of the reporting week  preliminary data reported to NHSN for the previous reporting week (Sunday – Saturday).<br/> **Earliest date available:** 2020-08-08 |
+
+## Table of Contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+## Estimation
```
nmdefries, line 41: I'm going to reorganize some of this and then add more comments.

nmdefries:
Is "reporting dates" what we're calling "time values" now? Do we have new official terminology?

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,91 @@
...
+* **Number of data revisions since 18 Nov 2024:** 0
+* **Date of last change:** Never
+* **Available for:** state, hhs, nation (see [geography coding docs](../covidcast_geography.md))
+* **Time type:** week (see [date format docs](../covidcast_times.md))
+* **License:** [Public Domain US Government](https://www.usa.gov/government-works)
+
+## Overview
+
+[The National Healthcare Safety Network (NHSN)](https://www.cdc.gov/nhsn/index.html) is the nation’s most widely used healthcare-associated infection tracking system.
+This dataset represents preliminary weekly hospital respiratory data and metrics aggregated to national and state/territory levels reported to CDC’s National Health Safety Network (NHSN) beginning August 2020.
```
nmdefries, line 21: todo: clarify wording "beginning August 2020". Unclear if this is referring to issue date or reference date.

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+## Table of Contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+## Estimation
+
+All data is weekly such that each reported value represents a total from Sunday to Saturday of the reference week.
+The `output_signal_name` signal mirrors the `input signal name` field for all geographic resolutions except HHS.
```
nmdefries, line 50: issue: Please list all relationships between signal names in the source and our signal names.

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+## Missingness
+
+Data for reference dates through April 30, 2024 were reported during a federally-mandated reporting period as specified by the Secretary of the Department of Health and Human Services.
+Data for reference dates May 1, 2024 – October 31, 2024 were voluntarily reported in the absence of a mandate.
+As a result, during this period the total number of hospitalized patients on a subsection of hospitals and may not be fully representative.
+Data for reference dates beginning November 1, 2024 were reported during the current mandated reporting period.
+
+All data and metrics capturing information on respiratory syncytial virus (RSV) were voluntarily reported until November 1, 2024.
```
nmdefries, line 64: issue: I don't see that we're reporting any RSV signals. Are we supposed to be?

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+
+State and nation-level values are pulled directly from the source; HHS-level values are aggregated up from the state level by summing the values of member states.
+
+
+## Missingness
+
+Data for reference dates through April 30, 2024 were reported during a federally-mandated reporting period as specified by the Secretary of the Department of Health and Human Services.
+Data for reference dates May 1, 2024 – October 31, 2024 were voluntarily reported in the absence of a mandate.
+As a result, during this period the total number of hospitalized patients on a subsection of hospitals and may not be fully representative.
+Data for reference dates beginning November 1, 2024 were reported during the current mandated reporting period.
```
nmdefries, line 62: issue: My understanding is that the initial reporting mandate required hospitalizations due to COVID be reported, but didn't require hospitalizations due to flu. Flu hospitalizations are required to be reported under the current mandate, however. We should note all of this since we also report flu signals.

suggestion (optional): it would be nice to link to detailed versions of the various mandates so we can find info about them more easily in the future.

nmdefries: question: since this source contains weekly data, but the mandate gives day dates, does that mean that weeks that cross the mandate/no-mandate boundary could be reporting only a partial week?

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+
+All data and metrics capturing information on respiratory syncytial virus (RSV) were voluntarily reported until November 1, 2024.
+
+
+## Limitations
+
+Between reference dates 2024-05-01 and 2024-10-31, the total number of hospitalized patients on a subsection
+of hospitals and may not be fully representative, since reporting was voluntary.
```
nmdefries, line 70: todo: "subsection of hospitals and" seems to have dropped a word

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+See the [missingness section](#missingness) for more context.
+
+Standard errors and sample sizes are not applicable to these metrics.
+
+
+### Differences with HHS reports
+
+?
```
nmdefries, line 78: suggestion: Please do at least a brief comparison with the old HHS data. Since this is supposed to be a continuation of it, info about how many values match (if there were any overlapping dates), reporting frequency, reporting lag, geographic coverage, etc. would be useful.

`dsew-cpr` [did this to compare](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/dsew-cpr.html#differences-with-hhs-reports).

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+## Lag and Backfill
```
nmdefries, line 81: suggestion: What is backfill like? We normally say something like [% of values later modified](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/hhs.html#lag-and-backfill) or % of values never modified, and how long after the initial report modifications happen.

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,94 @@
...
+
+Data for reference dates through April 30, 2024 were reported during a federally-mandated reporting period as specified by the Secretary of the Department of Health and Human Services.
+Data for reference dates May 1, 2024 – October 31, 2024 were voluntarily reported in the absence of a mandate.
+As a result, during this period the total number of hospitalized patients on a subsection of hospitals and may not be fully representative.
+Data for reference dates beginning November 1, 2024 were reported during the current mandated reporting period.
+
+All data and metrics capturing information on respiratory syncytial virus (RSV) were voluntarily reported until November 1, 2024.
+
+
+## Limitations
```
nmdefries, line 67: suggestion: Any other limitations?

For example, the old HHS said [some types of hospitals and medical facilities were not included](https://cmu-delphi.github.io/delphi-epidata/api/covidcast-signals/hhs.html#limitations). Will that also apply here?

nmdefries (changes requested):
Requesting more detail in some sections.

docs/api/covidcast-signals/nhsn.md
aysim319: I wasn't sure how and where to add them (since the docs doesn't seem to be super standardized) so I left them blank for now.

aysim319:
> Total number of patients hospitalized with confirmed COVID-19 over the entire week (Sunday – Saturday). This is preliminary data that is <...whatever that means>. Only includes hospitalizations reported before Wednesday of the following week.
Make sure to clarify what "preliminary data" means -- my initial point of confusion is that the prelim and non-prelim data seem to be reported at the same time and for the same week. So what's the difference?

the preliminary comes few days earlier, both will report the same latest reporting week, but compared to the main one which you would have to wait a full weeks (friday-saturday) to pass, the preliminary data comes wednesday-ish.

preliminary (*last updated jan 2nd)
Screenshot 2025-01-07 at 11 54 46 AM

main (*last updated jan 3rd)
Screenshot 2025-01-07 at 11 54 36 AM

I believe that fluview mentioned that for the week of the new years the reporting updates would be delayed

docs/api/covidcast-signals/nhsn.md
nmdefries: Yep, that's correct

nmdefries:
> Is "reporting dates" what we're calling "time values" now? I think someone was working on defining new official terminology

In my edit, I switched all mentions of "reporting date" to "reference date" for clarity. I think "reporting date" sounds like it means "report date"/"issue date".

docs/api/covidcast-signals/nhsn.md
```diff
@@ -0,0 +1,69 @@
...
+title: NHSN ED Visits
+parent: Data Sources and Signals
+grand_parent: Main Endpoint (COVIDcast)
+nav_order: 1
+---
+# National Syndromic Surveillance Program Emergency Department Visits
+{: .no_toc}
+
+* **Source name:** `nhsn`
+* **Earliest issue available:** August 08, 2020
```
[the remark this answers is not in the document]
nmdefries: I'm still confused about this so I looked at the data a bit.

So it looks like for every issue, all dates back to Aug 2020 are reported. That means that backfill could happen at any time. I'll plan to do a brief backfill analysis and add results in.

I'll need multiple issues to look at backfill, but I can only see one issue (2024w51, returned in the data as Dec 15 2024). If I try to request other issues (2024 weeks 41-50), I don't get any results. So I'm not sure if this is a source problem or perhaps a patching problem or we haven't fully released the NHSN indicator yet?

docs/api/covidcast-signals/nhsn.md
aysim319: that I'm not sure; I added the disclaimer:

> Data reported as of December 1, 2020 are subject to thorough, routine data quality review procedures, including identifying and excluding invalid values from metric calculations and application of error correction methodology; data prior to this date may have anomalies that are not yet resolved.

but I'm not quite sure where to find the actual percentage of changes that happen.

is backfill different from patching? not quite sure you're asking here

docs/api/covidcast-signals/nhsn.md
aysim319: I would assume so

docs/api/covidcast-signals/nhsn.md
nmdefries: Generally sources don't report how often values are modified, so we'll need to do the analysis ourselves. I can do that, but [don't see more than one issue of `nhsn` in covidcast](https://github.com/cmu-delphi/delphi-epidata/pull/1570#discussion_r1906060187) to compare across.

RE backfill vs patching,

- patching is something that we (Delphi) do to fix an error with the timing or processing of data -- there was an outage on our side that caused us to miss data uploads; there was an outage upstream where data uploads weren't made; there was a bug in how we calculate signals.
- backfill (AKA "revisions") is when the source (not Delphi) revises data for geos/dates that they have previously been released. This can be due to an error (typo caused them to say there were 1010 cases instead of 101 cases, so they correct it to "101") or more commonly because they don't receive all reports for a given date/geo right away.

Health insurance claims is a great example of why the source might take a long time to receive all reports for a given date/geo. Imagine we're interested in covid-related claims for Texas on Jan 1. Some medical offices are really quick to submit their claims and the insurance company will receive those the same day (issue Jan 1). Most medical offices are slower and will take a week to submit, so the claims count will increase over the next week (issues Jan 2-Jan 8). Some medical offices will take months to submit, so the claims count won't be finalized until May (issue May 1).

The health insurance company modifies the claims count associated with Jan 1 every day between Jan 1 and May 1. That's backfill. We see the value changing over time (sometimes fast, sometimes slow), but it's not an _error_, per se.

(In some cases, the source could modify past values to fix an error and that really is patching in the same sense that we "patch", except that they generally don't do it in an issue-aware way, but we don't usually call that "patching".)

Hopefully that's clear. Let me know if not/if you have any followup questions.

nmdefries: So I want to include in this section a little analysis of the backfill behavior _by the source_. How long does it take the value for a reference date to no longer change substantially? For `google-symptoms`, because the whole system is instantaneous, there is no backfill. For insurance claims-based signals, it might take a month for a value to stabilize, and a high percentage of reference dates/geos are eventually modified.

docs/api/covidcast-signals/nhsn.md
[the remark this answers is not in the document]
nmdefries: From Amaris, I set the time value as 202352-202452 and issue date as 202444-202502 and got 3 issue dates ([202451, 202452, 202501]. the latest date available for the first run is dec 16th which is 202451. I need to patch in the earlier issue dates 202447 (the earilest available date from forcasting was Nov 18th) which there's a PR for

docs/api/covidcast-signals/nhsn.md
nmdefries: I added this in https://github.com/cmu-delphi/delphi-epidata/pull/1570/commits/4e4fe8350ebc963819eb5601945c878b0863c5fb

nmdefries (approved):
This is good, except for the failing test.

PR #1611, opened 2025-02-26
remove pinned version of jinja from requirements

melange396:
Our code doesnt directly use jinja, but Flask does (or at least, it did at one point).  In the past, we pinned jinja to earlier versions to get around an incompatibility with jinja 3.1.0 (in #891 and #877), but dependabot has since taken us well past that (first in #1370 and most recently in #1610).  This PR removes jinja from our enumerated requirements altogether.


korlaxxalrok (approved):
Looks good! 👍

PR #1615, opened 2025-03-03
Update CTIS documentation to point to ICPSR

capnrefsmmat:
Now that the CTIS data is [archived with ICPSR](https://www.icpsr.umich.edu/web/ICPSR/studies/39207), we can point there and deprecate pages describing access to microdata via CMU. ICPSR also archives the contingency tables, including a user guide, so we can remove the redundant contingency table documentation and point to ICPSR instead.

### Prerequisites:

- [ ] Build is successful

### DIFF
--- a/docs/symptom-survey/collaboration-revision.md
+++ b/docs/symptom-survey/collaboration-revision.md
@@ -2,9 +2,14 @@
 title: Collaboration and Survey Revision
 parent: <i>inactive</i> COVID-19 Trends and Impact Survey
 nav_order: 1
+nav_exclude: true
 ---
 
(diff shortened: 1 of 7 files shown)


melange396 (approved):
Looks good to me!  Do we need to do anything to remove data from `https://www.cmu.edu/delphi-web/surveys/*` (since we are removing all the references to it)?

capnrefsmmat:
I don't think we need to delete the aggregate data yet -- I dunno if there are other links to it -- but we can do that as we also decommission the microdata SFTP access and move people over to ICPSR.

Issue #1388, opened 2024-02-11
Some NY data missing from `fluview`


isaacyeaton:
It seems the NY data stops after epiweek `202349`.

| Fails | Epiweek | URL |
|:--:|:--:| :--- |
| No | `202349` | https://api.delphi.cmu.edu/epidata/fluview/?regions=ny&epiweeks=202349 |
| Yes | `202350` | https://api.delphi.cmu.edu/epidata/fluview/?regions=ny&epiweeks=202350 |
| Yes | `202404` | https://api.delphi.cmu.edu/epidata/fluview/?regions=ny&epiweeks=202404 |

However, CDC does have the data.

<img width="1784" alt="image" src="https://github.com/cmu-delphi/delphi-epidata/assets/988763/27f04d69-dca0-4e0c-adab-2b70ec262f47">

PR #1686, opened 2025-11-06
Catch and reissue `fluview` imputation errors as warnings to avoid failures

nmdefries:
### Summary:
[Context](https://delphi-org.slack.com/archives/C0130CSQRN3/p1761055120030719)

Imputed NY state data was not available because imputation failed on issue 2023w50. Of all states, only PR was reported for that issue, which caused an error to be raised. It isn't a bug -- the imputation code was set up to fail when the system (set of states) was all-missing, none-missing, or underdetermined (too many missing states to impute). Because the data is truly missing, we can't fix the root cause of the error or similar errors.

Instead, switch the errors to informational messages, but let imputation continue.

The imputation script should automatically fill all historical dates, so we won't need to patch this manually.

Closes https://github.com/cmu-delphi/delphi-epidata/issues/1388

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/src/acquisition/fluview/impute_missing_values.py
+++ b/src/acquisition/fluview/impute_missing_values.py
@@ -51,7 +51,6 @@ import argparse
 # third party
 import mysql.connector
 import numpy as np
-from warnings import warn
 
 # first party
 import delphi.operations.secrets as secrets


src/acquisition/fluview/impute_missing_values.py
```diff
@@ -293,7 +296,15 @@ def impute_missing_values(database, test_mode=False):
...
-        F, known, unknown = get_fusion_parameters(known_values.keys())
+        try:
+            F, known, unknown = get_fusion_parameters(known_values.keys())
+        except StatespaceException as e:
+            message = str(e)
+            if message == UNDERDETERMINED_MSG:
+                warn(message)
+            else:
+                print(message)
+            continue
```
melange396: `warn()` goes to stderr instead of stdout, and we can lose the log message.  might as well just print them all:
```suggestion
            # imputation cannot be performed for this issue/epiweek, log and move on
            print(e)
            continue
```

PR #1690, opened 2025-11-11
Add hsa_nci geo resolution + Adjust csv regex patterns

minhkhul:
Addresses issue(s) https://github.com/cmu-delphi/epidata-etl/issues/394

### Summary:

- Add hsa and hsa_nci geo resolution
- Adjust regex pattern to match specific geo resolutions.
- Add various `hsa_nci` and `hsa` cases to unit test to make sure the new regex is working as intended.

### Prerequisites:

- [x] Build is successful

### DIFF
--- a/integrations/server/test_covidcast.py
+++ b/integrations/server/test_covidcast.py
@@ -476,3 +476,14 @@ class CovidcastTests(CovidcastBase):
     self.assertEqual(expected, colond)
     self.assertEqual(expected, dashed)
     self.assertEqual(expected, enumed)
+
+  def test_hsa_nci(self):
+    row = CovidcastTestRow.make_default_row(geo_type='hsa_nci', geo_value='99')
+    self._insert_rows([row])
+    response = self.request_based_on_row(row)
+    expected = [row.as_api_row_dict()]
+    self.assertEqual(response, {
+      'result': 1,
+      'epidata': expected,
+      'message': 'success',
+    })


src/acquisition/covidcast/csv_importer.py
```diff
@@ -52,18 +52,21 @@ class CsvRowValue:
 class CsvImporter:
   """Finds and parses covidcast CSV files."""
 
+  # set of allowed resolutions (aka "geo_type")
+  GEOGRAPHIC_RESOLUTIONS = {'county', 'hrr', 'msa', 'dma', 'state', 'hhs', 'nation', "hsa", "hsa_nci"}
```
melange396: I think we should keep `hsa` out of the list, at least for now -- if/when we truly need it, itll be easy enough to put back.  This is partly because the other (Dartmouth) definition of HSA will be incompatible with this one and we dont want ambiguity about which is being used.  Keep the length sorting and associated comment though, in case people forget about that case in the future!
```suggestion
  GEOGRAPHIC_RESOLUTIONS = {'county', 'hrr', 'msa', 'dma', 'state', 'hhs', 'nation', 'hsa_nci'}
```

src/acquisition/covidcast/csv_importer.py
```diff
@@ -52,18 +52,21 @@ class CsvRowValue:
...
+
+  # regex pattern for matching geo types, note: sort longer string first to avoid wrong substring matches
+  geo_types_pattern = "|".join(sorted(GEOGRAPHIC_RESOLUTIONS, key=len, reverse=True))
+
   # .../source/yyyymmdd_geo_signal.csv
-  PATTERN_DAILY = re.compile(r'^.*/([^/]*)/(\d{8})_(\w+?)_(\w+)\.csv$')
+  PATTERN_DAILY = re.compile(rf'^.*/([^/]*)/(\d{{8}})_({geo_types_pattern})_(\w+)\.csv$')
```
melange396: Im partial to this string construction so we dont have to double-up the curly braces...  What do you think?  (same for the other pattern below, of course)
```suggestion
  PATTERN_DAILY = re.compile(r'^.*/([^/]*)/(\d{8})_(' + geo_types_pattern + r')_(\w+)\.csv$')
```

melange396 (changes requested):
Very nice job!  Your approach is a great way of handling this!

There are a couple other things we will need to take care of:
* The following is essentially dead code now; we should remove it at the least, but potentially replace its functionality: https://github.com/cmu-delphi/delphi-epidata/blob/ff823c65860c33f45710e2bdcd14aaa264fcb4d4/src/acquisition/covidcast/csv_importer.py#L194-L197 (an easy way would be to add "` (or geo_type)`" to [the log message where it will now error out](https://github.com/cmu-delphi/delphi-epidata/blob/ff823c65860c33f45710e2bdcd14aaa264fcb4d4/src/acquisition/covidcast/csv_importer.py#L164)).
* The new geo_type and some basic validation code should also be added to the section at: https://github.com/cmu-delphi/delphi-epidata/blob/ff823c65860c33f45710e2bdcd14aaa264fcb4d4/src/acquisition/covidcast/csv_importer.py#L306-L346...  According to https://seer.cancer.gov/seerstat/variables/countyattribs/hsa.html, valid codes should be 1-3 digit numbers, or the special code of "1022" (which NSSP doesnt appear to use, FWIW).
* I ***DO NOT*** think we should add the new geo_type to the part of the server code that does input validation, as its done with the older geomapper code which does not support this definition of HSA (it uses the incompatible Dartmouth version, see [doc](https://github.com/cmu-delphi/covidcast-indicators/blob/57e83053cbd1b61e1334cb2781ff4a0a6546576f/_delphi_utils_python/data_proc/geomap/README.md?plain=1#L32)/[urlref](https://github.com/cmu-delphi/covidcast-indicators/blob/57e83053cbd1b61e1334cb2781ff4a0a6546576f/_delphi_utils_python/data_proc/geomap/geo_data_proc.py#L23)/[crosswalk](https://github.com/cmu-delphi/covidcast-indicators/blob/57e83053cbd1b61e1334cb2781ff4a0a6546576f/_delphi_utils_python/delphi_utils/data/2020/zip_hsa_table.csv)).  However, we should mention in a comment there that we are intentionally sidestepping it for that reason: https://github.com/cmu-delphi/delphi-epidata/blob/fc450d5ba296c8ba882e775989b5b2ec5efc89fc/src/server/_params.py#L61-L69

src/acquisition/covidcast/csv_importer.py
```diff
@@ -188,10 +191,6 @@ def find_csv_files(scan_dir, issue=(date.today(), epi.Week.fromdate(date.today()
 
       # # extract and validate geographic resolution
```
melange396:
```suggestion
      # extract geographic resolution
```

src/server/_params.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, geo_type: str, geo_values: Union[bool, Sequence[str]]):
             if geo_values == ['']:
                 raise ValidationFailedException(f"geo_value is empty for the requested geo_type {geo_type}!")
             # TODO: keep this translator in sync with CsvImporter.GEOGRAPHIC_RESOLUTIONS in acquisition/covidcast/ and with GeoMapper
+            # We are not updating hsa_nci here as this was done with the older geomapper code which does not support the same hsa definition.
```
melange396:
```suggestion
            # NOTE: We are not including `hsa_nci` here as the geomapper code does not support that version of the HSA definition.
```

PR #1696, opened 2025-12-09
Warn and skip instead of failing when group represents the intersection of multiple demographic groups

nmdefries:
### Summary:

The `flusurv` code uses demographic group IDs to assign signal labels. Historically, for a given obs we only saw one of four group IDs be non-zero. The previous code would fail if an obs used more than one non-zero group code.

However, the source has introduced multi-demographic groups (only age group x flu type). Warn and skip these for now.

### Prerequisites:

- [x] Build is successful


dsweber2 (approved):
Confirmed with @nmdefries that the data contained in these rows is already reported elsewhere, so bypassing them is reasonable.

PR #1729, opened 2026-03-26
Persistent redis API keys

dsweber2:
### Summary:

This makes the redis instance also act as a key store. Instead of deleting the update time, the script checks to see that all of the keys are present, and adds them with a default of now if it's somehow `NULL`.

### Prerequisites:

- [x] Build is successful


dsweber2:
Funny that sonar cloud is flagging a password line I didn't change as the fault of this PR. happy to put the default in an environmental variable like in the most recent commit

melange396:
SonarCloud has had more noise than signal for us, and we have been mostly ignoring it for a while now...  however, its good to keep an eye on it in case it finds something crucial/critical.

redis is not really designed for persisting things.  if we go with this PR as-is, if the redis service gets restarted, the delphi-api will work fine, but cast-api will deny all api keys until `update_last_usage.py` is run again.  there is more discussion in [the slack thread](https://delphi-org.slack.com/archives/C08136HA2TZ/p1774538021304149) about alternative approaches, but this looks like it should work for the most part.

nolangormley (commented):
David and I talked about this more offline. I think there's enough safeguards added that this will work okay. We should explore running a refresh of the redis data on restart though.

@korlaxxalrok would that be as simple as running `update_last_usage.py`, like @melange396 mentioned?

korlaxxalrok:
> David and I talked about this more offline. I think there's enough safeguards added that this will work okay. We should explore running a refresh of the redis data on restart though.
> 
> @korlaxxalrok would that be as simple as running `update_last_usage.py`, like @melange396 mentioned?

It seems like this should work 👍

And we have persistence configured (we're using the RDB [flavor](https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/#rdb-advantages) with just the defaults present from doing the initial install). We get a dump of the DB to an `.rdb` file on the host after various "keys updated" thresholds are crossed. We don't back this file up in a historical way, but for just weathering a restart what we have should be decent.

melange396:
for the record, this doesnt capture any of the structure of user roles, which is ok until there are protected/controlled data sources in `epidata-etl` (like quidel)

nolangormley (approved):
Given @korlaxxalrok 's reply, I'm good with this one!

@melange396 is right about the roles, but that is a WIP on the cast-api side. I believe @dsweber2 already has a PR for that up.

PR #1751, opened 2026-07-01
include backfill method and exact function run

dsweber2:
### Summary:

Adds a backfill method to the API script in case the script is ever turned off

### Prerequisites:

- [ ] Build is successful

### DIFF
--- a/src/server/admin/api_key_form_script.js
+++ b/src/server/admin/api_key_form_script.js
@@ -14,7 +14,7 @@ function onSubmit(e) {
     var allResponses = form.getResponses();
     var latestResponse = allResponses[allResponses.length - 1];
 
-    var user_api_key = Math.random().toString(16).substr(2, 18);
+    var user_api_key = Utilities.getUuid().replace(/-/g, '');
     var user_email = latestResponse.getRespondentEmail();
 
     var payload = {
@@ -53,7 +53,7 @@ function onSubmit(e) {
         });
     } else if (result.getResponseCode() == 409) {
         Logger.log('Email already registered, replacing key for %s', user_email);
-        var new_api_key = Math.random().toString(16).substr(2, 18);
+        var new_api_key = Utilities.getUuid().replace(/-/g, '');
         var replacePayload = {
             'token': WEBHOOK_SECRET,
             'user_api_key': new_api_key,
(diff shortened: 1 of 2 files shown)


src/server/admin/api_key_form_script.js
```diff
@@ -13,7 +14,7 @@ function onSubmit(e) {
     var allResponses = form.getResponses();
     var latestResponse = allResponses[allResponses.length - 1];
 
-    var user_api_key = Math.random().toString(16).substr(2, 18);
+    var user_api_key = Utilities.getUuid().replace(/-/g, '');
```
dshemetov, line 17: This seems like a core change to the key gen algorithm. Is that intended?

src/server/admin/api_key_form_script.js
```diff
@@ -51,18 +52,132 @@ function onSubmit(e) {
...
         });
     } else if (result.getResponseCode() == 409) {
-        Logger.log('Registration was not successful, %s %s', result.getContentText("UTF-8"), result.GetResponseCode);
-        MailApp.sendEmail({
-            to: user_email,
-            subject: "Delphi Epidata API Registration",
-            noReply: true,
-            body: `
-            API Key was not generated.
+        Logger.log('Email already registered, replacing key for %s', user_email);
```
dshemetov, line 55: suggestion (non-blocking): this code block is reused a couple times that it might be worth turning it into a function, where the email body is an argument. And then you could mock that function out for dry run testing, if needed

dshemetov:
This sonarcloud code analysis is way too loud and noisy for the level of comments it generates (const var vs var is really not a code red emergency). If you find some way to be able to read a gh pr diff without that stuff inline, I'd be happy to hear about it

src/server/admin/api_key_form_script.js
dsweber2: that was mollifying sonarcloud

dshemetov: ...so we're changing our user key gen algo to appease a linter?

dsweber2: It's not a linter, its a security check, and it is right about it. I don't think this is a big deal to change

nolangormley (approved):
lgtm

PR #1755, opened 2026-08-15
V5 source documentation

JavierMtzRdz:
addresses issue(s) #ISSUE 

### Summary:

This PR
- Performs minor corrections to the provider and normalization names referenced in NWSS documentation. 
- removes nav ordering among V5 sources, and
- adds an initial version of the VA respiratory dataset. 

### Prerequisites:

- [X] Build is successful


docs/api/v5-signals/va_respiratory.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+# Veterans Affairs Seasonal Respiratory Surveillance (V5)
+{: .no_toc}
+
+| Attribute | Details |
+| :--- | :--- |
+| **Source Name** | `va_respiratory` |
+| **Data Source** | [VA Seasonal Disease Review](https://www.accesstocare.va.gov/Healthcare/SeasonalDiseaseReview) |
+| **Geographic Levels** | Facility (`va_facility`, cases only), State (`state`), MSA (`msa`, cases only), HHS Region (`hhs`), Census Division (`census_division`), Census Region (`census_region`), National (`nation`) |
```
JavierMtzRdz, line 15: Typically, we only mention the geoleves and refer to a page where their codes are explained, such as in the "Other endpoints" section. Since that section isn't available yet, I've temporarily added it here.

julesjulesjules:
Small look at v5:
image
- only pophive and nwss here "work" (the other 4 go to 404 error pages - i assume there are other efforts underway to add those in, so not sure if it makes sense to remove the hyperlink for now or not)

NWSS review: 
- table of contents is weirdly in the middle? 
image
- "Avian Influenza A(H5)"
image
(i have some other qualms about language here, but being specific about the flu type here is important to me)
- under suffixes, these are labelled the same, but under post processing methods they are different
image
image
- we say we're not aggregating, and then we say we are.. i think we just need a different synonym, for better comprehension - maybe "collecting"?
image

julesjulesjules:
VA Respiratory: 

- can these two links open in a new tab, rather than overwriting the current? 
image
- should be "diseases" (COVID-19 is a disease, not a pathogen)
image
- possibly add something like "Both files are available for download at the data source linked above" or something? 
image
- maybe add an (ex. "mi" for "Michigan" if "state" was selected for the geo_type)
image
(also, side note: do we include what the "key type" column means for things in the new v5 schema????) 
- Does VA not report true zeros?
image

julesjulesjules:
The table of contents categories vary widely across the data sources currently available here - i realize there should be some flexibility and differences, because different data sources have different stuff, but there also should be some balance of consistency, i think! so here's the list: 
https://docs.google.com/spreadsheets/d/15x6FfBrSIlkSnnWB0NE51D3eyhp_jWimweOobmmaZ-w/edit?usp=sharing
I tried to line things up / highlight things that seemed related but were out of order or slightly off in name, etc. to display the differences

JavierMtzRdz:
Addressed the review feedback:
- Moved the table of contents to right after the attributes table on all three pages.
- Reworded the H5 avian flu pathogen label for specificity.
- Aligned the Suffixes table's linearized-concentration label with the Post-processing table.
- Reworded NWSS's Source and Licensing section to say "collects" instead of "aggregates", to stop contradicting the Estimation section.
- External links now open in a new tab site-wide (via a small JS include), instead of per-link.
- Fixed "pathogens" -> "diseases" and added a download note on the VA page.
- Added a geo_value example and clarified that VA doesn't report true zeros; suppressed counts are only zero-filled during aggregation, not elsewhere.
- Standardized section naming/order (Schema Details, Estimation and Indicator Processing, Missingness before Limitations) across all three pages per the linked spreadsheet.

julesjulesjules (approved):
(left my comments in the main thread, and JavierMtzRdz addressed them!)

nmdefries (dismissed):
Generally looks fine to me, renders correctly, etc, but relying on Jules for detailed review of the content.

JavierMtzRdz:
@nmdefries It looks like the approval expired when I made minor changes and merged the dependent PR. Could you approve it again? It is ready to merge.

dshemetov (approved):
:shipit: :shipit: :shipit:

PR #1769, opened 2026-08-31
docs: add hidden standardized example for v5 source documentation and…

JavierMtzRdz:
… include V5 documentation for NHSN, NSSP and claims data

### Summary:

This PR extends #1755 by adding documentation for NHSN, NSSP, and claims data. It adheres to the standardized format outlined in `signal_doc_standard.md`, as recommended in https://docs.google.com/spreadsheets/d/15x6FfBrSIlkSnnWB0NE51D3eyhp_jWimweOobmmaZ-w/edit?usp=sharing.

### Prerequisites:

- [X] Build is successful


julesjulesjules:
Starting with NHSN

This format is likely going to get unwieldy for the side bar! I don't have a better idea right this second, but just noting: 
image

Probably don't need the V5 here. Guessing it was included to help differentiate it from the V4 page?
image

Shorten to "Week, ending Saturday"
image

Add in "(NHSN)"
image

Just want to confirm that it is count? "pct" is in the variable name, traditionally a percent indication
image


Possibly link to this or put an image of this somewhere to link to? https://www2.census.gov/geo/pdfs/maps-data/maps/reference/us_regdiv.pdf 
image


Are we doing real epi-weeks, with the January handling? or just weeks ending on saturdays? If we are doing real epiweeks, we should also note that in the top "at-a-glance" table with the Temporal Granularity section. If we aren't, we shouldn't say we are.
image

What do these two things mean? 
image

For this, could we change the purple section to say exactly what source names are the same in V4 --> V5? And then for the red, perhaps if we add little headers or bold before the content of each bullet, "New Signals" and "Expanded Geographies"?
image

Red - i still think we should either link to or describe why this matters / what a key type "gets you"
Green - we should have e.g. for all the types? or link to somewhere where a user can see all the options for all the geo_types? For example, are the census regions named? (i.e. "west" or are they numbered or something else?)
Light blue - what does "aggregation path" mean? i thought "fill_method" was how we were filling in values for missing weeks or calculated weeks
Pink - based on all the information above this table in the documentation, this should just be "Count" - unless the "pct" note mentioned above is actually a percent.
image

Move "hhs regions" to stay consistent in the smallest-to-largest listing you've been doing
image

This isn't true, there isn't anything about fill method in the geographic aggregation section, perhaps just remove?
image

We haven't done this with any previous documentation i've read so far on the v4 --> v5 switch
image
So perhaps remove? or we can instead link to various other documentation options for how to query?

Just some simplifications: 
image

image

If my picture is correct, i think for increased clarity, we should change this to - "The preliminary file is published approximately four days after the reference week ends. The finalized weekly files publish 6 to 7 days after the reference week ends. The CDC continues to revise... "
image

julesjulesjules:
NSSP - 

image
Again, would remove the "V5" here

Is it still CDC Socrata on the backend? I know RSocrata was taken away, and I thought now it was a big new thing....
image
this --- 
image
but maybe I just don't understand the underlying system :) 

Same temporal granularity note as for NHSN above, and in the NHSN one, we list the geographies in ascending order (largest last) and here we're all mixed up? I think we should stick with ascending order
image

Blue - we say it's the same and then we say we add a new signal, so perhaps we say "the same signal definitions plus an additional acute respiratory illness signal"
Pink - the whole is linked, so the "and" is a different color than the rest of the "regular"/non-code text (just noting in case that was not intended)
image

Pink - add "(NSSP)" after Program
Blue - add "of the previous three diseases listed" after category
ooo, also, let's add "Respiratory syncytial virus (RSV)" instead of just "RSV", just for this first mention
image

what's a "trailing" 3-week mean?
image

blue - can we link to a map of these areas?
orange - what separate path? the github mirror? or a different path within Socrata?
image

Add the mirror fallback and the "different path" source inside the tl;dr top table? 
image

Is there a published paper from delphi on this method that we could include in the documentation? Considering this feels like "we just made this up to give you more made up geographies, who knows if these actually provide reliable signals" 
image

Same epiweek comment as in NHSN
image

Previously we've put an "*" in here to indicate "all" with just referencing the beginning of the variable name
image

We should align language here - for nssp: 
image
for nhsn: 
image
I prefer the nhsn language, but happy to talk about it, if there is a reason they're different!

We should do "{full name} ({abbreviation})" format up here, so that we can reference the abbreviations later 
image

julesjulesjules:
Continued NSSP: 

I think we want to format this section similarly across sources (so see NHSN above) and also, this second bullet is confusing to me, so perhaps we could elaborate/clarify?
image

Oh! perhaps if we also link this little table to the above bullet, that would help - 
image

(also same feedback in the Schema table as in NHSN)

(also same "example query" feedback - i think that's better to just link to a full query documentation page/set of pages)

Perhaps change this first sentence to match the format of the NHSN section --- "Finalized weekly files publish"
image
Also, does this mean that continuing facilities do not submit revisions?

julesjulesjules:
Outpatient Claims: 

- Standardize "Geographic Levels" geographies in order - smallest to largest
- (generally, i guess, apply some corrections from above to this as well - but i'll try to note them all)

Why do we do this population weighted thing if we get raw daily counts? oh, maybe because we know we're not getting everything... well, how do we know what we're missing, to correct for that?
image

For this, does it make sense to publish what ICD-10's count? For NSSP, etc. - that information is available through the original source, but not here, I don't think. 
image

The relationship to V4 section is now all of a sudden not bullet points? I think it was bullet points in both NSSP and NHSN - I think this would be better if it was all the same format and grouped in the same order (roughly) - like a geographies section, a signals section, and then an other/misc type section, maybe.

Above where? could we link to it? (i'm assuming it means the big formula in the metric definition section, which should be linkable)
image

Is this correct, if we're reporting percentage over a 7-day window? 
image

Is this correct, if we're still calculating things if they have 5 days in the 7 day window? Is the day not just dropped and a "value" for the day is still calculated? Maybe the wording is just odd to me? 
image

julesjulesjules:
Inpatient Claims: 

- Did I already ask what the "Extra Key Columns" means? I can't remember, but that with the "Key Type" column in the Schema section probably needs a bit more info

Applies to this and outpatient claims for sure -- for consistency, should be title case like the rest of the Table of Contents list
image

image
These links in the banners don't open in a new tab, they open and overwrite the current tab.

- in general, is there a reason these are two separate sources, when it seems like, from this point in the documentation, that the source data is actually the same? it's just a bunch of different signals? probably not a problem/question for right now, but something to think about, possibly. 

Guessing this means that this signal is only available back to 2022; any chance we can also note that or add a superscript or something in the first tl;dr header table "Temporal Scope Start" section? Also possibly change the language here to better indicate that. "History of this signal begins 2022-08-01" or something?
image

See, inpatient & outpatient are pretty much identical! But the inpatient (on the left) has the geographic aggregation section split into two - i like that better, so i think we should split the outpatient paragraph (on the right) to match.
image

(since they're basically the same, assume all my outpatient feedback applies here too :D )

Text edits: 
image

We list all the geographic levels, but just give an e.g. for geographic name. Probably we should simplify everything to e.g. examples, and then i think we should add something separate later that links to/shows the full actual combination of geotypes and signals that are available
image

Add a period to outpatient, remove this small end phrase in inpatient. 
image

Inpatient (left) should be edited to match outpatient phrasing (right)
image

JavierMtzRdz:
Regarding the order of the sources, I agree, but we can address it in a different PR. 


NHSN: 
* All wording/text changes on NHSN/NSSP incorporated
* Good catch on `inpatient_beds_occupied_pct_ew`! I posted a question about this on Slack. 
* RE the maps, I'm finishing a page that consolidates all geographic references so they are not repeated. Added your link temporarily. 
* Corrected the Temporal Handling section. 
* RE the abbreviations like `totalconf<pathogen>newadmhosprep`, I added a reference to the data dictionary.
* RE the comment "For this, could we change the purple section to say exactly what source names are the same in V4 --> V5?", the source is "nhsn". I think it is clear, but would you like something more explicit, like "The V4 `nhsn` source (as named in the V4 API) was published [...]"
* Added little headers to the bullet points in the "Relationship to V4" section.
* RE the comment about the key type, I included a link to the relevant information in the `v5_api_queries.md` file.
* RE the `geo_value` column, I am completing the v5 page that corresponds to the [33 geo levels page](https://cmu-delphi.github.io/delphi-epidata/api/geographic_codes.html) referenced in #1755. 
* RE the `fill_method` column, it is the imputation method used during aggregation. Tried to clarify a little more. 
* RE the `value` column, I am awaiting clarification. 
* Changed geo level order. 
* RE the Fill method section, the mention of Geo Aggregation was describing "[the geographic aggregations] are a plain sum of the member state values." However, I found it clearer to include all related information in the Geographic aggregations section. I renamed that section "Geographic Handling" to align with the naming of the Time Handling section. 
* RE the example query, the detailed instructions on how to build queries were thoroughly covered in the `v5_api_queries.md` file, so I removed that section. 
* RE the missingness & Privacy section, I incorporated the simplifications, except for the "for small counts" as there may be other types of suppression, but that does not happen here either. 
* The wording in the Limitations section has been changed. 
* Updated the Lag & Backfill section to reflect the latest recommendations.

julesjulesjules:
> Good catch on inpatient_beds_occupied_pct_ew! I posted a question about this on Slack.
I think we still haven't heard heads or tails on this, maybe we can ping again in 24 hours?

> RE the comment "For this, could we change the purple section to say exactly what source names are the same in V4 --> V5?", the source is "nhsn". I think it is clear, but would you like something more explicit, like "The V4 nhsn source (as named in the V4 API) was published [...]"
Oh, sorry, i meant "signals" -- like, could we do a 1:1 table or list on the signal comparisons

JavierMtzRdz:
### NSSP

*  RSocrata was archived, but the backend portal at data.cdc.gov is still 100% Socrata.
* RE the comment "Pink - the whole is linked, so the "and" is a different color than the rest of the "regular"/non-code text (just noting in case that was not intended)", it is intended as a link altogether.
* RE the comment "what's a "trailing" 3-week mean?", it's an average calculated continuously over a set number of past periods
* Added a link to the  `has_nci` geo level. 
* Clarified how the states are grouped in geo aggregation. 
* Regarding the weighted-average aggregation, this method is not introduced by Delphi. It is a particular type of spatial aggregation or upscaling used in survey methods and spatial statistics (e.g., I found this [interesting paper](https://link.springer.com/article/10.1186/1476-072X-10-54) and [this one](https://journals.plos.org/digitalhealth/article?id=10.1371/journal.pdig.0000039)). Among the sources I found, the best one is ["Statistical Methods in Spatial Epidemiology"](https://onlinelibrary.wiley.com/doi/book/10.1002/9780470035771), but I'm not sure whether this general reference is helpful here. What do you think?
* Also, added "*" to smoothing section.
* RE the relationship to V4 section, they are different because NSSP used the same signals (+ a new one), and I wanted to point that out. Does "The V4 `nssp` source published ED visit percentages for nation, state, county, HSA, HRR, MSA, and HHS regions." sound good to you?
* Improved the relationship to V4 section bullets. 
* Clarified Lag & Backfill section.

JavierMtzRdz:
### NHSN  1:1 table

> Oh, sorry, i meant "signals" -- like, could we do a 1:1 table or list on the signal comparisons

The table would look like the image below. I feel that would be repetitive with the bullets as there are no renames.  

Screenshot 2026-09-09 at 5 49 08 PM

JavierMtzRdz:
/preview-docs
### Outpatient Claims:

* Applied previous relevant corrections.
* Adjusted the Geographic Aggregation section. 
* Added detailed ICD codes.
* Adjusted the relationship to V4 section to include bullets and linked the trailing ratio. 
* The temporal granularity is correct: the 7-day window means that each value is the mean of the reference date and the 6 previous days, and this is also true for the previous reference date and all daily dates. 
* RE the Missing & Privacy section, it is correct. It means that if the reference day or any of the previous six days are missing, the value for the reference day is still calculated unless the numerator is less than 100. Tried to clarify it a little more. 

### Outpatient Claims:

* Maintained uppercase in titles
* RE why claims sources are split into two sources: it is largely because it is more convenient to process them through two pipelines into two different Postgres databases. Cast API currently assumes each source name maps to a single database table and that the metadata generator builds signal lists from that table.
* Added similar outpatient comments and adjusted Relationship to v4 accordingly. 

## This round of revisions is finished

julesjulesjules:
> RE the comment "what's a "trailing" 3-week mean?", it's an average calculated continuously over a set number of past periods
So like: 
```
seven_sample_avg = zoo::rollsum(value, 
                                             k = 7, 
                                             fill = NA, 
                                             align = "right")/7
```
specifically aligning right instead of center or left?

> Regarding the weighted-average aggregation, this method is not introduced by Delphi. It is a particular type of spatial aggregation or upscaling used in survey methods and spatial statistics (e.g., I found this [interesting paper](https://link.springer.com/article/10.1186/1476-072X-10-54) and [this one](https://journals.plos.org/digitalhealth/article?id=10.1371/journal.pdig.0000039)). Among the sources I found, the best one is ["Statistical Methods in Spatial Epidemiology"](https://onlinelibrary.wiley.com/doi/book/10.1002/9780470035771), but I'm not sure whether this general reference is helpful here. What do you think?
I think it would be good to at least say something like - this method is called x and is described in further detail in this resource / these resources and link them! Just so people know we aren't making stuff up :D 

> Does "The V4 nssp source published ED visit percentages for nation, state, county, HSA, HRR, MSA, and HHS regions." sound good to you?
Yes!

> Re: the NHSN signal name table 
It is repetitive (but I do kinda like it :D possibly worth considering making a big dumb page that does make that table for all transition sources, just to be uber thorough, but we can table that. Could we change that top sentence of the "What's changed" section to: 
"The V4 nhsn source published the same admission counts for nation, HHS regions, and state. The signal names that overlap remain the same in the V4 to V5 transition." to sort of capture the essence of that table?

and 👍  to the claims comments

JavierMtzRdz:
/preview-docs
> So like:
seven_sample_avg = zoo::rollsum(value, 
                                             k = 7, 
                                             fill = NA, 
                                             align = "right")/7

Exactly. In this particular case, the time window k and the denominator are both 3.

> I think it would be good to at least say something like - this method is called x and is described in further detail in this resource / these resources and link them! Just so people know we aren't making stuff up :D

Added!

> It is repetitive (but I do kinda like it :D possibly worth considering making a big dumb page that does make that table for all transition sources, just to be uber thorough, but we can table that. Could we change that top sentence of the "What's changed" section to

I prefer having a table organized by source rather than a lengthy one listing all signals for every source. However, if you have another preference, please let me know. I have added a table for all sources!

nmdefries (dismissed):
Stamped. Relying on @julesjulesjules for detailed checks on content.

julesjulesjules (approved):

PR #1819, opened 2026-09-21
Nhsn ext cols fixup

dsweber2:
the gemfile changes are just what I needed to be able to run this locally, and .

there were a couple of signals with the wrong name ordering, and I added a note about the eventual long-format of age groups.

The API script is a change that happened there a while ago I forgot to make a change for


JavierMtzRdz (dismissed):
lgtm. Thanks!

JavierMtzRdz (approved):

